PR #14, opened 2025-01-12
ya missed an exclamation mark

rawb0t:
## Description
Adding a "!" to a block of example code where I'm *pretty* sure it's supposed to go.

## Test plan

I'll be honest, I didn't test it. I just read it and I'm pretty sure it needs this.

---


hayes-mysten (approved):
thanks!

PR #17, opened 2025-01-16
Add basic BCS types and utils to walrus sdk

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/walrus/src/utils/bcs.ts
+++ b/packages/walrus/src/utils/bcs.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { bcs } from '@mysten/bcs';
+import { bcs, fromBase64, toBase64 } from '@mysten/bcs';
 
 const MerkleNode = bcs.enum('MerkleNode', {
 	Empty: null,
@@ -27,8 +27,22 @@ export const BlobMetadata = bcs.enum('BlobMetadata', {
 	V1: BlobMetadataV1,
 });
 
+export const BlobId = bcs.bytes(32).transform({
+	input: (blobId: string | Iterable<number>) =>
+		typeof blobId === 'string' ? blobIdToBytes(blobId) : blobId,
+	output: (bytes: Uint8Array) => blobIdFromBytes(bytes),
+});
+
+export function blobIdFromBytes(blobId: Uint8Array): string {
+	return toBase64(blobId).replace(/=*$/, '').replaceAll('+', '-').replaceAll('/', '_');
+}
+
+export function blobIdToBytes(blobId: string): Uint8Array {
+	return fromBase64(blobId.replaceAll('-', '+').replaceAll('_', '/'));
+}
+
 export const BlobMetadataWithId = bcs.struct('BlobMetadataWithId', {
-	blob_id: bcs.bytes(32),
+	blob_id: BlobId,
 	metadata: BlobMetadata,
 });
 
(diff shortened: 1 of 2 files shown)


packages/walrus/src/utils/bcs.ts
```diff
@@ -0,0 +1,43 @@
...
+});
+
+export const BlobMetadataV1 = bcs.struct('BlobMetadataV1', {
+	encoding_type: EncodingType,
+	unencoded_length: bcs.u64(),
+	hashes: bcs.vector(SliverPairMetadata),
+});
+
+export const BlobMetadata = bcs.enum('BlobMetadata', {
+	V1: BlobMetadataV1,
```
williamrobertson13, line 27: 😉

packages/walrus/src/utils/index.ts
```diff
@@ -0,0 +1,67 @@
...
+	const secondary = minCorrect - safetyLimit;
+
+	const size = Math.floor((Math.max(unencodedLength, 1) - 1) / (primary * secondary)) + 1;
+
+	const sliversSize = (primary + secondary) * size * nShards;
+	const metadata = nShards * DIGEST_LEN * 2 + BLOB_ID_LEN;
+	return nShards * metadata + sliversSize;
+}
+
+function decodingSafetyLimit(nShards: number): number {
```
williamrobertson13, line 21: while a lot of these utilities are simple to implement in TypeScript, I was wondering if it would make sense to have any of these exposed via WASM in case the underlying encoding schemes or calculations change in the Walrus codebase?

packages/walrus/src/utils/bcs.ts
```diff
@@ -0,0 +1,43 @@
...
+	RedStuff: null,
+});
+
+export const BlobMetadataV1 = bcs.struct('BlobMetadataV1', {
+	encoding_type: EncodingType,
+	unencoded_length: bcs.u64(),
+	hashes: bcs.vector(SliverPairMetadata),
+});
+
+export const BlobMetadata = bcs.enum('BlobMetadata', {
```
williamrobertson13, line 26: mind throwing this guy in too?

side note: do you know why `blob_id` would be typed as `any` after being parsed with this schema?

```
export const BlobMetadataWithId = bcs.struct('BlobMetadataWithId', {
	blob_id: bcs.bytes(32),
	metadata: BlobMetadataV1,
});
```

williamrobertson13 (approved):

packages/walrus/src/utils/index.ts
hayes-mysten: I don't think they can change post mainnet, and I'd eventually prefer not depend on wasm if it can be avoided.  Not sure we will get there with the encoding pieces, but I'd rather have the basics in TS

packages/walrus/src/utils/bcs.ts
hayes-mysten: no... not sure why that would happen. let me see if I can replicate it

hayes-mysten: Screenshot 2025-01-16 at 11 25 18 AM

williamrobertson13: huh, well ignore me I might just have a bug or some other issue in my test script 😅 
 
image

packages/walrus/src/utils/index.ts
williamrobertson13: sounds good, just wanted to check

williamrobertson13 (approved):
still lgtm - thanks for the extra blob_id transforms

PR #18, opened 2025-01-17
Add initial walrus client

hayes-mysten:
## Description

## Test plan

---


packages/typescript/src/bcs/pure.ts
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) Mysten Labs, Inc.
```
hayes-mysten, line 1: this was previously part of mysten/sui/transactions, and wasn't exposed, but it pretty useful

packages/typescript/src/bcs/pure.ts
```diff
@@ -0,0 +1,91 @@
...
+		? ShapeFromPureTypeName<U>[]
+		: T extends `option<${infer U extends PureTypeName}>`
+			? ShapeFromPureTypeName<U> | null
+			: never;
+
+type PureTypeValidationError<T extends string> = T & {
+	error: `Invalid Pure type name: ${T}`;
+};
+
+export function pureBcsSchemaFromName<T extends PureTypeName>(
```
hayes-mysten, line 55: not sure about this name, maybe this should be `pureBcsSchemaFromTypeTag`?

packages/walrus/src/client.ts
```diff
@@ -0,0 +1,71 @@
...
+	| {
+			suiRpcUrl: string;
+			suiClient?: never;
+	  };
+
+export type WalrusClientConfig = {
+	packageConfig: WalrusPackageConfig;
+} & SuiClientOrRpcUrl;
+
+export class WalrusClient {
```
hayes-mysten, line 32: I think having a single client that interacts with both sui and walrus nodes probably makes more sense.  I played around with a few different ways things could be combined, but I think having one semi-stateful client makes the most sense.  We might want to break things out internally, or eventually this wrap multiple clients for different things, but having a single user facing client feels like the best place to start

packages/typescript/src/bcs/pure.ts
williamrobertson13: maybe `pureBcsSchemaFromTypeName` to match the type `PureTypeName`? `typeTag` also works too, I'm indifferent 😆

packages/walrus/src/client.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		return systemState;
+	}
+
+	async stakingState() {
+		return this.#objectLoader.loadFieldObject(
+			this.#packageConfig.stakingPoolId,
+			{
+				type: 'u64',
+				value: (await this.#systemObject()).version,
```
williamrobertson13, line 66: is it safe to use the system object version to fetch the inner staking dynamic field object?

packages/walrus/src/client.ts
williamrobertson13: yeah makes sense to me 👍🏼

packages/walrus/src/index.ts
```diff
@@ -1,4 +1,25 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-export {};
+import { getFullnodeUrl } from '@mysten/sui/client';
```
williamrobertson13, line 4: [just checking] did you mean to leave this test code here haha

packages/walrus/src/client.ts
```diff
@@ -0,0 +1,71 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { SuiClient } from '@mysten/sui/client';
+
+import { StakingInnerV1 } from './contracts/walrus/staking_inner.js';
+import { SystemStateInnerV1 } from './contracts/walrus/system_state_inner.js';
+import { System } from './contracts/walrus/system.js';
+import { SuiObjectDataLoader } from './utils/object-loader.js';
+
+export interface WalrusPackageConfig {
```
williamrobertson13, line 11: any preference on just passing `mainnet | testnet | devnet | custom` and hardcoding the package IDs for folks vs making people pass the package config? I know the client API setup is subject to change so this comment isn't really important but figured I'd ask haha

packages/walrus/src/utils/object-loader.ts
```diff
@@ -0,0 +1,132 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { BcsType } from '@mysten/bcs';
+import { pureBcsSchemaFromName } from '@mysten/sui/bcs';
+import type { PureTypeName, ShapeFromPureTypeName } from '@mysten/sui/bcs';
+import type { SuiClient, SuiObjectData, SuiObjectResponse } from '@mysten/sui/client';
+import DataLoader from 'dataloader';
+
+export class SuiObjectDataLoader extends DataLoader<string, SuiObjectData> {
```
williamrobertson13, line 10: very cool util 😄

williamrobertson13 (approved):

PR #21, opened 2025-01-18
Update walrus contracts and add readBlob method to client

hayes-mysten:
## Description

## Test plan

---


packages/walrus/src/utils/task-pool.ts
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+interface Task<T = unknown> {
+	run: (abortSignal: AbortSignal) => Promise<T>;
+	promise: PromiseWithResolvers<T>;
+}
+
+export class TaskPool {
```
Jordan-Mysten, line 9: If we want to avoid implementing this (and potentially getting something wrong) there are a lot of open source versions of this we could use.

packages/walrus/examples/read-blob.ts
```diff
@@ -0,0 +1,41 @@
...
+import { computeMetadata } from '../src/wasm.js';
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+
+const suiClient = new SuiClient({
+	url: getFullnodeUrl('testnet'),
+});
+
+const walrusClient = new WalrusClient({
+	network: 'testnet',
```
Jordan-Mysten, line 16: Still hate that `SuiClient` doesn't encode this information.

packages/walrus/src/utils/task-pool.ts
hayes-mysten: this is temporary.  @williamrobertson13 is investigating what we actually want to do for this

packages/walrus/examples/read-blob.ts
hayes-mysten: yeah... this is something we really should prioritize in our post-walrus sdk work

packages/walrus/src/client.ts
```diff
@@ -74,6 +94,92 @@ export class WalrusClient {
 		return this.#objectLoader.load(this.#packageConfig.stakingPoolId, Staking());
 	}
 
+	async #getNodes() {
+		if (!this.#nodes) {
+			const { promise, resolve, reject } = Promise.withResolvers<{
```
williamrobertson13, line 99: just double checking, `Promise.withResolvers` is stage 4 and available mostly everywhere so we don't need to polyfill this (it's enabled behind a flag in Node v21 AFAIK)?

packages/walrus/src/client.ts
```diff
@@ -74,6 +94,92 @@ export class WalrusClient {
...
+		const response = await fetch(`https://${nodeAddress}${path}`, init);
+
+		if (!response.ok) {
+			throw new Error(`Failed to fetch from node ${nodeAddress}: ${response.statusText}`);
+		}
+
+		return response;
+	}
+
+	async #fetchFromShard(index: number, path: string, init: RequestInit) {
```
williamrobertson13, line 171: I was actually starting to come around more on using the API client generated by hey-api and then just making a simple API client wrapper with retry capabilities baked in, we can always do this as a follow-up though just wanted to throw that out there haha

packages/walrus/src/client.ts
williamrobertson13: o btw does this really need to be a deferred promise vs just doing `return new Promise((resolve, reject) => {});`?

packages/walrus/src/client.ts
```diff
@@ -94,4 +200,82 @@ export class WalrusClient {
...
+			},
+		);
+
+		return SliverData.parse(new Uint8Array(await response.arrayBuffer()));
+	}
+
+	async readBlob(blobId: string) {
+		// calculate this using the unencoded_length and primary source symbols + other data
+		// wasm bindings? re-implement?
+		const concurrencyLimit = 20;
```
williamrobertson13, line 242: might want to stick a big `TODO` here

packages/walrus/src/client.ts
```diff
@@ -94,4 +200,82 @@ export class WalrusClient {
...
+					if (error.name === 'AbortError') {
+						return;
+					}
+
+					throw error;
+				}),
+		);
+
+		await Promise.allSettled(sliverPromises);
+
```
williamrobertson13, line 274: I reckon I'll take care of the retry/async task management stuff that needs to happen here, not sure if you want to leave a TODO in the meantime or what

williamrobertson13 (approved):

PR #23, opened 2025-01-21
fix: validate queryEvent parameter

LI-YONG-QI:
## Descriptiona

Fix the issue that parameter of `queryEvent` can't be a multi objects.

The Typescript type checker won’t throw error if developers pass multi objects as `query`, but this is invalid format for sending request to rpc and the error message returned from rpc response isn't clear. Therefore, I added a validation logic and more clear error message before request to check keyof input object

```typescript
      await toolbox.client.queryEvents({
		query: {
			TimeRange: { startTime: '0', endTime: Date.now().toString() },
			Sender: toolbox.address(), // not throw error in VSCode type safe checker
		},
	});
```

I have not found any additional rules or pattern for contribution open source code, so if I make something wrong, please tell me

## Test plan

Added error test in e2e test cases

---

### DIFF
--- a/packages/typescript/src/client/client.ts
+++ b/packages/typescript/src/client/client.ts
@@ -522,6 +522,9 @@ export class SuiClient {
 	 * Get events for a given query criteria
 	 */
 	async queryEvents(input: QueryEventsParams): Promise<PaginatedEvents> {
+		if (Object.keys(input.query).length > 1)
+			throw new Error('Invalid query parameters, please provide only one query object');
+
 		return await this.transport.request({
 			method: 'suix_queryEvents',
 			params: [
--- a/packages/typescript/test/e2e/read-events.test.ts
+++ b/packages/typescript/test/e2e/read-events.test.ts
@@ -38,4 +38,15 @@ describe('Event Reading API', () => {
 		});
 		expect(query1.data.length).toEqual(0);
 	});
+
+	it('Error: Get All Events', async () => {
+		await expect(async () => {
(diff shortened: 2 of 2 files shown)


packages/typescript/src/client/client.ts
```diff
@@ -522,6 +522,9 @@ export class SuiClient {
 	 * Get events for a given query criteria
 	 */
 	async queryEvents(input: QueryEventsParams): Promise<PaginatedEvents> {
+		if (Object.keys(input.query).length > 1)
+			throw new Error('Invalid query parameters, please provide only one query object');
```
hayes-mysten, line 526: We don't generally do runtime validation of query inputs, and have generally relied on errors returned from the API and typescript to enforce correct usage.

I am a little hesitant to start adding this kind of validation only to specific methods, leading to an inconsistent experience.

If you wanted to improve this, I think the correct way to do this would be to update the generator that creates these types to be more strict: https://github.com/MystenLabs/ts-sdks/blob/184c784679648fed2b9e1e48386ceecc91a19886/packages/typescript/scripts/generate.ts#L573

This could either be by updating the types to be something like `{ TimeRange: string, Sender?: never } | { TimeRange?: never, Sender: string }` or by creating a wrapping type utility that dynamically transforms these union types

LI-YONG-QI: @hayes-mysten  Thanks for your reviewing! I might need some time to try to improve and learn how to fix that

PR #32, opened 2025-01-27
SuiNS SDK

tonylee08:
## Description

SuiNS SDK

## Test plan

Tested all features locally, doing more thorough testing with frontend team. Moving SDK from suins-contracts repository to this repo


packages/suins/src/pyth/PriceServiceConnection.ts
```diff
@@ -0,0 +1,48 @@
...
+	 *
+	 * @param endpoint endpoint URL to the price service.
+	 * @param config Optional configuration for custom setups.
+	 */
+	constructor(endpoint: string, config?: PriceServiceConnectionConfig) {
+		this.httpClient = axios.create({
+			baseURL: endpoint,
+			timeout: config?.timeout || 5000,
+		});
+		axiosRetry(this.httpClient, {
```
hayes-mysten, line 29: @williamrobertson13 this might be worth updating once we figure out what we want to standardize on in walrus

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+	 * @param priceIds Array of hex-encoded price IDs.
+	 * @returns Array of buffers containing the price update data.
+	 */
+	async getPriceFeedsUpdateData(priceIds: HexString[]): Promise<Buffer[]> {
+		const latestVaas = await this.getLatestVaas(priceIds);
+		return latestVaas.map((vaa) => Buffer.from(vaa, 'base64'));
+	}
+}
+export class SuiPythClient {
+	private pythPackageId: ObjectId | undefined;
```
hayes-mysten, line 30: we have been using private fields (prefixed with `#`) instead of the typescrypt `private` keyword

hayes-mysten: I also generally prefer marking fields as optional (with `field?: type`) over explicitly adding `undefined` to the type

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+	private wormholePackageId: ObjectId | undefined;
+	private priceFeedObjectIdCache: Map<HexString, ObjectId> = new Map();
+	private priceTableInfo: { id: ObjectId; fieldType: ObjectId } | undefined;
+	private baseUpdateFee: number | undefined;
+	constructor(
+		public provider: SuiClient,
```
hayes-mysten, line 36: can we define these as class properties rather than using the constructor shorthand?

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+		public wormholeStateId: ObjectId,
+	) {}
+	/**
+	 * Verifies the VAAs using the Wormhole contract.
+	 *
+	 * @param vaas Array of VAA buffers to verify.
+	 * @param tx Transaction block to add commands to.
+	 * @returns Array of verified VAAs.
+	 */
+	async verifyVaas(vaas: Buffer[], tx: Transaction) {
```
hayes-mysten, line 47: We generally don't use `Buffer` because its specific to node.  Uint8Array is our default for raw bytes

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+		const wormholePackageId = await this.getWormholePackageId();
+		const verifiedVaas = [];
+		for (const vaa of vaas) {
+			const [verifiedVaa] = tx.moveCall({
+				target: `${wormholePackageId}::vaa::parse_and_verify`,
+				arguments: [
+					tx.object(this.wormholeStateId),
+					tx.pure(
```
hayes-mysten, line 55: this seems overly complicated.  You should just be able to do:  `bcs.vector('u8', vaa)` if vaa is a `UInt8Array`

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+export class SuiPriceServiceConnection extends PriceServiceConnection {
+	/**
+	 * Fetch price feed update data.
+	 *
+	 * @param priceIds Array of hex-encoded price IDs.
+	 * @returns Array of buffers containing the price update data.
+	 */
+	async getPriceFeedsUpdateData(priceIds: HexString[]): Promise<Buffer[]> {
+		const latestVaas = await this.getLatestVaas(priceIds);
+		return latestVaas.map((vaa) => Buffer.from(vaa, 'base64'));
```
hayes-mysten, line 26:
```ts
import { fromBase64 } from '@mysten/sui/utils'

latestVaas.map((vaa) => fromBase64(vaa));
```

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+				throw new Error(`Price feed ${feedId} not found, please create it first`);
+			}
+			priceInfoObjects.push(priceInfoObjectId);
+			[priceUpdatesHotPotato] = tx.moveCall({
+				target: `${packageId}::pyth::update_single_price_feed`,
+				arguments: [
+					tx.object(this.pythStateId),
+					priceUpdatesHotPotato,
+					tx.object(priceInfoObjectId),
+					coins[coinId],
```
hayes-mysten, line 126: `coinWithBalance({ balance: baseUpdateFee })`

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+		});
+		return priceInfoObjects;
+	}
+	/**
+	 * Get the priceFeedObjectId for a given feedId if not already cached
+	 * @param feedId
+	 */
+	async getPriceFeedObjectId(feedId: HexString): Promise<ObjectId | undefined> {
+		const normalizedFeedId = feedId.replace('0x', '');
+		if (!this.priceFeedObjectIdCache.has(normalizedFeedId)) {
```
hayes-mysten, line 145: You don't update the cache until you already loaded the data, this means if you request the same id multiple times without waiting you will request the same thing multiple times.  I generally prefer to cache the promise for the result instead of the result directly

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+				name: {
+					type: 'vector<u8>',
+					value: 'price_info',
+				},
+			});
+			if (!result.data || !result.data.type) {
+				throw new Error('Price Table not found, contract may not be initialized');
+			}
+			let type = result.data.type.replace('0x2::table::Table<', '');
+			type = type.replace('::price_identifier::PriceIdentifier, 0x2::object::ID>', '');
```
hayes-mysten, line 187: use `parseStructTag` instead

packages/suins/src/pyth/pyth.ts
```diff
@@ -0,0 +1,268 @@
...
+		return this.priceTableInfo;
+	}
+	/**
+	 * Extracts the VAA bytes embedded in an accumulator message.
+	 *
+	 * @param accumulatorMessage The accumulator price update message.
+	 * @returns VAA bytes as a Buffer.
+	 */
+	// eslint-disable-next-line @typescript-eslint/ban-types
+	extractVaaBytesFromAccumulatorMessage(accumulatorMessage: Buffer): Buffer {
```
hayes-mysten, line 199: this doesn't look like it should be a class method

packages/suins/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+export { SuinsTransaction } from './suins-transaction.js';
+export type { Network, SuinsClientConfig, Config } from './types.js';
+export {
+	getConfigType,
+	getDomainType,
+	getPricelistConfigType,
+	getRenewalPricelistConfigType,
+	getCoinDiscountConfigType,
+	ALLOWED_METADATA,
+	mainPackage,
```
hayes-mysten, line 13: this seems like it could have a better name

packages/suins/src/index.ts
```diff
@@ -0,0 +1,15 @@
...
+} from './constants.js';
```
hayes-mysten, line 14: most of these aren't really constants

packages/suins/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"lint:fix": "pnpm run eslint:fix && pnpm run prettier:fix"
+	},
+	"engines": {
+		"node": ">=16"
+	},
+	"dependencies": {
+		"@mysten/sui": "^1.21.0",
+		"axios": "^0.24.0",
+		"axios-retry": "^3.9.1",
+		"buffer": "^6.0.3"
```
hayes-mysten, line 47: I think we want to avoid this if at all possible

packages/suins/package.json
```diff
@@ -0,0 +1,68 @@
...
+	},
+	"devDependencies": {
+		"@mysten/build-scripts": "workspace:*",
+		"@types/tmp": "^0.2.6",
+		"bn.js": "^5.2.1",
```
hayes-mysten, line 52: what is for?

packages/suins/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"cross-env": "^7.0.3",
+		"tmp": "^0.2.3",
```
hayes-mysten, line 54: this looks unused

packages/suins/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"@mysten/sui": "^1.21.0",
+		"axios": "^0.24.0",
+		"axios-retry": "^3.9.1",
+		"buffer": "^6.0.3"
+	},
+	"devDependencies": {
+		"@mysten/build-scripts": "workspace:*",
+		"@types/tmp": "^0.2.6",
+		"bn.js": "^5.2.1",
+		"cross-env": "^7.0.3",
```
hayes-mysten, line 53: unused

packages/suins/package.json
```diff
@@ -0,0 +1,68 @@
...
+		"tmp": "^0.2.3",
+		"ts-retry-promise": "^0.8.1",
+		"typescript": "^5.7.2",
+		"vite": "^6.0.7",
+		"vitest": "^2.1.8",
+		"wait-on": "^8.0.1"
```
hayes-mysten, line 59: unused

packages/suins/src/suins-client.ts
```diff
@@ -0,0 +1,359 @@
...
+
+	/**
+	 * Returns the price list for SuiNS names in the base asset.
+	 */
+
+	// Format:
+	// {
+	// 	[ 3, 3 ] => 500000000,
+	// 	[ 4, 4 ] => 100000000,
+	// 	[ 5, 63 ] => 20000000
```
hayes-mysten, line 54: Using arrays as keys in a Map is probably a bad idea... you can get them out by iterating over them, but if you do map.get([3, 3]) it won't get you back the entry

packages/suins/src/suins-client.ts
```diff
@@ -0,0 +1,359 @@
...
+	 * 3. if the years are not between 1 and 5
+	 */
+	async calculatePrice({
+		name,
+		years,
+		isRegistration = true,
+	}: {
+		name: string;
+		years: number;
+		isRegistration: boolean;
```
hayes-mysten, line 282: this looks like it's supposed to be optional

packages/suins/src/suins-transaction.ts
```diff
@@ -0,0 +1,544 @@
...
+				params.maxAmount,
+			);
+			return receipt;
+		}
+	};
+
+	/**
+	 * Applies a coupon to the payment intent.
+	 */
+	applyCoupon = (intent: TransactionObjectArgument, couponCode: string) => {
```
hayes-mysten, line 240: this class inconsistently uses class methods vs properties (`applyCoupon = (...) => {` vs ``applyCoupon(...)  {`)

packages/suins/src/types.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export type VersionedPackageId = {
+	latest: string;
+	v1: string;
+	[key: string]: string;
+};
+
+export type Config = Record<'mainnet' | 'testnet', PackageInfo>;
+
+export type CoinConfigType = {
```
hayes-mysten, line 16: for these kinds of simple object types, prefer using `interface` over `type`

packages/suins/src/index.ts
leecchh: Moving to helpers as functions

packages/suins/src/suins-client.ts
leecchh: The only use case here is iteration, can leave as is for now. Direct lookups are not expected.

packages/suins/src/suins-transaction.ts
leecchh: Updated all to class methods

packages/suins/src/types.ts
leecchh: Updated

leecchh:
@hayes-mysten Thanks for all the suggestions, almost all have been implemented. Please take another look, especially around caching the promise for getPriceFeedObjectId

packages/suins/package.json
```diff
@@ -0,0 +1,63 @@
...
+	"scripts": {
+		"clean": "rm -rf tsconfig.tsbuildinfo ./dist",
+		"build": "build-package",
+		"prepublishOnly": "pnpm build",
+		"prettier:check": "prettier -c --ignore-unknown .",
+		"prettier:fix": "prettier -w --ignore-unknown .",
+		"eslint:check": "eslint --max-warnings=0 .",
+		"eslint:fix": "pnpm run eslint:check --fix",
+		"lint": "pnpm run eslint:check && pnpm run prettier:check",
+		"lint:fix": "pnpm run eslint:fix && pnpm run prettier:fix"
```
hayes-mysten, line 38: you need a test command for tests to run in CI

hayes-mysten (commented):
this looks a lot better, thanks for fixing all that so quickly!  Noticed a few more things on a second pass

packages/suins/src/suins-client.ts
```diff
@@ -0,0 +1,360 @@
...
+		}
+	}
+
+	/**
+	 * Returns the price list for SuiNS names in the base asset.
+	 */
+
+	// Format:
+	// {
+	// 	[ 3, 3 ] => 500000000,
```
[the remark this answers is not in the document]
tonylee08: Understood 👍

packages/suins/package.json
tonylee08: There are only live tests for suins (no local e2e due to oracle integration). What's best practice to include this?

hayes-mysten: I don't think we really have anything for that... probably fine to just leave it as is for now

hayes-mysten (approved):

PR #36, opened 2025-01-29
[wip][walrus] calculate sensible concurrency limits for reads

williamrobertson13:
## Description

This PR adds configuration options for read concurrency which pretty closely mirrors the existing Walrus client built in Rust. Basically, you can customize the read concurrency for slivers so that we'll never request more data than `maxDataInFlight` at a time or go over the `maxConcurrentSliverReads`. I imagine we will also want to copy over similar limits for writes and throw in some other customization options (ala custom fetch or retry configs).
 
927 byte blob example
----
- 1000 shards, 329 primary source symbols, 662 secondary source symbols
- 329 primary * 662 secondary = 217798 total source symbols for a blob
- 927 bytes / 217798 total source symbols = each symbol is 0.00425 bytes
- 0.004251626499662533 bytes per symbol * 329 primary source symbols = 1.3987 bytes per sliver
- 12,500,000 bytes can be in flight at a time by default, so 12,500,000 / 1.3987 bytes = 8,936,326 concurrent requests
- By default, we only allow n - 2f concurrent sliver reads at time = 334 
- so the limit = min(8,936,326, 334) = 334 concurrent reads
(there are some small extra bits in the PR for supporting empty blobs with 0 bytes)

## Test plan
- I think this math is roughly correct 😆

### DIFF
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -10,7 +10,13 @@ import { Staking } from './contracts/staking.js';
 import { SystemStateInnerV1 } from './contracts/system_state_inner.js';
 import { System } from './contracts/system.js';
 import { BlobMetadataWithId, SliverData } from './utils/bcs.js';
-import { getPrimarySourceSymbols, getShardIndicesByNodeId, toShardIndex } from './utils/index.js';
+import {
+	getMaxConcurrentSliverReads,
+	getShardIndicesByNodeId,
+	getSliverSizeForBlob,
+	getSourceSymbols,
+	toShardIndex,
+} from './utils/index.js';
 import { SuiObjectDataLoader } from './utils/object-loader.js';
 import { TaskPool } from './utils/task-pool.js';
 import { decodePrimarySlivers } from './wasm.js';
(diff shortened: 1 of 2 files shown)


packages/walrus/src/client.ts
```diff
@@ -78,6 +95,8 @@ export class WalrusClient {
 			this.#packageConfig = config.packageConfig!;
 		}
 
+		this.#communicationConfig = config.communicationConfig;
```
williamrobertson13, line 97: some of the default values for the communication config depend on data from the system state, I'm not sure the best way to handle this:
- set the defaults on the fly when they are accessed / needed
- make a static factory-esque function to initialize a new client instance with the system data already fetched
- ???

packages/walrus/src/client.ts
```diff
@@ -42,7 +53,10 @@ type WalrusNetworkOrPackageConfig =
 			packageConfig: WalrusPackageConfig;
 	  };
 
-export type WalrusClientConfig = WalrusNetworkOrPackageConfig & SuiClientOrRpcUrl;
+export type WalrusClientConfig = {
+	communicationConfig?: WalrusCommunicationConfig;
```
Jordan-Mysten, line 57: I don't love this name but we can bikeshed it later

packages/walrus/src/client.ts
Jordan-Mysten: Could we just make this non-configurable? What's the justification for changing these values?

Jordan-Mysten (approved):

williamrobertson13:
nvm not going this direction

PR #51, opened 2025-02-03
[walrus] replace hey-api codegen with a custom fetch-based client

williamrobertson13:
## Description

I learned the hard way to just roll my own simple API client 😆. None of the OpenAPI codegen libraries (openapi-fetch, hey-api, Stainless, Speakeasy) give you what you fully need, we have to have wrappers to parse BCS anyway, and the Walrus OpenAPI spec is generally very inconsistent (broken) in its current form. IMO all the tools get you 95% there before you start running into typing issues, having to write spec parsers, not being able to support streaming request bodies and responses, platform compatibility issues, etc.

This PR introduces a `StorageNodeClient` which is just a manual fetch-based HTTP client with some of the error handling inspired by Stainless and hey-api/client-fetch (useful for doing debugging + doing retry logic based on specific error scenarios). There are a few follow-ups on this like potential `node.fsReadStream` support + attaching some default request headers, but I can do that after this lands!

## Test plan
- The test examples still work and I tested the error responses too

Example error:
image


packages/walrus/src/utils/randomness.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export function getRandom<T>(array: T[]): T {
```
williamrobertson13, line 4: there will be some more randomness-related utils in here, i.e. for selecting the random commitee members weighted by number of shards for reads

packages/walrus/src/storage-node/client.ts
```diff
@@ -0,0 +1,183 @@
...
+	 *
+	 * If not provided, defaults to the global `fetch` function (`globalThis.fetch`).
+	 *
+	 * @default globalThis.fetch
+	 */
+	fetch?: Fetch;
+};
+
+export type RequestOptions = {
+	nodeUrl: string;
```
williamrobertson13, line 35: I considered baking in some of the node selection logic into this client but thought it would be best to leave this more of a pure/traditional client and making you pass a `nodeUrl` with each request

packages/walrus/src/utils/randomness.ts
Jordan-Mysten: Just to confirm it's fine that this isn't true randomness right? I imagine yes but just wanted to make sure

packages/walrus/src/storage-node/client.ts
```diff
@@ -0,0 +1,183 @@
...
+			const errorJSON = safeParseJSON(errorText);
+			const errorMessage = errorJSON ? undefined : errorText;
+			throw StorageNodeAPIError.generate(response.status, errorJSON, errorMessage);
+		}
+
+		return response;
+	}
+}
+
+function safeParseJSON(value: string) {
```
Jordan-Mysten, line 177: "safe" cries in zod schemas

packages/walrus/src/storage-node/client.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export type StorageNodeClientOptions = {
+	/**
+	 * An optional custom fetch function.
+	 *
+	 * If not provided, defaults to the global `fetch` function (`globalThis.fetch`).
+	 *
+	 * @default globalThis.fetch
+	 */
+	fetch?: Fetch;
```
Jordan-Mysten, line 31: We talked about making this actually `(request: Request) => Response` instead so that we can eventually shove in extra params

Jordan-Mysten (approved):

packages/walrus/src/utils/randomness.ts
williamrobertson13: Yeah it doesn't need to be truly random AFAIK!

packages/walrus/src/client.ts
```diff
@@ -132,67 +131,35 @@ export class WalrusClient {
...
-					});
-				})
-				.catch((error) => {
-					// handle abort signal
-					if (error.name === 'AbortError') {
-						return;
-					}
+		const sliverPromises = Array.from({ length: minSymbols }).map(async (_, shardIndex) => {
+			const storageNode = await this.#getNodeByShardIndex(shardIndex);
+			const sliverPairIndex = toPairIndex(shardIndex, blobId, numShards);
```
hayes-mysten, line 150: I am a little unsure about this naming.  sliverPairs is a list like `{ primary: { index: 0, symbols: [...] }, secondary: index: 999, symbols: [...]} }`,    I have been referring to the inner index (which is different between primary and secondary as the `sliverIndex`.    For primary slivers its the same as the index in the sliverPairs array.

For secondary slivers (used in the write flows) the `sliverPairIndex` name doesn't make sense to me, but I think just `sliverIndex` works consistently everywhere

packages/walrus/src/client.ts
```diff
@@ -132,67 +131,35 @@ export class WalrusClient {
...
+		});
 
 		// TODO: implement retry/scheduling logic
-		await Promise.allSettled(sliverPromises);
+		const sliverResults = await Promise.allSettled(sliverPromises);
+		const slivers = sliverResults
+			.map((result) => (result.status === 'fulfilled' ? result.value : null))
+			.filter((sliver) => !!sliver);
 
 		if (slivers.length < minSymbols) {
```
hayes-mysten, line 164: I am assuming adding more slivers to read from is going to come as part of the work to add the sliver selection logic?

packages/walrus/src/client.ts
```diff
@@ -461,88 +428,48 @@ export class WalrusClient {
 		return { digest };
 	}
 
-	async writeSliver({ blobId, sliverIndex, type, sliver, signal }: WriteSliverOptions) {
+	async writeSliver({ blobId, sliverPairIndex, sliverType, sliver, signal }: WriteSliverOptions) {
```
hayes-mysten, line 431: commented on this above, but I don't think sliverPairIndex is the right name to use here

packages/walrus/src/storage-node/client.ts
```diff
@@ -0,0 +1,183 @@
...
+		{ nodeUrl, ...options }: RequestOptions,
+	): Promise<StoreBlobMetadataResponse> {
+		const isBcsInput = typeof metadata === 'object' && 'V1' in metadata;
+		const body = isBcsInput ? BlobMetadata.serialize(metadata).toBytes() : metadata;
+
+		await this.#request(`${nodeUrl}/v1/blobs/${blobId}/metadata`, {
+			...options,
+			method: 'PUT',
+			body,
+			headers: { 'Content-Type': 'application/octet-stream' },
```
hayes-mysten, line 81: we should merge headers from options here

packages/walrus/src/storage-node/client.ts
```diff
@@ -0,0 +1,183 @@
...
+	): Promise<StoreSliverResponse> {
+		const isBcsInput = typeof sliver === 'object' && 'symbols' in sliver;
+		const body = isBcsInput ? SliverData.serialize(sliver).toBytes() : sliver;
+
+		await this.#request(`${nodeUrl}/v1/blobs/${blobId}/slivers/${sliverPairIndex}/${sliverType}`, {
+			...options,
+			method: 'PUT',
+			body,
+			headers: { 'Content-Type': 'application/octet-stream' },
+		});
```
hayes-mysten, line 121: Do we need to ensure the body of this request gets consumed?  I don't know if it will cause issues (leave the request hanging if you never read the body)

packages/walrus/src/storage-node/error.ts
```diff
@@ -0,0 +1,117 @@
...
+	/** JSON body of the response that caused the error. */
+	readonly error: TError;
+
+	constructor(status: TStatus, error: TError, message: string | undefined) {
+		super(StorageNodeAPIError.#makeMessage(status, error, message));
+		this.status = status;
+		this.error = error;
+	}
+
+	static #makeMessage(status: number | undefined, error: unknown, message: string | undefined) {
```
hayes-mysten, line 22: I didn't know static private methods were a thing

hayes-mysten (approved):

packages/walrus/src/client.ts
williamrobertson13: yep going to do that in a follow-up, I wanted to keep this PR semi-mananagable

packages/walrus/src/client.ts
williamrobertson13: Discussed this offline - I'm actually going to address the secondary sliver writes and this mix-up in a different PR. I ran into some other things that might need a little tweaking that I want to run by you tomorrow

packages/walrus/src/storage-node/error.ts
williamrobertson13: me either 😆

packages/walrus/src/storage-node/client.ts
williamrobertson13: Ahhh this is a good call out, you think we should have a `timeout` client option for aborting requests that take longer than X amount of time? I'm not sure of a better way to handle this

packages/walrus/src/storage-node/client.ts
williamrobertson13: Yep will do this in a follow-up tomorrow - need to double check the header merging util bits but should be a quick fix

packages/walrus/src/storage-node/client.ts
williamrobertson13: Originally I wasn't returning anything bc the result is kind of useless, but I went ahead and returned the parsed response. I think I misunderstood the original Q - I thought you were saying what happens if the server doesn't consume a streamed request body and leaves this hanging forever

PR #54, opened 2025-02-04
[walrus] write secondary slivers to storage nodes

williamrobertson13:
## Description
We aren't mapping secondary slivers to the right storage node (_I think_), and we also aren't writing the secondary slivers at all atm.

## Test plan
- Eeeeh not really a great way to test this, but this matches my understanding from the whitepaper + codebase


packages/walrus/src/client.ts
```diff
@@ -575,44 +575,30 @@ export class WalrusClient {
...
 
 		const sliversByNodeMap = new Map<number, SliversForNode>();
 
 		while (sliverPairs.length > 0) {
 			// remove from list so we don't preserve references to the original data
 			const { primary, secondary } = sliverPairs.pop()!;
+			const shardIndex = toShardIndex(primary.index, blobId, numShards);
+			const node = await this.#getNodeByShardIndex(shardIndex);
 
-			const primaryShardIndex = toShardIndex(primary.index, blobId, systemState.committee.n_shards);
```
williamrobertson13, line 589: I don't think this logic is correct. When you encode a blob, you get a list of sliver pairs where index `i` refers to the i-th sliver pair. The sliver pair index `i` can be derived from either the primary or secondary sliver indices themselves, i.e. `sliver_pair_index = primary_sliver.index` or `sliver_pair_index = numShards - secondary_sliver.index - 1`.

So really what you want to do here is write both slivers to the shard derived from the sliver pair index for the given sliver pair. I think this logic is mistakenly converting the sliver index to a shard index (which is correct technically for the primary sliver but not for the secondary sliver).

[I just wrote this out to sanity check my own understanding, but if I'm wrong plz let me know]

image

packages/walrus/src/client.ts
```diff
@@ -575,44 +575,30 @@ export class WalrusClient {
...
+			sliversByNodeMap.get(node.nodeIndex)!.primary.push({
 				sliverIndex: primary.index,
-				shardIndex: primaryShardIndex,
+				shardIndex,
 				sliver: SliverData.serialize(primary).toBytes(),
 			});
 
-			sliversByNodeMap.get(secondaryNode.nodeIndex)!.secondary.push({
+			sliversByNodeMap.get(node.nodeIndex)!.secondary.push({
 				sliverIndex: secondary.index,
```
hayes-mysten, line 600: I think we should just replace this with sliverPairIndex so we don't have to calculate it later

hayes-mysten (approved):

packages/walrus/src/client.ts
hayes-mysten: I think just `sliverPairIndex: primary.index` should work here

PR #67, opened 2025-02-16
[walrus] make reads/writes more resilient during epoch changes

williamrobertson13:
## Description

This PR makes the read flow more resilient to epoch changes and adds some basic logic for doing retries when we detect an epoch change is progress. For a list of changes here:

- Reads should be served from the previous committee during an epoch change if the blob was certified earlier than the current epoch. This is because nodes from the current committee might still be receiving shards from the previous committee, so the previous committee has the most up-to-date state

- To get the epoch that a blob was certified at, we need to retrieve the _verified_ blob status from the storage nodes. There's quite a bit of complicated logic here, but basically you need to get quorum on blob status responses so you can be assured you have at least one honest node reporting the correct blob status. After getting quorum, you need to aggregate the unique status responses by aggregate weight and if you have `f + 1` total weight you know you have the verified status. There can be inconsistencies between storage nodes on the status of a blob, so you additionally need to sort the statuses by ordering of "latest in the blob lifecycle" to "earliest in the blob lifecycle" as the latest reported statuses are the more correct statuses

- This introduces a `retryOnPossibleEpochChange` util to the WalrusClient that'll nuke the object cache and retry a function once if the function failed possibly due to an epoch change. **Since the client is going to be long lived in most cases, we should add probabllllllly add a caching layer (or modify the underlying cache map in the data loader) in a follow-up to invalidate the Sui object cache leading up to epoch changes.**

## Test plan

- I tested the general logic locally with 30s epoch changes, but I haven't found a way to simulate changing committees or some of the edge cases that warrant doing all the complicated blob status retrieval logic. I'll ask the Walrus core folks for help with this

### DIFF
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -1,26 +1,43 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import type { InferBcsType } from '@mysten/bcs';
 import { bcs, fromBase64 } from '@mysten/bcs';
 import { SuiClient } from '@mysten/sui/client';
 import type { Signer } from '@mysten/sui/cryptography';
 import { coinWithBalance, Transaction } from '@mysten/sui/transactions';
 import { bls12381_min_pk_verify } from '@mysten/walrus-wasm';
 
-import { TESTNET_WALRUS_PACKAGE_CONFIG } from './constants.js';
+import { statusLifecycleRank, TESTNET_WALRUS_PACKAGE_CONFIG } from './constants.js';
 import { Blob } from './contracts/blob.js';
+import type { Committee } from './contracts/committee.js';
 import { StakingInnerV1 } from './contracts/staking_inner.js';
 import { StakingPool } from './contracts/staking_pool.js';
 import { Staking } from './contracts/staking.js';
 import { Storage } from './contracts/storage_resource.js';
 import { SystemStateInnerV1 } from './contracts/system_state_inner.js';
 import { init as initSystemContract, System } from './contracts/system.js';
+import {
+	BehindCurrentEpochError,
+	BlobNotCertifiedError,
+	NoBlobMetadataReceivedError,
+	NoBlobStatusReceivedError,
+	NotEnoughBlobConfirmationsError,
+	NotEnoughSliversReceivedError,
+	NoVerifiedBlobStatusReceivedError,
+	WalrusClientError,
+} from './error.js';
 import { StorageNodeClient } from './storage-node/client.js';
+import { NotFoundError, UserAbortError } from './storage-node/error.js';
+import type { BlobStatus } from './storage-node/types.js';
 import type {
 	CertifyBlobOptions,
+	CommitteeInfo,
 	DeleteBlobOptions,
 	ExtendBlobOptions,
+	GetCertificationEpochOptions,
 	GetStorageConfirmationOptions,
+	GetVerifiedBlobStatusOptions,
 	ReadBlobOptions,
 	RegisterBlobOptions,
 	SliversForNode,
@@ -39,6 +56,9 @@ import {
 	encodedBlobLength,
 	getPrimarySourceSymbols,
 	getShardIndicesByNodeId,
+	hash,
+	isAboveValidity,
+	isQuorum,
 	signersToBitmap,
 	storageUnitsFromSize,
 	toPairIndex,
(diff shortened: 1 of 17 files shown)


packages/walrus/src/utils/index.ts
```diff
@@ -105,3 +120,12 @@ export function nodesByShardIndex(committee: InferBcsType<ReturnType<typeof Comm
 
 	return nodesByShardIndex;
 }
+
+export async function hash(value: any) {
```
williamrobertson13, line 124: eh this kinda sucks to do but from my understanding nodes can have differing views on certification epochs, deletable counts, etc. so you really need to deep compare the status objects. I should probably also use a stable stringify util, but I figured this would suffice

(open to alternative ideas!)

packages/walrus/src/client.ts
```diff
@@ -790,34 +951,50 @@ export class WalrusClient {
...
+				return await fn.apply(this, args);
+			} catch (error) {
+				if (
+					error instanceof BlobNotCertifiedError ||
+					error instanceof BehindCurrentEpochError ||
+					error instanceof NoBlobMetadataReceivedError ||
+					error instanceof NotEnoughSliversReceivedError ||
+					error instanceof NotEnoughBlobConfirmationsError
+				) {
+					this.#objectLoader.clearAll();
```
hayes-mysten, line 993: I think we should reset #nodes too (and any other cached data), we should probably just have a method that resets all the client state

packages/walrus/src/client.ts
```diff
@@ -54,15 +74,7 @@ export class WalrusClient {
...
-	#nodes?:
-		| {
-				byShardIndex: Map<number, StorageNode>;
-				committee: StorageNode[];
-		  }
-		| Promise<{
-				byShardIndex: Map<number, StorageNode>;
-				committee: StorageNode[];
-		  }>;
+	activeCommittee?: CommitteeInfo | Promise<CommitteeInfo>;
```
hayes-mysten, line 77: this should probably be private

packages/walrus/src/client.ts
```diff
@@ -790,34 +951,50 @@ export class WalrusClient {
...
+		const node = committeeInfo.byShardIndex.get(index);
 		if (!node) {
-			throw new Error(`Node for shard index ${index} not found`);
+			throw new WalrusClientError(`Node for shard index ${index} not found`);
 		}
-
 		return node;
 	}
+
+	#retryOnPossibleEpochChange<T extends (...args: any[]) => any>(fn: T): T {
```
hayes-mysten, line 981:
```suggestion
	#retryOnPossibleEpochChange<T extends (...args: any[]) => Promise<any>>(fn: T): T {
```

packages/walrus/src/client.ts
```diff
@@ -164,12 +180,152 @@ export class WalrusClient {
...
+							}
+						});
+				});
+			},
+		);
+
+		const aggregatedStatuses = new Map<string, { status: BlobStatus; totalWeight: number }>();
+		for (const value of statuses) {
+			const { status, weight } = value;
+			const key = await hash(status);
```
Jordan-Mysten, line 255: - Why not just `status.type`?
- If that doesn't work not just `JSON.stringify()`? (this is smaller _end_ result but you need to do this compute anyway and you throw the result on the ground afterwards so this is literally just a worse version of that)

packages/walrus/src/client.ts
```diff
@@ -164,12 +180,152 @@ export class WalrusClient {
...
+				let successWeight = 0;
+				let numNotFoundWeight = 0;
+				let settledCount = 0;
+
+				committee.nodes.forEach((node) => {
+					const weight = node.shardIndices.length;
+
+					this.#storageNodeClient
+						.getBlobStatus({ blobId }, { nodeUrl: node.networkUrl, signal: controller.signal })
+						.then((status) => {
```
Jordan-Mysten, line 216: Why are you doing this instead of async / await?

packages/walrus/src/client.ts
williamrobertson13: oops thanks

packages/walrus/src/client.ts
williamrobertson13: 🧠 :hurtrealbad:  (fixed)

williamrobertson13:
> This looks pretty good, but it looks like we only ever reset state when reading blobs, do we want to do something similar during writes so that we can get an updated committee without reading?

yep so the only retry case for blob writes AFAIK is if sliver writes fail, I'll double check afterward to see if there's anything we should do if the metadata writes fail as well

PR #70, opened 2025-02-17
Add apple as an AuthProvider

woat:
## Description

I'm adding apple based on https://docs.sui.io/guides/developer/cryptography/zklogin-integration#get-jwt

## Test plan

I just ran this:

```
      const authUrl = await flow.createAuthorizationURL({
        provider: 'apple',
        network: 'mainnet',
        clientId: 'CLIENT_ID',
        redirectUrl: 'REDIRECT_URI',
      });
```

---

### DIFF
--- a/packages/enoki/src/EnokiFlow.ts
+++ b/packages/enoki/src/EnokiFlow.ts
@@ -50,7 +50,7 @@ export interface ZkLoginSession {
 	proof?: ZkLoginSignatureInputs;
 }
 
-export type AuthProvider = 'google' | 'facebook' | 'twitch';
+export type AuthProvider = 'google' | 'facebook' | 'twitch' | 'apple';
 
 const createStorageKeys = (apiKey: string) => ({
 	STATE: `@enoki/flow/state/${apiKey}`,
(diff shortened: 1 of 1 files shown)


packages/enoki/src/EnokiFlow.ts
```diff
@@ -154,6 +154,14 @@ export class EnokiFlow {
 				break;
 			}
 
+			case 'apple': {
+				params.set('response_type', 'code%20id_token')
+				params.set('scope', 'email')
+				params.set('response_mode', 'form_post')
```
Jordan-Mysten, line 160: I don’t think this response mode is compatible with most apps. We have plans to solve this eventually via a custom endpoint that can do the token exchange but haven’t built it out yet.

woat: image


So does apple auth work with Enoki at all? If so, is there some example flow I could look at?

Jordan-Mysten: It does, but it doesn’t work with the SDK. I’ll try and grab some resources around how to use it tomorrow and drop them here

woat: Ok thonks ser, I shall patiently wait for your response. 
image

woat: image

PR #71, opened 2025-02-18
[walrus] make blob metadata fetches more robust

williamrobertson13:
## Description
Storage nodes don't always contain the metadata for a blob, so this PR implements the same strategy the Rust client uses for metadata reads – we randomize the committee, try to fetch metadata from the first node and return if successful, otherwise start fetching in parallel until we get a successful response or reach quorum on the blob being blocked/not certified.

## Test plan
- Tested with reading a bunch of blobs

---

### DIFF
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -188,7 +188,7 @@ export class WalrusClient {
 	async getBlobMetadata({ blobId, signal }: GetBlobMetadataOptions) {
 		const controller = new AbortController();
 		signal?.addEventListener('abort', () => {
-			controller.abort();
+			controller.abort(signal.reason);
 		});
 
 		const committee = await this.#getReadCommittee({ blobId, signal });
@@ -268,7 +268,7 @@ export class WalrusClient {
 	async getVerifiedBlobStatus({ blobId, signal }: GetVerifiedBlobStatusOptions) {
 		const controller = new AbortController();
 		signal?.addEventListener('abort', () => {
-			controller.abort();
+			controller.abort(signal.reason);
 		});
 
 		// Read from the latest committee because, during epoch change, it is the committee
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -19,7 +19,9 @@ import { SystemStateInnerV1 } from './contracts/system_state_inner.js';
 import { init as initSystemContract, System } from './contracts/system.js';
 import {
 	BehindCurrentEpochError,
+	BlobBlockedError,
 	BlobNotCertifiedError,
+	NoBlobMetadataReceivedError,
 	NoBlobStatusReceivedError,
 	NotEnoughBlobConfirmationsError,
 	NotEnoughSliversReceivedError,
@@ -28,13 +30,14 @@ import {
 	WalrusClientError,
 } from './error.js';
 import { StorageNodeClient } from './storage-node/client.js';
-import { NotFoundError, UserAbortError } from './storage-node/error.js';
-import type { BlobStatus } from './storage-node/types.js';
+import { LegallyUnavailableError, NotFoundError, UserAbortError } from './storage-node/error.js';
+import type { BlobMetadataWithId, BlobStatus } from './storage-node/types.js';
 import type {
 	CertifyBlobOptions,
 	CommitteeInfo,
 	DeleteBlobOptions,
 	ExtendBlobOptions,
+	GetBlobMetadataOptions,
 	GetCertificationEpochOptions,
 	GetStorageConfirmationOptions,
 	GetVerifiedBlobStatusOptions,
@@ -53,6 +56,7 @@ import type {
 } from './types.js';
 import { blobIdToInt, IntentType, SliverData, StorageConfirmation } from './utils/bcs.js';
 import {
+	chunk,
 	encodedBlobLength,
 	getPrimarySourceSymbols,
 	getShardIndicesByNodeId,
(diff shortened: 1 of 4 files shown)


packages/walrus/src/utils/promise-queue.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export class PromiseQueue<T> {
```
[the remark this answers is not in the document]
hayes-mysten: can we copy the one from the ts sdk:

```ts
export class ParallelQueue {
	#queue: Array<() => void> = [];
	activeTasks = 0;
	maxTasks: number;

	constructor(maxTasks: number) {
		this.maxTasks = maxTasks;
	}

	runTask<T>(task: () => Promise<T>): Promise<T> {
		return new Promise<T>((resolve, reject) => {
			if (this.activeTasks < this.maxTasks) {
				this.activeTasks++;

				task()
					.finally(() => {
						if (this.#queue.length > 0) {
							this.#queue.shift()!();
						} else {
							this.activeTasks--;
						}
					})
					.then(resolve, reject);
			} else {
				this.#queue.push(() => {
					task()
						.finally(() => {
							if (this.#queue.length > 0) {
								this.#queue.shift()!();
							} else {
								this.activeTasks--;
							}
						})
						.then(resolve, reject);
				});
			}
		});
	}
}
```

I am more confident that that one actually works correctly (unless this is already well tested)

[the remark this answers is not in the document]
hayes-mysten: these look like ALMOST the same thing.  I do like the naming in the new version better

packages/walrus/src/utils/promise-queue.ts
```diff
@@ -0,0 +1,36 @@
...
+						this.#processQueue();
+					});
+			});
+
+			this.#processQueue();
+		});
+	}
+
+	#processQueue() {
+		while (this.#pending < this.#maxConcurrency && this.#queue.length > 0) {
```
hayes-mysten, line 31: I don't think its possible for this loop to ever iterate more than once

packages/walrus/src/client.ts
```diff
@@ -308,6 +306,91 @@ export class WalrusClient {
...
+				maxConcurrency: this.#blobMetadataMaxConcurrencyLimit,
+			});
+
+			return new Promise<BlobMetadataWithId>((resolve, reject) => {
+				let settledCount = 0;
+
+				executors.forEach((executor) => {
+					queue
+						.add(executor)
+						.then((result) => {
```
Jordan-Mysten, line 356: Why this instead of async / await?

packages/walrus/src/client.ts
```diff
@@ -308,6 +306,91 @@ export class WalrusClient {
...
+										`The specified blob ${blobId} is blocked.`,
+									);
+									controller.abort(abortError);
+									reject(abortError);
+								}
+							}
+						})
+						.finally(() => {
+							settledCount += 1;
+							if (settledCount === executors.length) {
```
Jordan-Mysten, line 381: You can avoid this logic by making the `forEach` a `map` and then awaiting a `Promise.all` of it that checks if the promise has been settled or not. Seems easier to reason about to me.

packages/walrus/src/utils/promise-queue.ts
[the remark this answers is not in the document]
williamrobertson13: discussed offline - not going this route

packages/walrus/src/client.ts
```diff
@@ -308,6 +306,91 @@ export class WalrusClient {
 		return currentEpoch;
 	}
 
+	async getBlobMetadata({ blobId, signal }: GetBlobMetadataOptions) {
+		const controller = new AbortController();
+		signal?.addEventListener('abort', () => {
+			controller.abort();
+		});
+
+		const committee = await this.#getReadCommittee({ blobId, signal });
```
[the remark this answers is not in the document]
williamrobertson13: cached this now

packages/walrus/src/client.ts
```diff
@@ -185,6 +185,83 @@ export class WalrusClient {
...
+			return await attemptGetMetadata();
+		} catch (error) {
+			const chunkSize = Math.floor(metadataExecutors.length / this.#blobMetadataConcurrencyLimit);
+			const chunkedExecutors = chunk(metadataExecutors, chunkSize);
+
+			return await new Promise<BlobMetadataWithId>((resolve, reject) => {
+				chunkedExecutors.forEach(async (executors) => {
+					for (const executor of executors) {
+						try {
+							const result = await executor();
```
hayes-mysten, line 233: should this `controller.throwIfAborted()` before executing?

packages/walrus/src/client.ts
```diff
@@ -185,6 +185,83 @@ export class WalrusClient {
 		return decodePrimarySlivers(numShards, blobMetadata.metadata.V1.unencoded_length, slivers);
 	}
 
+	async getBlobMetadata({ blobId, signal }: GetBlobMetadataOptions) {
+		const controller = new AbortController();
+		signal?.addEventListener('abort', () => {
+			controller.abort();
```
hayes-mysten, line 191: can we bubble the original abort error here?

```ts
signal?.addEventListener('abort', () => {
			controller.abort(signal.reason);
		});
		```

hayes-mysten (approved):

packages/walrus/src/client.ts
williamrobertson13: the fetch client should take care of this internally!

https://github.com/MystenLabs/ts-sdks/blob/wrobertson/retrieve_metadata/packages/walrus/src/storage-node/client.ts#L227

hayes-mysten: ah, nice!

PR #72, opened 2025-02-18
[walrus] add default timeout configuration option to the http client

williamrobertson13:
## Description

This is just some low-hanging fruit to add default timeouts to the storage node HTTP client. I made the default 30s which seems reasonable? Might up to 60s if this ends up being too low

## Test plan
- Tested locally with a timeout of 20ms
image

### DIFF
--- a/packages/walrus/src/storage-node/client.ts
+++ b/packages/walrus/src/storage-node/client.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { BlobMetadata, BlobMetadataWithId, SliverData } from '../utils/bcs.js';
-import { StorageNodeAPIError, UserAbortError } from './error.js';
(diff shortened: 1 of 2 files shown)


packages/walrus/src/storage-node/client.ts
```diff
@@ -30,28 +30,37 @@ export type StorageNodeClientOptions = {
 	 * @default globalThis.fetch
 	 */
 	fetch?: Fetch;
+
+	/**
+	 * An optional timeout for requests.
+	 * @default 30_000ms (3 seconds)
```
hayes-mysten, line 36: 30 or 3 seconds?

williamrobertson13: oops 30 seconds - fixed 😆

hayes-mysten (approved):

PR #75, opened 2025-02-24
[walrus][out on PTO - plz take over] make sliver reads more robust

williamrobertson13:
## Description

This PR makes the sliver fetching more robust, specifically:
  - We want to try to fetch slivers from nodes at random proportional to how many shards they possess
  - If we fail fetching from a given node, we skip subsequent requests to that node
  - If we hit too many 404/451 errors, we bail out if quorum was reached
  - If we hit so many errors so that we can't fetch `minSymbols`, we bail out

I'm out all this week on PTO, so just getting this up and if there's a more preferred way to go about the async control flow management here please be my guest! I just used a simple chunking approach where we execute against the chunked sliver indices in column-wise order so that we preserve the weighted random ordering (felt kinda meh about this but it works). 

## Test plan
---
- Tested reads on a bunch of blobs

### DIFF
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -31,7 +31,7 @@ import {
 } from './error.js';
 import { StorageNodeClient } from './storage-node/client.js';
 import { LegallyUnavailableError, NotFoundError, UserAbortError } from './storage-node/error.js';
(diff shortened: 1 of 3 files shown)


hayes-mysten (approved):
Updated the weightedShuffle function, but looks good to me.  Going to merge it as is, and we can add any improvements as a follow up

PR #77, opened 2025-02-25
[walrus] fix versioned contract

hayes-mysten:
## Description

## Test plan

---


packages/walrus/src/client.ts
```diff
@@ -123,7 +123,7 @@ export class WalrusClient {
 	}
 
 	get systemContract() {
-		return initSystemContract(this.packageConfig.packageId);
+		return initSystemContract(this.packageConfig.latestPackageId);
```
hayes-mysten, line 126: We'll want to replace this with mover eventually, but I think this is fine for now

packages/walrus/src/client.ts
```diff
@@ -173,11 +173,6 @@ export class WalrusClient {
 	}
 
 	async getBlobMetadata({ blobId, signal }: GetBlobMetadataOptions) {
-		const controller = new AbortController();
-		signal?.addEventListener('abort', () => {
-			controller.abort(signal.reason);
```
hayes-mysten, line 178: doing this results in warnings in node when we add too many listeners to one signal

packages/walrus/examples/aggregator/index.ts
```diff
@@ -26,13 +26,26 @@ const cache = new Map<string, Blob>();
...
+
+	if (cache.has(blobId)) {
+		return c.body(cache.get(blobId)!.stream());
+	}
+
+	try {
+		var blob = await walrusClient.readBlob({ blobId });
 		cache.set(blobId, new Blob([blob]));
+	} catch (error) {
+		if (error instanceof NotFoundError) {
```
williamrobertson13, line 41: I think you want to actually catch `BlobNotCertifiedError` / `BlobBlockedError` as `NotFoundError` is the internal 404 error returned from the storage nodes

Side note: The naming of `BlobNotCertifiedError` is maybe a little meh, maybe we should just rename this to `BlobNotFoundError` to match the Rust codebase. The meaning of this error is a little weird because it means the blob exists but isn't certified on Walrus or just plainly doesn't exist. Glad to take care of this in a follow-up PR tho

packages/walrus/src/client.ts
williamrobertson13: Thanks for fixing this, did not know `AbortSignal.any` existed

williamrobertson13 (approved):

PR #79, opened 2025-02-26
walrus blob attributes

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/walrus/src/client.ts
+++ b/packages/walrus/src/client.ts
@@ -5,12 +5,14 @@ import type { InferBcsType } from '@mysten/bcs';
 import { bcs, fromBase64 } from '@mysten/bcs';
 import { SuiClient } from '@mysten/sui/client';
 import type { Signer } from '@mysten/sui/cryptography';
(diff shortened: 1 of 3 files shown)


packages/walrus/src/client.ts
```diff
@@ -996,6 +1019,158 @@ export class WalrusClient {
...
+						],
+					}),
+				);
+			}
+
+			Object.keys(attributes).forEach((key) => {
+				const value = attributes[key];
+
+				if (value === null) {
+					if (existingAttributes && existingAttributes[key]) {
```
williamrobertson13, line 1091: not sure if there's a super tiny edge case here but might be safer to do `key in existingAttributes`

williamrobertson13 (approved):
lgtm! One thing to maybe think about - should we have `readBlob` return a `Blob` instead of a `UInt8Array` with the `type` automatically inferred from the blob attributes if a `contentType` attribute exists?

hayes-mysten:
going to punt the returning Blob idea to the future.  We've mostly stuck to Uint8Arrays everywhere in our SDKs so far, and I want to be careful about what primitives we use

PR #82, opened 2025-02-27
Add check for malformed google iss in toZkLoginPublicIdentifier

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/typescript/src/zklogin/jwt-utils.ts
+++ b/packages/typescript/src/zklogin/jwt-utils.ts
@@ -1,6 +1,10 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import type { JwtPayload } from './jwt-decode.js';
+import { jwtDecode } from './jwt-decode.js';
+import { normalizeZkLoginIssuer } from './utils.js';
+
 function base64UrlCharTo6Bits(base64UrlChar: string): number[] {
 	if (base64UrlChar.length !== 1) {
 		throw new Error('Invalid base64Url character: ' + base64UrlChar);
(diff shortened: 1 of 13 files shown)


packages/typescript/src/zklogin/publickey.ts
```diff
@@ -142,6 +142,10 @@ export function toZkLoginPublicIdentifier(
 		? toBigEndianBytes(addressSeed, 32)
 		: toPaddedBigEndianBytes(addressSeed, 32);
 
+	if (iss === 'accounts.google.com') {
```
williamrobertson13, line 145: tiny nit: would it be worth making a tiny `normalizeIssuer` util so we can use it here and in https://github.com/MystenLabs/ts-sdks/blob/03975f4ade68ae5f4206c2b0818bafe8a7de071e/packages/typescript/src/zklogin/address.ts#L21?

williamrobertson13 (approved):

Jordan-Mysten (commented):
Do we want to rework the JWT decoding in a separate PR?

packages/typescript/src/zklogin/jwt-utils.ts
```diff
@@ -111,3 +115,24 @@ export function extractClaimValue<R>(claim: Claim, claimName: string): R {
...
+		throw new Error('Missing jwt data');
+	}
+
+	if (Array.isArray(aud)) {
+		throw new Error('Not supported aud. Aud is an array, string was expected.');
+	}
+
+	return {
+		...decodedJWT,
+		iss: normalizeZkLoginIssuer(iss),
```
Jordan-Mysten, line 134: Should we also add something like `rawIss`?

packages/typescript/src/zklogin/publickey.ts
```diff
@@ -142,7 +142,7 @@ export function toZkLoginPublicIdentifier(
 		? toBigEndianBytes(addressSeed, 32)
 		: toPaddedBigEndianBytes(addressSeed, 32);
 
-	const issBytes = new TextEncoder().encode(iss);
+	const issBytes = new TextEncoder().encode(normalizeZkLoginIssuer(iss));
```
Jordan-Mysten, line 145: This shouldn't be needed right?

Jordan-Mysten: Unless this is just for safety

Jordan-Mysten (approved):

packages/typescript/src/zklogin/publickey.ts
hayes-mysten: yeah, for safety incase someone decodes the jwt and calls this directly

PR #83, opened 2025-02-28
initialize enoki-connect wallet-standard

pchrysochoidis:
## Description

Implements the wallet standard for EnokiConnect wallet

## Test plan

---

### DIFF
--- a/packages/enoki-connect/src/index.ts
+++ b/packages/enoki-connect/src/index.ts
@@ -1,0 +1,8 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export {
+	EnokiConnectWallet,
+	registerEnokiConnectWallet,
+	type SupportedNetwork,
+} from './wallet/index.js';
--- a/packages/enoki-connect/src/wallet/index.ts
+++ b/packages/enoki-connect/src/wallet/index.ts
@@ -1,0 +1,366 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Transaction } from '@mysten/sui/transactions';
+import { fromBase64, toBase64 } from '@mysten/sui/utils';
+import type {
+	IdentifierString,
+	StandardConnectFeature,
+	StandardConnectMethod,
+	StandardDisconnectFeature,
+	StandardDisconnectMethod,
+	StandardEventsFeature,
+	StandardEventsListeners,
+	StandardEventsOnMethod,
+	SuiSignAndExecuteTransactionFeature,
+	SuiSignAndExecuteTransactionMethod,
+	SuiSignPersonalMessageFeature,
+	SuiSignPersonalMessageMethod,
+	SuiSignTransactionBlockFeature,
+	SuiSignTransactionBlockMethod,
+	SuiSignTransactionFeature,
+	SuiSignTransactionMethod,
+	Wallet,
+	WalletIcon,
+} from '@mysten/wallet-standard';
+import {
+	getWallets,
+	ReadonlyWalletAccount,
+	SUI_MAINNET_CHAIN,
+	SUI_TESTNET_CHAIN,
+} from '@mysten/wallet-standard';
+import type { Emitter } from 'mitt';
+import mitt from 'mitt';
+import { DappPostMessageChannel, decodeJwtSession } from '@mysten/window-wallet-core';
+
+export type SupportedNetwork = 'mainnet' | 'testnet';
+
+type WalletEventsMap = {
+	[E in keyof StandardEventsListeners]: Parameters<StandardEventsListeners[E]>[0];
+};
+
+const SUPPORTED_CHAINS = [SUI_MAINNET_CHAIN, SUI_TESTNET_CHAIN] as const;
+const ACCOUNT_FEATURES = [
+	'sui:signTransaction',
+	'sui:signAndExecuteTransaction',
+	'sui:signPersonalMessage',
+	'sui:signTransactionBlock',
+	'sui:signAndExecuteTransactionBlock',
+] as const;
+
+export class EnokiConnectWallet implements Wallet {
+	readonly id: string;
+	#events: Emitter<WalletEventsMap>;
+	#accounts: ReadonlyWalletAccount[];
+	#hostOrigin: string;
+	#walletName: string;
+	#dappName: string;
+	#icon: WalletIcon;
+	#defaultChain: IdentifierString;
+	#publicAppSlug: string;
+
+	get name() {
+		return this.#walletName;
+	}
+
+	get icon() {
+		return this.#icon;
+	}
+
+	get version() {
+		return '1.0.0' as const;
+	}
+
+	get chains() {
+		return SUPPORTED_CHAINS;
+	}
+
+	get accounts() {
+		return this.#accounts;
+	}
+
+	get features(): StandardConnectFeature &
+		StandardDisconnectFeature &
+		StandardEventsFeature &
+		SuiSignTransactionBlockFeature &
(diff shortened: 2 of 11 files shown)


packages/enoki-connect/src/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export * from './wallet/index.js';
```
hayes-mysten, line 4: I've been trying to move away from `export *` for SDKs.  We should try to be intentional about our public API surface, and this makes it too easy to export internal functionality that we then need to maintain for a long time

packages/enoki-connect/src/wallet/channel/events.ts
```diff
@@ -0,0 +1,77 @@
...
+	payload: EnokiConnectRequestData,
+});
+
+export type EnokiConnectRequest = InferOutput<typeof EnokiConnectRequest>;
+
+export const EnokiConnectResponseData = variant('type', [
+	object({
+		type: literal('connect'),
+		address: string(),
+		publicKey: string(),
```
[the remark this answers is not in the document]
hayes-mysten: I'd also be curious if we could share the actual implementation.  I'm not sure if its a good idea, but I'd like to at least explore that as an option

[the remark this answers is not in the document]
pchrysochoidis: We discussed that offline but also during a standup. TLDR probably better to keep them separate since web wallet will have more/different functionality as well. cc @hayes-mysten @Jordan-Mysten

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -0,0 +1,321 @@
...
+		account,
+		chain,
+	}) => {
+		transactionBlock.setSenderIfNotSet(account.address);
+
+		const popup = this.#getNewPopupChannel();
+		const response = await popup.send({
+			type: 'sign-transaction',
+			chain,
+			transaction: JSON.stringify(transactionBlock.getData()),
```
hayes-mysten, line 144: We should be using `await transaction.toJson({ client })` this will break with any async thunks or intents

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -0,0 +1,321 @@
...
+			transactionBlockBytes: response.bytes,
+			signature: response.signature,
+		};
+	};
+
+	#signTransaction: SuiSignTransactionMethod = async ({ transaction, account, chain }) => {
+		const popup = this.#getNewPopupChannel();
+		const tx = Transaction.from(await transaction.toJSON());
+
+		tx.setSenderIfNotSet(account.address);
```
hayes-mysten, line 159: This looks like it was coppied from an older version of the @claykohut-mysten's PR.  There are several changes we discussed there before it got merged that haven't made it into this version

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -154,38 +161,60 @@ export class EnokiConnectWallet implements Wallet {
...
-			session: localStorage.getItem(this.#getSessionKey()) ?? '',
+			chain,
+			session: this.#getStoredSession(),
 		});
 
 		return {
 			bytes: response.bytes,
 			signature: response.signature,
+			digest: response.digest,
+			effects: response.effects || '',
```
hayes-mysten, line 201: This doesn't seem like a safe thing to default to an empty string, this will just cause it to fail when the we attempt to parse the effects

packages/enoki-connect/src/wallet/index.ts
pchrysochoidis: I updated this based on the [latest StashedWallet](https://github.com/MystenLabs/ts-sdks/blob/main/packages/zksend/src/experimental/wallet/index.ts#L64). Is there another pr?

packages/enoki-connect/src/wallet/index.ts
pchrysochoidis: Hmm should I make it required in the response then?

Otherwise should I default to an empty object instead? What's best?

packages/enoki-connect/src/wallet/index.ts
hayes-mysten: I was saying it's easier to see the actual changes if this is all in one PR.  This is essentially just moving the implementation from one place to another, so the diff should show that (+ any changes added to make it generic). It's a lot harder to reason about as separate PRs since it all looks like new code

hayes-mysten: I got my PRs mixed up sorry. I thought this was a comment on the wallet-core PR

packages/enoki-connect/src/wallet/index.ts
hayes-mysten: I think required in the response is probably the right answer

packages/enoki-connect/src/wallet/index.ts
pchrysochoidis: I updated this but just a note we don't have access to a client so I just use `.toJSON()`.

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -0,0 +1,366 @@
...
+		this.#publicAppSlug = publicAppSlug;
+		this.id = `enoki-connect-${publicAppSlug}`;
+	}
+
+	#signTransactionBlock: SuiSignTransactionBlockMethod = async ({
+		transactionBlock,
+		account,
+		chain,
+	}) => {
+		transactionBlock.setSenderIfNotSet(account.address);
```
hayes-mysten, line 152: This still looks like it's out of sync with what we are doing for slush: https://github.com/MystenLabs/ts-sdks/blob/main/packages/slush-wallet/src/wallet/index.ts#L205-L221

We shouldn't need to set the sender or clone the transaction here

hayes-mysten, line 152: We shouldn't need to mutation the transaction, this can work like it does here: https://github.com/MystenLabs/ts-sdks/blob/main/packages/slush-wallet/src/wallet/index.ts#L206-L222

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -0,0 +1,366 @@
...
+			effects: response.effects,
+		};
+	};
+
+	#signPersonalMessage: SuiSignPersonalMessageMethod = async ({ message, account }) => {
+		const popup = this.#getNewPopupChannel();
+		const response = await popup.send({
+			type: 'sign-personal-message',
+			// TODO: use chain from signPersonalMessage input when available
+			chain: this.#defaultChain,
```
hayes-mysten, line 217: looks like this was added to the standard so we can update this

hayes-mysten (approved):

packages/enoki-connect/src/wallet/index.ts
pchrysochoidis: Done

packages/enoki-connect/src/wallet/index.ts
pchrysochoidis: Done

packages/enoki-connect/src/wallet/index.ts
```diff
@@ -0,0 +1,366 @@
...
+		};
+	};
+
+	#signAndExecuteTransaction: SuiSignAndExecuteTransactionMethod = async ({
+		transaction,
+		account,
+		chain,
+	}) => {
+		const popup = this.#getNewPopupChannel();
+		const tx = Transaction.from(await transaction.toJSON());
```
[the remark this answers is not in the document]
pchrysochoidis: Done - I updated slush here https://github.com/MystenLabs/ts-sdks/pull/240 as well

PR #84, opened 2025-03-03
add enoki connect in wallet provider

pchrysochoidis:
## Description

## Test plan

---

### DIFF
--- a/packages/dapp-kit/src/components/EnokiConnectWalletManager.tsx
+++ b/packages/dapp-kit/src/components/EnokiConnectWalletManager.tsx
@@ -1,0 +1,83 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { SupportedNetwork } from '@mysten/enoki-connect';
+import { registerEnokiConnectWallet } from '@mysten/enoki-connect';
+import { useEffect } from 'react';
+
+export type EnokiConnectWalletConfig = {
+	publicAppSlug: string;
+	dappName: string;
+	enokiApiUrl?: string;
+	network?: SupportedNetwork;
+};
+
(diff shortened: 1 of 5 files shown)


packages/dapp-kit/src/components/EnokiConnectWalletManager.tsx
```diff
@@ -0,0 +1,70 @@
...
+	dappName: string;
+};
+
+export function EnokiConnectWalletManager({
+	appIds,
+	enokiApiUrl,
+	network,
+	dappName,
+}: EnokiConnectWalletManagerProps) {
+	const uniqueAppIds = new Set(appIds);
```
hayes-mysten, line 27: should this be memoized? it's probably fine as is, but creating data structures in render feels wrong

hayes-mysten (approved):

packages/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -42,6 +44,9 @@ export type WalletProviderProps = {
 	/** Enables the Stashed wallet */
 	stashedWallet?: StashedWalletConfig;
 
+	/** Enables Enoki Connect wallets */
```
williamrobertson13, line 47: Hmmm so our stance has been that we don't include Enoki-related features into dApp Kit. Since we're making dApp Kit framework agnostic, I think it would be best to actually not ship this PR and follow suit with what we do in https://docs.enoki.mystenlabs.com/ts-sdk/register#react-integration. I think there are probably some ways to make the interop here smoother, but I'm not quite there yet 😅 

cc @hayes-mysten @Jordan-Mysten

pchrysochoidis:
Closing this in favour of an approach like this -> https://docs.enoki.mystenlabs.com/ts-sdk/register#react-integration

PR #92, opened 2025-03-08
update `SuiMoveNormalizedEnum.variants` type to vector

vladilen11:
## Description

update `SuiMoveNormalizedEnum.variants` type to vector

## Test plan

Need to make sure that the correct data is queried after the enum type is updated.

---

relate:
https://github.com/MystenLabs/sui/issues/20914
https://github.com/MystenLabs/sui/pull/21359/files

### DIFF
--- a/packages/graphql-transport/src/mappers/move.ts
+++ b/packages/graphql-transport/src/mappers/move.ts
@@ -325,7 +325,7 @@ export function mapNormalizedMoveEnum(
 						) ?? [],
 				},
 			})) ?? [],
-		variants: Object.fromEntries(
+		variants:
 			enumNode.variants?.map((variant) => [
 				variant.name,
 				variant.fields?.map((field) => ({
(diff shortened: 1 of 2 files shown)


hayes-mysten:
Thanks for updating this! I think we will also need to update the generates JSON RPC types as well

hayes-mysten:
I've updated all the types in the sui client to match the current JSON rpc spec, but the change you were adding has been reverted: https://github.com/MystenLabs/sui/pull/21359

If this change is added back into JSON rpc, we can re-run the generator to update the types in the SDK.  I don't work on the JSON rpc codebase itself, so I am not sure what the future plans are for handling ordering of enum values.

hayes-mysten:
Sorry, I linked to the wrong PR, here is the revert I mentioned: https://github.com/MystenLabs/sui/pull/21414

PR #98, opened 2025-03-11
[Unification] Experimental subpackage for stashed web adapter

claykohut-mysten:
## Description

- Adds `experimental` sub package (eventually we'll break this out into a seperate package, but naming is currently up in the air. This is meant to unblock us for now)
- Adds new functionality for web wallet

## Test plan

TBD

---


packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
...
+		session: string('`session` is required'),
+		network: string('`network` is required'),
+	}),
+]);
+export type StashedRequestData = InferOutput<typeof StashedRequestData>;
+
+export const StashedRequest = object({
+	requestId: pipe(string('`requestId` is required'), uuid()),
+	appOrigin: pipe(string(), url('`appOrigin` must be a valid URL')),
+	appName: optional(string()),
```
hayes-mysten, line 32: I think we had made this option to make it easier to enable this by default.  Are we still comfortable with potentially unknown apps connecting on the stashed side?

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,437 @@
...
+	}
+}
+
+const embedStashedIframe = () => {
+	/* @ts-ignore */
+	embeddedIframe = document.createElement('iframe');
+	embeddedIframe.style.display = 'none';
+	embeddedIframe.src = `${stashedWalletOrigin}/embed`;
+	document.body.appendChild(embeddedIframe);
+	// every 3 seconds, check if the wallet is connected
```
hayes-mysten, line 367: This comment says 3 seconds, but the interval is 1 second, is there also an interval on the stashed side?

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,437 @@
...
+		origin,
+		network = 'mainnet',
+	}: {
+		origin?: string;
+		network?: StashedSupportedNetwork;
+	} = {},
+) {
+	const wallets = getWallets();
+
+	let addressFromRedirect: string | null = null;
```
hayes-mysten, line 412: is this supposed to be set somewhere?

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { InferOutput } from 'valibot';
+import { array, literal, object, optional, pipe, string, url, uuid, variant } from 'valibot';
+
+export const StashedRequestData = variant('type', [
+	object({
+		type: literal('connect'),
+		network: string('`network` is required'),
```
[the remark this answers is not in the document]
pchrysochoidis: I was thinking about removing it actually! 😅  and use the chain identifier to derive it (for signing etc.) I ended up keeping it to make the flow a bit better with EnokiFLow that currently does't handle very well network changes.

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: Updated comment to say 1 second. There isn't an interval on the stashed side, we just respond to the status request as it comes in

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: Looks like this is carry over from the old package -- not sure if we need to handle the redirect edge case [referenced here](https://github.com/MystenLabs/ts-sdks/blob/main/packages/dapp-kit/src/hooks/wallet/useStashedWallet.ts#L24), added back for now until we discuss

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
...
+	appName: optional(string()),
+	payload: StashedRequestData,
+});
+
+export type StashedRequest = InferOutput<typeof StashedRequest>;
+
+export const StashedResponseData = variant('type', [
+	object({
+		type: literal('connect'),
+		selectedAddresses: array(string()),
```
hayes-mysten, line 41: do we need `selected` here?  we do address selection in the Sui wallet on connect, but that is more of an implementation detail.  We can probably just call this addresses.  

For future proofing, we could also potentially make this something like `accounts: { address: string }[]`.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,444 @@
...
+	};
+
+	#setAccount(address?: string) {
+		if (address) {
+			this.#accounts = [
+				new ReadonlyWalletAccount({
+					address,
+					chains: [SUI_MAINNET_CHAIN],
+					features: ['sui:signTransactionBlock', 'sui:signPersonalMessage'],
+					// NOTE: Stashed doesn't support getting public keys, and zkLogin accounts don't have meaningful public keys anyway
```
hayes-mysten, line 231: This shouldn't be true anymore, we should return public keys on connect

packages/zksend/src/experimental/wallet/channel/events.ts
hayes-mysten: Actually, we definitely need to do make this an object.  We need to return publicKeys as part of the connect response

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
...
+		session: string(),
+	}),
+	object({
+		type: literal('sign-transaction-block'),
```
hayes-mysten, line 45: @Jordan-Mysten do we want to figure out sign-and-execute as part of this PR?  If we only do signing, that makes congestion pricing for gas a problem

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,444 @@
...
+
+	#signTransactionBlock: SuiSignTransactionBlockMethod = async ({ transactionBlock, account }) => {
+		transactionBlock.setSenderIfNotSet(account.address);
+
+		const data = transactionBlock.serialize();
+
+		const popup = new StashedPopup({
+			name: this.#name,
+			origin: this.#origin,
+			network: this.#network,
```
hayes-mysten, line 154: ignoring `chain` here feels kinda problematic.   My preference would be for wallet implementations to be multi-network.  If we are going to tie the wallet to a specific network, we need to at least throw an error here when chain is set incorrectly

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,444 @@
...
+			bytes: response.bytes,
+			signature: response.signature,
+		};
+	};
+
+	#signPersonalMessage: SuiSignPersonalMessageMethod = async ({ message, account }) => {
+		const popup = new StashedPopup({
+			name: this.#name,
+			origin: this.#origin,
+			network: this.#network,
```
hayes-mysten, line 201: same comment here, we have added network to the signPersonalMessage options in the wallet standard to acc ount for zklogin signatures.  It's optional, so we probably need to treat it slightly differently than we do when signing transactions, but we should be checking that field

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,444 @@
...
+					publicKey: new Uint8Array(),
+				}),
+			];
+		} else {
+			this.#accounts = [];
+		}
+
+		this.#events.emit('change', { accounts: this.accounts });
+	}
+	addAccount(address: string) {
```
hayes-mysten, line 241: Is this unused?

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,444 @@
...
+				}),
+			);
+		}
+
+		this.#accounts = this.#accounts.filter((account) => account.address !== address);
+
+		this.#events.emit('change', { accounts: this.accounts });
+	}
+
+	#setMultipleAccounts(addresses: string[]) {
```
hayes-mysten, line 274: There are 3 nearly identical version of setting the active accounts.  Can we combine these into one?  I think always setting to a list of accounts makes sense?

packages/zksend/src/experimental/wallet/channel/events.ts
Jordan-Mysten: Yeah we should do both

packages/zksend/src/experimental/wallet/index.ts
Jordan-Mysten: The signing flows on wallet support chain so we should be able to pass it here

packages/zksend/src/experimental/wallet/channel/events.ts
Jordan-Mysten: Let's make this required, we can default it to an empty string worst case.

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
...
+import { array, literal, object, optional, pipe, string, url, uuid, variant } from 'valibot';
+
+export const StashedRequestData = variant('type', [
+	object({
+		type: literal('connect'),
+		network: string('`network` is required'),
+	}),
+	object({
+		type: literal('sign-transaction-block'),
+		data: string('`data` is required'),
```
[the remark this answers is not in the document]
claykohut-mysten: `data` is what's used in zksend sdk today https://github.com/MystenLabs/ts-sdks/blob/main/packages/zksend/src/wallet/channel/events.ts#L13

Can update to something more descriptive

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,81 @@
...
+	appName: string('`appName` is required'),
+	payload: StashedRequestData,
+});
+
+export type StashedRequest = InferOutput<typeof StashedRequest>;
+
+export const StashedResponseData = variant('type', [
+	object({
+		type: literal('connect'),
+		addresses: array(string('`addresses` is required')),
```
hayes-mysten, line 41: I think this got lost in the diff, but I think we need this to be something like:

```ts
accounts: { address: string, publicKey: string }[]
```

packages/zksend/src/experimental/wallet/channel/events.ts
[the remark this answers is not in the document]
Jordan-Mysten: Maybe just transaction, cc @hayes-mysten for thoughts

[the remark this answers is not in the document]
hayes-mysten: `transaction` makes sense to me

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,413 @@
...
+		const popup = new StashedPopup({
+			name: this.#name,
+			origin: this.#origin,
+			network: this.#network,
+		});
+
+		const tx = Transaction.from(await transaction.toJSON());
+		tx.setSenderIfNotSet(account.address);
+
+		const data = tx.serialize();
```
hayes-mysten, line 181: serialize is deprecated, we should use the updated JSON format here

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export type StashedRequest = InferOutput<typeof StashedRequest>;
+
+export const StashedResponseData = variant('type', [
+	object({
+		type: literal('connect'),
+		accounts: array(
+			object({
+				address: string('`address` is required'),
+				publicKey: optional(instance(Uint8Array, '`publicKey` must be a Uint8Array')),
```
hayes-mysten, line 56: Does this work across the message boundary?  I haven't looked at this closely, I am guessing it might work because it uses structedClone internally? 

It might be worth making this something that can be serialized to JSON so we can change the transport without changing the types in the future (by making bytes a base64 string)

claykohut-mysten: yep on it!

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+			session: getStashedSession().token,
+		});
+
+		return {
+			transactionBlockBytes: response.bytes,
+			signature: response.signature,
+		};
+	};
+
+	#signTransaction: SuiSignTransactionMethod = async ({ transaction, account, chain }) => {
```
claykohut-mysten, line 182: I'm not entirely clear where this `chain` comes from  -- can I set this from a dapp somehow? 

In my test dapp I need to call signTransaction with testnet, but i'm always receiving "mainnet" here

packages/zksend/experimental/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
Jordan-Mysten, line 1: I think you can remove this since the `packages/zksend/experimental/unified-adapter/package.json` file is the correct one

packages/zksend/pnpm-lock.yaml
Jordan-Mysten, line 1: There shouldn't be a separate pnpm-lock here.

packages/zksend/src/experimental/wallet/types.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export type StashedSupportedNetwork = 'mainnet' | 'testnet';
```
Jordan-Mysten, line 4: you should be able to remove this

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+		};
+	}
+
+	constructor({
+		name,
+		address,
+		origin = DEFAULT_STASHED_ORIGIN,
+	}: {
+		name: string;
+		network: StashedSupportedNetwork;
```
Jordan-Mysten, line 139: This shouldn't be needed anymore

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	const { accounts = [], token } = JSON.parse(localStorage.getItem(STASHED_SESSION_KEY) || '{}');
+	return { accounts, token };
+};
+
+const getPostMessagePayload = () => {
+	return {
+		version: PACKAGE_VERSION,
+		origin_url: window.location.href,
+		user_agent: navigator.userAgent,
+		screen_resolution: `${window.screen.width}x${window.screen.height}`,
```
Jordan-Mysten, line 60: Honestly not a bad idea to add all of these properties to the standard protocol as optional metadata... Maybe just a `metadata: record<string, primitive>`? @hayes-mysten I think we talked about this before any objections?

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+
+export const STASHED_WALLET_NAME = 'Stashed' as const;
+type StashedAccount = { address: string; publicKey?: string };
+
+const getStashedSession = (): { accounts: StashedAccount[]; token: string } => {
+	const { accounts = [], token } = JSON.parse(localStorage.getItem(STASHED_SESSION_KEY) || '{}');
+	return { accounts, token };
+};
+
+const getPostMessagePayload = () => {
```
Jordan-Mysten, line 55: We should similarly create a standard for the iframe messaging and ensure that we have schemas + validation + version it.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	return {
+		version: PACKAGE_VERSION,
+		origin_url: window.location.href,
+		user_agent: navigator.userAgent,
+		screen_resolution: `${window.screen.width}x${window.screen.height}`,
+		language: navigator.language,
+		platform: navigator.platform,
+		timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
+		timestamp: new Date().toISOString(),
```
Jordan-Mysten, line 64: `Date.now()` is probably easier to parse

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+
+const getStashedSession = (): { accounts: StashedAccount[]; token: string } => {
+	const { accounts = [], token } = JSON.parse(localStorage.getItem(STASHED_SESSION_KEY) || '{}');
+	return { accounts, token };
+};
+
+const getPostMessagePayload = () => {
+	return {
+		version: PACKAGE_VERSION,
+		origin_url: window.location.href,
```
Jordan-Mysten, line 58: camelCase these

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	}
+
+	get version() {
+		return '1.0.0' as const;
+	}
+
+	get chains() {
+		return [SUI_MAINNET_CHAIN] as const;
```
Jordan-Mysten, line 87: You can use `SUI_CHAINS` since we support them all now

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+
+	constructor({
+		name,
+		address,
+		origin = DEFAULT_STASHED_ORIGIN,
+	}: {
+		name: string;
+		network: StashedSupportedNetwork;
+		origin?: string;
+		address?: string | null;
```
Jordan-Mysten, line 141: This can be removed since we don't support possession-of-address as an authentication model anymore

packages/zksend/src/experimental/wallet/index.ts
Jordan-Mysten: Dapps set this either automatically via dapp kit and network selection (still a little rough around the edges, idk if this actually works), or manually when calling.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	};
+
+	#signTransaction: SuiSignTransactionMethod = async ({ transaction, account, chain }) => {
+		const popup = new StashedPopup({
+			name: this.#name,
+			origin: this.#origin,
+			chain,
+		});
+
+		const tx = Transaction.from(await transaction.toJSON());
```
Jordan-Mysten, line 189: Is there a reason we're constructing a separate transaction object here instead of just using the one passed in?

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+
+		const response = await popup
+			.send({
+				type: 'sign-and-execute-transaction',
+				transaction: data,
+				address: account.address,
+				chain,
+				session: getStashedSession().token,
+			})
+			.catch((error) => {
```
Jordan-Mysten, line 232: Can remove this

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+
+// Function to clear the interval
+function clearWalletStatusInterval() {
+	if (walletStatusIntervalId !== null) {
+		clearInterval(walletStatusIntervalId);
+		walletStatusIntervalId = null;
+	}
+}
+
+const embedStashedIframe = () => {
```
Jordan-Mysten, line 371: We should try/catch this so that failure doesn't matter here

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	let addressFromRedirect: string | null = null;
+
+	try {
+		const params = new URLSearchParams(window.location.search);
+		addressFromRedirect = params.get('stashed_address') || params.get('zksend_address');
+	} catch {
+		// Ignore errors
+	}
+
+	stashedWalletInstance = new StashedWallet({
```
Jordan-Mysten, line 431: Don't make this a global

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+	}, 1000);
+
+	window.addEventListener('message', (event) => {
+		if (event.origin !== stashedWalletOrigin) return;
+		const { type, payload } = event.data;
+
+		if (type === 'WALLET_STATUS') {
+			if (!walletStatusCheckEnabled || !stashedWalletInstance) return;
+
+			stashedWalletInstance.accounts.forEach((account) => {
```
Jordan-Mysten, line 398: Instead of doing this just pass the info you need to the `embedStashedIframe` function

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,454 @@
...
+			},
+			stashedWalletOrigin,
+		);
+	}, 1000);
+
+	window.addEventListener('message', (event) => {
+		if (event.origin !== stashedWalletOrigin) return;
+		const { type, payload } = event.data;
+
+		if (type === 'WALLET_STATUS') {
```
Jordan-Mysten, line 395: We should add schemas for the types inbound.

packages/zksend/src/experimental/wallet/index.ts
hayes-mysten: What is this information actually used for? I don't recall what we wanted all this for

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: will fix -- looks like `SuiSignTransactionInput` type has `transaction` set to:

```
 transaction: {
        toJSON: () => Promise<string>;
    };
```

Should we update that?

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: @Jordan-Mysten I moved the embedStashedIframe into the class to make it easier to get current accounts without having to use a global -- does that work?

packages/zksend/src/experimental/wallet/index.ts
Jordan-Mysten: We wanted to just future proof everything so that the stashed wallet can make more informed decisions (deny old SDKs, old browsers, etc)

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+		name,
+		origin,
+	});
+
+	stashedWalletOrigin = origin || DEFAULT_STASHED_ORIGIN;
+
+	const unregister = wallets.register(stashedWalletInstance);
+
+	/* @ts-ignore */
+	if (window.stashed) {
```
Jordan-Mysten, line 457: Instead of doing this, use the wallet standard to:
- Get the list of wallets
- Subscribe to the list of wallets

If the Stashed wallet ever appears, then un-register it.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+
+const PACKAGE_VERSION = 'v1';
+
+type WalletEventsMap = {
+	[E in keyof StandardEventsListeners]: Parameters<StandardEventsListeners[E]>[0];
+};
+
+const STASHED_SESSION_KEY = 'stashed:session';
+
+let stashedWalletOrigin: string;
```
Jordan-Mysten, line 46: Don't use global mutable state

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+	#clearWalletStatusInterval = () => {
+		if (this.#walletStatusIntervalId !== null) {
+			clearInterval(this.#walletStatusIntervalId);
+			this.#walletStatusIntervalId = null;
+		}
+	};
+
+	#embedStashedIframe = () => {
+		try {
+			/* @ts-ignore */
```
Jordan-Mysten, line 183: Use expect-error instead and document why you're ignoring the error

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+
+			this.#walletStatusCheckEnabled = true;
+
+			// Wait 5 seconds before checking wallet status to avoid race condition which returns empty accounts
+			setTimeout(() => {
+				this.#walletStatusIntervalId = setInterval(() => {
+					if (!this.#walletStatusCheckEnabled || !this.#embeddedIframe) return;
+					this.#embeddedIframe.contentWindow?.postMessage(
+						{
+							type: 'WALLET_STATUS_REQUEST',
```
Jordan-Mysten, line 197: Can you just invert this directionally so that the iframe emits change events? It's fine if not that just seems a lot simpler

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+		try {
+			/* @ts-ignore */
+			this.#embeddedIframe = document.createElement('iframe');
+			this.#embeddedIframe.style.display = 'none';
+			this.#embeddedIframe.src = `${stashedWalletOrigin}/embed`;
+			document.body.appendChild(this.#embeddedIframe);
+
+			this.#walletStatusCheckEnabled = true;
+
+			// Wait 5 seconds before checking wallet status to avoid race condition which returns empty accounts
```
Jordan-Mysten, line 191: If you need to do this then something is wrong with the logic on the embed page.

packages/zksend/src/experimental/wallet/index.ts
Jordan-Mysten: Ah that's right, I forgot about that. This should be fine then

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { InferOutput } from 'valibot';
+import { array, literal, object, optional, pipe, string, url, uuid, variant } from 'valibot';
+
+export const IframeMessageWalletStatusAccount = object({
+	account: object({
+		address: string(),
```
Jordan-Mysten, line 9: We should make this support adding accounts, so this should have the public key on it

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,468 @@
...
+		this.#embeddedIframe = null;
+		this.#walletStatusCheckEnabled = false;
+		this.#walletStatusIntervalId = null;
+	}
+
+	#handleWalletStatusMessage = (event: MessageEvent) => {
+		if (event.origin !== stashedWalletOrigin || !this.#walletStatusCheckEnabled) return;
+		try {
+			const message = parse(IframeMessageWalletStatusPayload, event.data);
+			if (message.type === 'WALLET_STATUS') {
```
Jordan-Mysten, line 158: I still feel like packing accounts into the JWT + issuing new JWTs is simpler because you can basically just derive all of the accounts from the single session token instead of needing to keep track of two pieces of state.

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: moving this logic to separate PR

packages/zksend/src/experimental/wallet/channel/events.ts
claykohut-mysten: moving this logic to separate PR

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: moving this logic to separate PR

packages/zksend/src/experimental/wallet/index.ts
claykohut-mysten: moving this logic to separate PR

claykohut-mysten:
@Jordan-Mysten @hayes-mysten can y'all take another look here when possible?

packages/zksend/src/experimental/wallet/channel/index.ts
```diff
@@ -0,0 +1,181 @@
...
+			requestId: this.#id,
+			appOrigin: window.origin,
+			appName: this.#name,
+			chain: this.#chain,
+		};
+
+		const encodedRequestData = btoa(JSON.stringify(requestData));
+
+		this.#popup.location.assign(
+			`${this.#origin}/dapp/${type}${data ? `#${encodedRequestData}` : ''}`,
```
Jordan-Mysten, line 90: Do you need the type to be part of the path or can you just extract it from the data in the wallet?

.changeset/fresh-toys-mate.md
```diff
@@ -0,0 +1,5 @@
+---
```
claykohut-mysten, line 1: bad merge added a bunch of changes -- fixing

packages/zksend/src/experimental/wallet/channel/index.ts
```diff
@@ -0,0 +1,181 @@
...
+		if (typeof window === 'undefined' || !window.opener) {
+			throw new Error(
+				'StashedHost can only be used in a window opened through `window.open`. `window.opener` is not available.',
+			);
+		}
+
+		this.#request = request;
+	}
+
+	static fromUrl(url: string = window.location.href) {
```
Jordan-Mysten, line 139: We probably should make this fromPayload so that the host can just pass the big blob instead of needing to exactly match the URL. Gives us more flexibility

packages/zksend/src/experimental/wallet/channel/index.ts
claykohut-mysten: We could definitely update the wallet implementation to infer the type based on the requestData payload, which would allow us to use a generic route pattern. 

Could be beneficial to be more flexible down the line, though it's slightly more work to implement now vs. keeping the current pattern.

If you think it's worth the extra time to set up a generic route handler, I'm happy to make that change.

claykohut-mysten:
@Jordan-Mysten latest round of feedback is in

packages/zksend/src/experimental/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export * from './wallet/index.js';
```
hayes-mysten, line 4: can we update these to export the specific parts of the package that are intended to be the public API surface?

export * makes it really easy to expose all kinds of internal functionality, and very hard to change those internals without breaking changes

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export type StashedRequest = InferOutput<typeof StashedRequest>;
+
+export const StashedResponseData = variant('type', [
+	object({
+		type: literal('connect'),
+		accounts: array(
+			object({
+				address: string('`address` is required'),
+				publicKey: optional(string('`publicKey` must be a string')),
```
hayes-mysten, line 64: does this need to be optional? publicKey SHOULD be available for all account types now

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,115 @@
...
+		address: string('`address` is required'),
+		chain: string('`chain` is required'),
+		session: string('`session` is required'),
+	}),
+	object({
+		type: literal('sign-and-execute-transaction'),
+		transaction: string('`transaction` is required'),
+		address: string('`address` is required'),
+		chain: string('`chain` is required'),
+		session: string('`session` is required'),
```
hayes-mysten, line 35: should we add an optional signatures (addtionalSignatures?) for sponsored transaction flows?  I know enoki gets the user signature first and executes on the sponsor side, but I think with the upcoming balance stuff the inverse flow might become a better option for some use cases

hayes-mysten: This is probably non-blocking because it can be added later

packages/zksend/src/experimental/wallet/channel/events.ts
```diff
@@ -0,0 +1,115 @@
...
+	}),
+	object({
+		type: literal('sign-transaction'),
+		bytes: string(),
+		signature: string(),
+	}),
+	object({
+		type: literal('sign-and-execute-transaction'),
+		bytes: string(),
+		signature: string(),
```
hayes-mysten, line 77: I know the wallet standard doesn't support this today, but related to the sponsored signatures comment above, should we make this an array (with a min length 1)?   I could imagine wallet implementations potentially sponsoring some transactions, which would require multiple signatures derived in the wallet

packages/zksend/src/wallet/channel/index.ts
```diff
@@ -155,7 +155,7 @@ export class StashedHost {
 		window.opener.postMessage(
 			{
 				id: this.#request.id,
-				source: 'zksend-channel',
+				source: 'stashed-channel',
```
hayes-mysten, line 158: Do we want to keep this name here? (I don't really care, just curious if this is an intentional decision)

hayes-mysten (approved):
Had a few more comments, but overall this looks awesome! Excited for this to land

packages/zksend/src/experimental/wallet/channel/events.ts
hayes-mysten: this would apply to sign transaction as well

packages/zksend/src/wallet/channel/index.ts
hayes-mysten: resolved offline

packages/zksend/src/experimental/wallet/channel/events.ts
Jordan-Mysten: IMO we should do that when we update the wallet standard since this maps into that anyway, given we need to entirely break the ecosystem when we do that too

packages/zksend/src/wallet/channel/index.ts
Jordan-Mysten: We can update this prior to the full release

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -0,0 +1,349 @@
...
+
+	#signTransaction: SuiSignTransactionMethod = async ({ transaction, account, chain }) => {
+		const popup = new StashedPopup({
+			name: this.#name,
+			origin: this.#origin,
+			chain,
+		});
+
+		const tx = Transaction.from(await transaction.toJSON());
+		tx.setSenderIfNotSet(account.address);
```
hayes-mysten, line 167: We are sending the address anyways, and will need to reconstruct the tx on the wallet, we could probably omit this and set it there if needed

packages/zksend/src/experimental/index.ts
claykohut-mysten: done -- should note that the current zksend package does the wildcard export -- is that a concern for when this update ships?

packages/zksend/src/experimental/wallet/channel/events.ts
claykohut-mysten: It's optional because we're not yet returning it from the stashed wallet side -- we plan to do this in an update and will later make it not optional

packages/zksend/src/experimental/wallet/channel/events.ts
claykohut-mysten: 🙏

packages/zksend/src/experimental/wallet/channel/index.ts
```diff
@@ -0,0 +1,191 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import '@mysten/wallet-standard';
```
Jordan-Mysten, line 4: Unused import?

packages/zksend/src/experimental/wallet/index.ts
Jordan-Mysten: Yeah good point

Jordan-Mysten (approved):
Tiny nits but lgtm

PR #109, opened 2025-03-13
[ledger] introduce LedgerSigner and add support for clear signing

williamrobertson13:
## Description

The new major version of the Sui Ledger app has a built-in transaction parser that is used for showing clear signing displays from the Ledger app when you try to sign certain transactions for things like staking and transfer. To support this from the SDK side, we need to pass the serialized object data from the transaction to the Ledger device.

This PR introduces an experimental `LedgerSigner` class that encapsulates the underlying transport library and automatically passes through the correct clear signing data when signing transactions.

[For reference, the tests in https://github.com/obsidiansystems/ledger-app-sui/tree/develop/ragger-tests demonstrate examples of clear signing and how data should be passed through to the Ledger device]

Accompanying demo PR for the wallet: https://github.com/MystenLabs/wallet-next/pull/1338/

## Test plan
- Tested SUI transfers in Sui Wallet, still need test a USDC transfer and staking/unstaking tho


packages/ledgerjs-hw-app-sui/src/Sui.ts
```diff
@@ -87,32 +87,59 @@ export default class Sui {
...
 	 *
 	 * @param txn - The transaction; this can be any of a node Buffer, Uint8Array, or a hexadecimal string, encoding the form of the transaction appropriate for hashing and signing.
-	 * @param path - the path to use when signing the transaction.
+	 * @param path - The path to use when signing the transaction.
+	 * @param objectList - A list of object IDs used for clear signing.
 	 */
 	async signTransaction(
 		path: string,
 		txn: string | Buffer | Uint8Array,
+		objectList: Array<string | Buffer | Uint8Array> = [],
```
Jordan-Mysten, line 96: Can't you just parse the TX and extract this?

williamrobertson13: I'm asking some more questions about the whole clear signing deal in https://mysten-labs.slack.com/archives/C07UM2CD5K2/p1741882433442669?thread_ts=1741158031.084739&cid=C07UM2CD5K2.

To answer your question (if we do parse the TX on the client) - yeah we can do that similar to your transaction linter, BUT I think ideally it would be nice to not have the TypeScript SDK be a dependency on the Ledger package so we could cleanly bundle a Ledger signer into the TypeScript SDK. Eh idk this is all very up in the air with the new Ledger development kit SDKs (https://github.com/LedgerHQ/device-sdk-ts/tree/develop/packages/signer/signer-solana) so maybe it isn't helpful to over index on that atm

packages/signers/src/ledger/ledger-signer.ts
```diff
@@ -0,0 +1,137 @@
...
+	signData(): never {
+		throw new Error('Ledger Signer does not support sync signing');
+	}
+
+	/**
+	 * Signing personal messages is not supported by the Sui Ledger app.
+	 * @throws Always throws an error indicating message signing is unsupported.
+	 */
+	signPersonalMessage(): never {
+		throw new Error('Ledger Signer does not support signing personal messages');
```
Jibz-Mysten, line 96: Curious is there a reason why we dont support `signPersonalMessage`

packages/signers/src/ledger/ledger-signer.ts
```diff
@@ -0,0 +1,137 @@
...
+		return this.#publicKey;
+	}
+
+	async signTransaction(bytes: Uint8Array) {
+		const intentMessage = messageWithIntent('TransactionData', bytes);
+		const { objectIds } = getClearSigningOptions(bytes);
+		const { signature } = await this.#suiLedgerClient.signTransaction(
+			this.#derivationPath,
+			intentMessage,
+			objectIds,
```
hayes-mysten: Does the ledger actually verify this?  It seems like a little scary to tell the ledger what ids are in the tx

packages/signers/src/ledger/ledger-signer.ts
```diff
@@ -0,0 +1,137 @@
...
+			}),
+		};
+	}
+
+	/**
+	 * Generic signing is not supported by Ledger.
+	 * @throws Always throws an error indicating generic signing is unsupported.
+	 */
+	sign(): never {
+		throw new Error('Ledger Signer does not support generic signing.');
```
hayes-mysten, line 85: does the `override` keyword work here? We want to remove this from the SDK, at which point it would be great if this become invalid so we remember to remove it

packages/signers/README.md
```diff
@@ -5,18 +5,24 @@ Management Services (KMS) like AWS KMS and GCP KMS.
...
+    - [API](#api)
+      - [fromKeyId](#fromkeyid)
+        - [Parameters](#parameters)
+        - [Examples](#examples)
+  - [GCP KMS Signer](#gcp-kms-signer)
+    - [Usage](#usage-1)
+      - [fromOptions](#fromoptions)
+        - [Parameters](#parameters-1)
+        - [Examples](#examples-1)
+  - [Ledger Signer](#ledger-signer)
```
Jordan-Mysten, line 21: I'm curious what our criteria is for putting signers here vs in the main SDK, I guess the SDK is the base primitive stuff the chain knows about and this is stuff built on top of that?

packages/ledgerjs-hw-app-sui/src/Sui.ts
dfordivam: The data of objects are not part of the TX, and need be provided separately. This data is only read, and not signed.

packages/ledgerjs-hw-app-sui/src/Sui.ts
```diff
@@ -87,38 +88,75 @@ export default class Sui {
...
 	 * @param txn - The transaction; this can be any of a node Buffer, Uint8Array, or a hexadecimal string, encoding the form of the transaction appropriate for hashing and signing.
-	 * @param path - the path to use when signing the transaction.
+	 * @param path - The path to use when signing the transaction.
+	 * @param options - Additional options used for clear signing purposes.
 	 */
 	async signTransaction(
 		path: string,
 		txn: string | Buffer | Uint8Array,
+		options?: {
+			bcsObjects: (string | Buffer | Uint8Array)[];
```
Jordan-Mysten, line 98: It really sucks that this library uses Buffer. I know the underlying ledger library does it just means it's not really web compat

packages/signers/src/ledger/ledger-signer.ts
Jordan-Mysten: @hayes-mysten Wait we want to remove generic sign? I know we want to remove signData but not sign

hayes-mysten: This wasn't currently the plan.  Other keypairs only implement sign
Screenshot 2025-03-26 at 10 25 50 AM

Maybe making it protected would be the right move?

Ledger is just kinda weird

packages/signers/src/ledger/bcs.ts
```diff
@@ -0,0 +1,80 @@
...
+} from '@mysten/sui/utils';
+
+const MoveObjectType = bcs.enum('MoveObjectType', {
+	Other: bcs.StructTag,
+	GasCoin: null,
+	StakedSui: null,
+	Coin: bcs.TypeTag,
+});
+
+export const SuiMoveObject = bcs.struct('SuiMoveObject', {
```
hayes-mysten, line 20: @Jordan-Mysten we've had to define "core" bcs types in a number of SDKs.  So far we haven't needed to reuse them, but I've wondered if we should include things like this object type and the dynamic field type in the main sdk (or some other package).

(not blocking for this PR)

packages/signers/src/ledger/ledger-signer.ts
dfordivam: JFYI the support of this is in progress on the Ledger app side.

dfordivam: From the JS library side, the existing tx signing API should be used. BTW the current Sui Wallet does support personal message signing (blind signing), so I don't see any reason to limit this via this PR.

williamrobertson13: Yep thanks for the reminder, small oversight when I was pulling this together :)

packages/ledgerjs-hw-app-sui/src/Sui.ts
williamrobertson13: Yeah totally agree - it's not that hard to port this over to get rid of Buffer, but it's not really worth it atm since these `hw-app-**` libraries are going to be deprecated by Ledger soon in favor of the new development-kit SDK which doesn't rely on Buffer

packages/signers/src/ledger/ledger-signer.ts
williamrobertson13: Yeah the Ledger computes the blake2b hash of the intent message bytes on the device, so we have to break out of the default signer class behavior. Re: the above messages though, I just added `override` to all of the LedgerSigner methods so if we change the underlying interface we'll know to make changes

packages/signers/src/ledger/bcs.ts
```diff
@@ -0,0 +1,87 @@
...
+	data: bcs.enum('Data', {
+		MoveObject: bcs.struct('MoveObject', {
+			type: MoveObjectType.transform({
+				input: (objectType: string): typeof MoveObjectType.$inferType => {
```
williamrobertson13, line 27: I know this is ugly 😅

packages/signers/src/ledger/bcs.ts
```diff
@@ -0,0 +1,87 @@
...
+						structTag.module === 'staking_pool' &&
+						structTag.name === 'StakedSui'
+					) {
+						return { StakedSui: true, $kind: 'StakedSui' };
+					}
+					return {
+						Other: {
+							...structTag,
+							typeParams: structTag.typeParams.map((typeParam) => {
+								return TypeTagSerializer.parseFromStr(normalizeStructTag(typeParam));
```
williamrobertson13, line 56: @hayes-mysten this Sui type `StructTag | string` -> BCS `TagType` conversion threw me for a spin and I wasn't really sure how to generalize the logic in the BCS library. This works FWIW but not sure if you had something else in mind

williamrobertson13:
Shew okay I think I have this in decent shape - ready for another round of review 🙏🏼

packages/signers/src/ledger/bcs.ts
hayes-mysten: I think this is fine for now, what I was thinking was that the StructTag and TypeTag bcs types should support their string representations as inputs (in addition to their current format) so you can just return `Other: normalizeStructTag(structTag)` here

packages/signers/README.md
```diff
@@ -140,3 +146,55 @@ const { signature } = await signer.signPersonalMessage(messageBytes);
...
+  object containing GCP credentials and configuration.
+  - `projectId`
+    **[string](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String)**
+    The GCP project ID.
+
+##### Examples
+
+```typescript
+import Transport from '@ledgerhq/hw-transport-node-hid';
+import Sui from '@mysten/ledgerjs-hw-app-sui';
```
hayes-mysten, line 175: should we call this something else?

hayes-mysten (approved):

packages/signers/README.md
williamrobertson13: Yeah oops I copy pasted this from elsewhere - fixed!

PR #110, opened 2025-03-14
Add first draft of new experimental client with extensions

hayes-mysten:
## Description

## Test plan

---


packages/typescript/src/experimental/transports/jsonRPC.ts
```diff
@@ -0,0 +1,220 @@
...
+
+export class JSONRpcTransport implements Experimental_SuiClientTypes.TransportMethods {
+	#jsonRpcClient: SuiClient;
+
+	constructor(jsonRpcClient: SuiClient) {
+		this.#jsonRpcClient = jsonRpcClient;
+	}
+
+	async getObjects(options: Experimental_SuiClientTypes.GetObjectsOptions) {
+		const batches = batch(options.objectIds, 50);
```
Jordan-Mysten, line 26: Really glad we're shoving this down as an implementation detail.

packages/typescript/src/experimental/transports/jsonRPC.ts
```diff
@@ -0,0 +1,220 @@
...
+					results.push(parseObject(object.data!));
+				}
+			}
+		}
+
+		return {
+			objects: results,
+		};
+	}
+	async getOwnedObjects(options: Experimental_SuiClientTypes.GetOwnedObjectsOptions) {
```
Jordan-Mysten, line 51: I'm fine with this but we did talk about the "is this actually the same thing as `getObjects` with a different filter". I think the pagination needs (bounded vs unbounded list) are _maybe_ a compelling reason to keep them separated, but with things like Multiparty objects I also think these bespoke methods will get trickier (we'll either have more of them, or we'll need to consolidate later)

packages/typescript/src/experimental/transports/jsonRPC.ts
```diff
@@ -0,0 +1,220 @@
...
+			transaction: {
+				digest: await tx.getDigest(),
+				// TODO: Effects aren't returned as bcs from dryRun, once we define structured effects we can return those instead
+				effects: result.effects as never,
+				signatures: [],
+				bcs: options.transaction,
+			},
+		};
+	}
+	async getReferenceGasPrice() {
```
Jordan-Mysten, line 142: This is gonna be more interesting with congestion pricing since you'll potentially need to pass the transaction to an API to get the estimated gas price for that specific transaction instead of just the RGP. But fine for now

Jordan-Mysten: Also I imagine we'll need current epoch methods too, just assuming these are just coming later

packages/typescript/src/experimental/errors.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { ObjectResponseError } from '../client/index.js';
+
+export class SuiClientError extends Error {}
+
+export class ObjectError extends SuiClientError {
```
Jordan-Mysten, line 8: Thank you so much for making this a lot easier to debug 🙏

packages/typescript/src/experimental/client.ts
```diff
@@ -0,0 +1,94 @@
...
+
+import type {
+	ClientWithExtensions,
+	Experimental_SuiClientTypes,
+	Simplify,
+	SuiClientRegistration,
+	UnionToIntersection,
+} from './types.js';
+
+export class Experimental_SuiClient implements Experimental_SuiClientTypes.TransportMethods {
```
Jordan-Mysten, line 13: We can do this later of course but we also want this to have a `network`

Jordan-Mysten (approved):
Generally LGTM, excited for this

packages/typescript/src/experimental/transports/jsonRPC.ts
hayes-mysten: What we might want instead of this is a systemState() method or something similar that get the current epoch + gas price + some other basic info

PR #111, opened 2025-03-17
[mvr-plugin] Rewrite plugin to use the new API

manolisliolios:
## Description

This PR:
1. Rewrites the plugin to use the new MVR API endpoints
2. Enables composite type resolution, so that it can work with the "static" generation tool
3. Adds e2e testing using a pretty complex package (multi-versions, cross-chain).

## Test plan

Added new e2e tests.

regarding backwards compatibility (breaking): the GQL endpoints won't really be usable anymore. We are still in an experimental state.

### DIFF
--- a/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
+++ b/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
@@ -26,8 +26,7 @@ export type NamedPackagesPluginOptions = {
 	 * Local overrides for the resolution plugin. Pass this to pre-populate
 	 * the cache with known packages / types (especially useful for local or CI testing).
 	 *
-	 * The types cache works best with first level types, so it might be better to list
-	 * non-generic types here.
+	 * The type cache expects ONLY first-level types to ensure the cache is more composable.
 	 *
 	 * 	Expected format example:
 	 *  {
@@ -78,7 +77,7 @@ export const namedPackagesPlugin = ({
 				pageSize,
 			),
 			resolveTypes(
-				getFirstLevelNamedTypes(names.types).filter((x) => !cache.types[x]),
+				[...getFirstLevelNamedTypes(names.types)].filter((x) => !cache.types[x]),
 				url,
 				pageSize,
 			),
@@ -94,7 +93,7 @@ export const namedPackagesPlugin = ({
 		replaceNames(transactionData, {
 			packages: { ...cache.packages },
 			// we include the "composed" type cache too.
-			types: { ...cache.types, ...composedTypes },
+			types: composedTypes,
 		});
 
 		await next();
--- a/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
+++ b/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
@@ -63,7 +63,7 @@ export const namedPackagesPlugin = ({
 	pageSize = 50,
 	overrides = { packages: {}, types: {} },
 }: NamedPackagesPluginOptions) => {
-	const cache = overrides || { packages: {}, types: {} };
+	const cache = overrides;
 
 	return async (
 		transactionData: TransactionDataBuilder,
--- a/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
+++ b/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
@@ -1,15 +1,14 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { parseStructTag } from '../../utils/sui-types.js';
 import type { BuildTransactionOptions } from '../json-rpc-resolver.js';
 import type { TransactionDataBuilder } from '../TransactionData.js';
 import type { NamedPackagesPluginCache } from './utils.js';
 import {
 	batch,
-	composeCachedTypes,
-	findTransactionBlockNames,
-	getNameMappingFromResult,
+	findNamesInTransaction,
+	getFirstLevelNamedTypes,
+	populateNamedTypesFromCache,
 	replaceNames,
 } from './utils.js';
 
@@ -70,10 +69,7 @@ export const namedPackagesPlugin = ({
 		_buildOptions: BuildTransactionOptions,
 		next: () => Promise<void>,
 	) => {
-		const names = findTransactionBlockNames(transactionData);
-
-		// We try to compose all types that we might have them fully resolved.
-		const composedTypes: Record<string, string> = composeCachedTypes(names.types, cache.types);
+		const names = findNamesInTransaction(transactionData);
 
 		const [packages, types] = await Promise.all([
 			resolvePackages(
@@ -82,7 +78,7 @@ export const namedPackagesPlugin = ({
 				pageSize,
 			),
 			resolveTypes(
-				names.types.filter((x) => !cache.types[x] && !composedTypes[x]),
+				getFirstLevelNamedTypes(names.types).filter((x) => !cache.types[x]),
 				url,
 				pageSize,
 			),
@@ -92,8 +88,7 @@ export const namedPackagesPlugin = ({
 		Object.assign(cache.packages, packages);
 		Object.assign(cache.types, types);
 
-		// After the resolution, we re-compose all types that must now be fully resolved.
-		Object.assign(composedTypes, composeCachedTypes(names.types, cache.types));
+		const composedTypes = populateNamedTypesFromCache(names.types, cache.types);
 
 		// when replacing names, we also need to replace the "composed" types collected above.
 		replaceNames(transactionData, {
@@ -151,7 +146,7 @@ export const namedPackagesPlugin = ({
 
 		await Promise.all(
 			batches.map(async (batch) => {
-				const response = await fetch(`${apiUrl}/v1/type-resolution/bulk`, {
+				const response = await fetch(`${apiUrl}/v1/struct-definition/bulk`, {
 					method: 'POST',
 					headers: { 'Content-Type': 'application/json' },
 					body: JSON.stringify({
@@ -171,17 +166,8 @@ export const namedPackagesPlugin = ({
 				for (const type of Object.keys(data?.resolution)) {
 					const typeData = data.resolution[type]?.type_tag;
 					if (!typeData) continue;
-					// skip primitive types, though they shouldn't be here in the first place.
-					if (!typeData.includes('::')) continue;
 
-					// Save only "first-level" mappings.
-					const mapping = getNameMappingFromResult(
-						parseStructTag(type),
-						parseStructTag(typeData),
-						{},
-					);
-
-					Object.assign(results, mapping);
+					results[type] = typeData;
 				}
 			}),
 		);
--- a/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
+++ b/packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
@@ -1,10 +1,17 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import { parseStructTag } from '../../utils/sui-types.js';
 import type { BuildTransactionOptions } from '../json-rpc-resolver.js';
 import type { TransactionDataBuilder } from '../TransactionData.js';
 import type { NamedPackagesPluginCache } from './utils.js';
-import { batch, findTransactionBlockNames, fixComposableTypes, replaceNames } from './utils.js';
+import {
+	batch,
+	composeCachedTypes,
+	findTransactionBlockNames,
+	getNameMappingFromResult,
+	replaceNames,
+} from './utils.js';
 
 export type NamedPackagesPluginOptions = {
 	/**
@@ -19,6 +26,9 @@ export type NamedPackagesPluginOptions = {
 	/**
 	 * Local overrides for the resolution plugin. Pass this to pre-populate
 	 * the cache with known packages / types (especially useful for local or CI testing).
+	 *
+	 * The types cache works best with first level types, so it might be better to list
+	 * non-generic types here.
 	 *
 	 * 	Expected format example:
 	 *  {
@@ -53,10 +63,7 @@ export const namedPackagesPlugin = ({
 	pageSize = 50,
 	overrides = { packages: {}, types: {} },
 }: NamedPackagesPluginOptions) => {
-	const cache = {
-		packages: { ...overrides.packages },
-		types: { ...overrides.types },
-	};
+	const cache = overrides || { packages: {}, types: {} };
 
 	return async (
 		transactionData: TransactionDataBuilder,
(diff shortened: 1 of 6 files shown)


packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -68,62 +64,96 @@ export const namedPackagesPlugin = ({
...
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify({
+						names: batch,
+					}),
+				});
+
+				const data = await response.json();
+
+				if (!response.ok) throw new Error(`Failed to resolve packages: ${data?.message}`);
```
hayes-mysten, line 110: Maybe move this before parsing the body, incase error responses aren't valid json

hayes-mysten: I guess you are trying to extract the message, but I'd probably do something like:

```ts
if (!response.ok) {
  const errorBody = resonse.json().catch(() => ({ })
  throw new Error(`Failed to resolve packages: ${errorBody?.message}`);
}

const data = await response.json();
```

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -56,6 +56,33 @@ export const findTransactionBlockNames = (
...
+	types: string[],
+	typeCache: Record<string, string>,
+): Record<string, string> => {
+	const newTypes: Record<string, string> = {};
+	for (const type of types) {
+		// if we already have a resolution for this type, skip.
+		if (typeCache[type]) continue;
+		let replacement = type;
+		for (const [name, address] of Object.entries(typeCache)) {
+			if (type.includes(name)) replacement = replacement.replaceAll(name, address);
```
hayes-mysten, line 77: won't this cause issues is you have things like `@wal` and `@walrus`

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Oh yeah, fair point. I'll rewrite as above!

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -56,6 +56,33 @@ export const findTransactionBlockNames = (
...
+export const fixComposableTypes = (
+	types: string[],
+	typeCache: Record<string, string>,
+): Record<string, string> => {
+	const newTypes: Record<string, string> = {};
+	for (const type of types) {
+		// if we already have a resolution for this type, skip.
+		if (typeCache[type]) continue;
+		let replacement = type;
+		for (const [name, address] of Object.entries(typeCache)) {
```
hayes-mysten, line 76: This seems very inefficient, and will get worse as your cache grows.  I'm not sure at what point that matters, but I would be hesitant to go in this direction.

packages/typescript/src/transactions/plugins/utils.ts
manolisliolios: It shouldn't, unless I am brain-lagging. We are replacing "types", so you'd have something like `@wal::xx::XX`, not just a single name.

hayes-mysten: @wal::xx::XX and @wal::xx::XXX would still run into the issue right?

manolisliolios: agh.. yeah it would! Let me re-write to properly parse these instead.

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -21,9 +23,10 @@ export type NameResolutionRequest = {
  * Looks up all `.move` names in a transaction block.
  * Returns a list of all the names found.
  */
-export const findTransactionBlockNames = (
-	builder: TransactionDataBuilder,
-): { packages: string[]; types: string[] } => {
+export function findTransactionBlockNames(builder: TransactionDataBuilder): {
```
hayes-mysten, line 26: We've stopped using TransactionBlock as a term in the SDK, just use `Transaction`

hayes-mysten: although, in this case, I would maybe call this `findNamesInTransaction` or something, it sounds like its trying to find the names of the transaction, not names used in the transaction

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -54,32 +57,95 @@ export const findTransactionBlockNames = (
...
+	const type = isStructTag(tag) ? tag : parseStructTag(tag);
+
+	let typeTag = `${type.address}::${type.module}::${type.name}`;
+	const cacheHit = typeCache[typeTag];
+
+	if (cacheHit) {
+		let [mvrName, module, name] = cacheHit.split('::');
+		type.address = mvrName;
+		type.module = module;
+		type.name = name;
```
hayes-mysten, line 97: We generally avoid mutating data, especially when its passed in through arguments

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -54,32 +57,99 @@ export const findTransactionBlockNames = (
...
+
+	let tag = isStructTag(type) ? type : parseStructTag(type);
+
+	if (hasMvrName(tag.address)) types.add(`${tag.address}::${tag.module}::${tag.name}`);
+
+	for (const param of tag.typeParams) {
+		findMvrNames(param).forEach((name) => types.add(name));
+	}
+
+	return [...types];
```
hayes-mysten, line 93: You are converting between sets and lists a lot, it might be easier to just use sets everywhere in your utils

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,120 @@ export type NamedPackagesPluginOptions = {
...
 
-const gqlQueryKey = (idx: number) => `key_${idx}`;
+	async function resolveTypes(types: string[], apiUrl: string, pageSize: number) {
+		if (types.length === 0) return {};
+
+		const batches = batch(types, pageSize);
+		const results: Record<string, string> = {};
+
+		await Promise.all(
+			batches.map(async (batch) => {
```
hayes-mysten, line 148: Do we want to actually do this in parallel?  Generally pagination limits are there to protect servers from being overloaded.  Having 2 parallel requests is often more expensive than having one larger request.

I'm wondering if we should be doing these in series rather than parallel

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -24,6 +26,9 @@ export type NamedPackagesPluginOptions = {
 	 * Local overrides for the resolution plugin. Pass this to pre-populate
 	 * the cache with known packages / types (especially useful for local or CI testing).
 	 *
+	 * The types cache works best with first level types, so it might be better to list
```
hayes-mysten, line 29: I think providing full types is actually completely ineffective, since we will still fetch the pieces.  We should either either break down provided types on initialization, or just tell people more explicitly that they should provide only base types (possible with validation)

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,120 @@ export type NamedPackagesPluginOptions = {
...
+		Object.assign(cache.packages, packages);
+		Object.assign(cache.types, types);
+
+		const composedTypes = populateNamedTypesFromCache(names.types, cache.types);
+
+		// when replacing names, we also need to replace the "composed" types collected above.
+		replaceNames(transactionData, {
+			packages: { ...cache.packages },
+			// we include the "composed" type cache too.
+			types: { ...cache.types, ...composedTypes },
```
hayes-mysten, line 97: this doesn't do anything.  composedTypes will always contain all types that need to be replaced, and will entirely overwrite anything from cache.types

hayes-mysten (approved):
This looks pretty good to me.  Left a few comments, but overall seems pretty close to shippable!

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Agreed, I'll just error out and make it "first-level" types only (feels easier to explain).

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: It's fine. Limits protects from attacks like "Give me 1M types at once", but there's also rate limitting which would play the role of the extra guard here.

Most PTBs wouldn't have more than a single page, and even if we had 5x a single page (250 unique types/packages), we would still be fine!

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Oh, true.

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Agreed, added validation to enforce first-level only!

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,128 @@ export type NamedPackagesPluginOptions = {
...
+	async function resolveTypes(types: string[], apiUrl: string, pageSize: number) {
+		if (types.length === 0) return {};
+
+		const batches = batch(types, pageSize);
+		const results: Record<string, string> = {};
+
+		await Promise.all(
+			batches.map(async (batch) => {
+				const response = await fetch(`${apiUrl}/v1/struct-definition/bulk`, {
+					method: 'POST',
```
Jordan-Mysten, line 158: Out of scope for this PR but you really should make this GET + respond with cache-control headers (even if for like 1 minute) + put something like CloudFlare in front of the service so that you can avoid some downstream traffic.

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,128 @@ export type NamedPackagesPluginOptions = {
...
-		);
-
-		// now we need to bulk resolve all the names + types, and replace them in the transaction data.
-		(await Promise.all(batches.map((batch) => query(suiGraphQLClient, batch)))).forEach((res) => {
-			Object.assign(cache.types, res.types);
-			Object.assign(cache.packages, res.packages);
+		const names = findNamesInTransaction(transactionData);
+
+		const [packages, types] = await Promise.all([
+			resolvePackages(
```
Jordan-Mysten, line 83: Why don't we resolve these together to avoid 2 requests in most cases?

Jordan-Mysten: Oh because it's different params. Why not make the API take both, it's a bulk api after all

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,128 @@ export type NamedPackagesPluginOptions = {
...
+					method: 'POST',
+					headers: { 'Content-Type': 'application/json' },
+					body: JSON.stringify({
+						names: batch,
+					}),
+				});
+
+				if (!response.ok) {
+					const errorBody = await response.json().catch(() => ({}));
+					throw new Error(`Failed to resolve packages: ${errorBody?.message}`);
```
Jordan-Mysten, line 129: Idk if we use error causes ever but would be nice to expose the response to devs too

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -53,77 +58,128 @@ export type NamedPackagesPluginOptions = {
...
+				}
+
+				const data = await response.json();
+
+				if (!data?.resolution) return;
```
Jordan-Mysten, line 134: Isn't this bad / fatal?

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -54,32 +57,99 @@ export const findTransactionBlockNames = (
...
+	return results;
+}
+
+/**
+ * Extracts all named types from a given type.
+ */
+function findMvrNames(type: string | StructTag) {
+	const types: Set<string> = new Set();
+
+	if (typeof type === 'string' && !hasMvrName(type)) return [];
```
Jordan-Mysten, line 83: Looks like you're returning empty arrays in some cases and sets in others

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: It'll fail when trying to do replacements in the next step, which will have a much better error on which name was not resolved!

packages/typescript/src/transactions/plugins/utils.ts
```diff
@@ -54,32 +57,99 @@ export const findTransactionBlockNames = (
...
+function findAndReplaceCachedTypes(
+	tag: string | StructTag,
+	typeCache: Record<string, string>,
+): StructTag {
+	const type = isStructTag(tag) ? tag : parseStructTag(tag);
+
+	let typeTag = `${type.address}::${type.module}::${type.name}`;
+	const cacheHit = typeCache[typeTag];
+
+	if (cacheHit) {
```
Jordan-Mysten, line 128: Stylistic nit: you could also not return early and just make this a ternary in the address property of the return value

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Their logic is really different, mixes up the logic too much on the API. What's your concern on issuing two requests, given they are parallelized?

Jordan-Mysten: My main reason is just 2 > 1 and smaller number better but it's not a blocking concern at all. I might be missing it but from my perspective the functions are nearly identical in terms of how they make the request and what they do with the response?

packages/typescript/src/transactions/plugins/utils.ts
manolisliolios: whoops ty, fixed

packages/typescript/src/transactions/plugins/utils.ts
manolisliolios: True, fixed

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: Discussed offline, we'll end up trying to cache the POST requests through cloudflare, to avoid any GET request size limits being hit on our batches.

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
manolisliolios: I don't think we have any meaningful errors here worth exposing. Most bulk apis would error out only if you have syntactic (input) errors, and the cause is included in the message! 

If you still think it makes sense, how would you return this to the dev? maybe by throwing a custom error object?

PR #112, opened 2025-03-17
[mvr-static] Adds plugin to detect MVR names in project

manolisliolios:
## Description

Adds the plugin to detect all MVR names in a project, and codegen a TS file which exports the cached version.

This enables MVR usage on build-time, rather than runtime on FE apps or backends.

## Test plan

Added tests to parse subdirectories, and resolve using the new API.


packages/mvr/README.md
```diff
@@ -0,0 +1,41 @@
...
+pnpm dlx @mysten/mvr
+```
+
+Available options:
+
+- `-d <directory>`: The directory to run the command in (defaults to `.`)
+- `-f <output-file>`: The output's file name (defaults to `mvr.ts`)
+- `-d <depth>`: The depth of recursive search for MVR names (defaults to `10`)
+- `-mainnetUrl <url>`: The URL to the mainnet MVR (defaults to `https://mainnet.mvr.mystenlabs.com`)
+- `-testnetUrl <url>`: The URL to the testnet MVR (defaults to `https://testnet.mvr.mystenlabs.com`)
```
hayes-mysten, line 23: I think it might make sense to do `url-testnet` instead, then we can do something like:  `--networks mainnet,devnet --url-devnet $DEVNET_URL` in the future

packages/mvr/src/parsing.ts
```diff
@@ -0,0 +1,321 @@
...
+	const outDir = resolve(process.cwd(), args.fileName);
+
+	if (existsSync(outDir)) {
+		const shouldOverwrite = await prompt({
+			type: 'confirm',
+			name: 'overwrite',
+			message: `The file ${outDir} already exists. Do you want to overwrite it?`,
+		});
+
+		if (!shouldOverwrite) return process.exit(0);
```
hayes-mysten, line 92: exit 1?

packages/mvr/src/parsing.ts
```diff
@@ -0,0 +1,321 @@
...
+	}
+
+	console.log(`Generating ${args.fileName}...`);
+
+	const detectedNames = await findNames(args.directory, Number(args.depth));
```
hayes-mysten, line 97: `parseInt`

packages/mvr/src/parsing.ts
```diff
@@ -0,0 +1,321 @@
...
+	testnet: NamedPackagesPluginCache,
+) {
+	const outputFile = `${WARNING_MESSAGE}
+const testnetResolution = {
+	packages: ${JSON.stringify(testnet.packages).replaceAll('"', "'")},
+	types: ${JSON.stringify(testnet.types).replaceAll('"', "'")},
+};
+
+const mainnetResolution = {
+	packages: ${JSON.stringify(mainnet.packages).replaceAll('"', "'")},
```
hayes-mysten, line 116: `{JSON.stringify(mainnet.packages, null, 2)`

packages/mvr/README.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+Available options:
+
+- `--directory <directory>`: The directory to run the command in (defaults to `.`)
+- `--output <file-name>`: The output's file name (defaults to `mvr.ts`)
+- `--depth <depth>`: The depth of recursive search for MVR names (defaults to `10`)
+- `--url-mainnet <url>`: The URL to the mainnet MVR (defaults to `https://mainnet.mvr.mystenlabs.com`)
+- `--url-testnet <url>`: The URL to the testnet MVR (defaults to `https://testnet.mvr.mystenlabs.com`)
+- `--include <dir_patterns>`: The directory patterns to include in the search (defaults to `**/*.{js,ts,jsx,tsx,mjs,cjs}`)
```
hayes-mysten, line 24: One thing to node about these patterns is that these patterns are mostly all valid bash expansions as well, so if users don't quote them properly they will be expanded by bash instead of glob.

This SHOULD be fine, but the effect is that you will get a large list of files as inputs rather than the single glob pattern.  I'd just double check that this works as expected

packages/mvr/src/parsing.ts
```diff
@@ -0,0 +1,307 @@
...
+		force: {
+			type: 'boolean',
+			default: false,
+			short: 'f',
+		},
+	},
+});
+
+export async function parser() {
+	const outDir = resolve(process.cwd(), args.output);
```
hayes-mysten, line 82: this is a file not a directory right?

hayes-mysten: You probably do want to use the node equivalent of `mkdir -p` to ensure the directory exists, otherwise you might error when writing the file

hayes-mysten (commented):
This looks pretty good to me!

hayes-mysten (approved):

PR #94, opened 2025-03-11
[enoki] Add wallet-standard implementation for enoki

hayes-mysten:
## Description

Original PR here: https://github.com/MystenLabs/sui/pull/18229

## Test plan

How did you test the new or updated feature?

---

PR #113, opened 2025-03-17
[enoki] deprecate EnokiFlow in favor of a wallet-standard interface

williamrobertson13:
## Description

This is a fork of https://github.com/MystenLabs/ts-sdks/pull/94 with some finishing touches and tweaks (namely not providing React bindings). The corresponding doc changes are in https://github.com/MystenLabs/enoki/pull/408!

https://enoki-docs-git-wrobertson-enokidocsfollowup-mysten-labs.vercel.app/ts-sdk for the docs preview

## Test plan
- Local testing with the demo app: auto-connect, switching networks, connecting, and signing

With wallets included in the connect modal:
image

Signed transaction with Google account:
image


packages/enoki/src/wallet/index.ts
```diff
@@ -0,0 +1,335 @@
...
+import { ENOKI_PROVIDER_WALLETS_INFO } from './providers.js';
+import type { RegisterEnokiWalletsOptions, WalletEventsMap } from './types.js';
+
+export class EnokiWallet implements Wallet {
+	#events: Emitter<WalletEventsMap>;
+	#accounts: ReadonlyWalletAccount[];
+	#name: string;
+	#id: string;
+	#icon: Wallet['icon'];
+	#flow: EnokiFlow;
```
williamrobertson13, line 39: I know we want to have this eventually not depend on the deprecated EnokiFlow. I could be super lazy and just copy over a not-deprecated `EnokiFlow` class that isn't exposed externally 😛... I guess I was just going to leave this as-is and take care of this after the deprecation notice bakes for a while

packages/enoki/src/EnokiFlow.ts
```diff
@@ -311,73 +313,4 @@ export class EnokiFlow {
 			ephemeralKeypair: Ed25519Keypair.fromSecretKey(fromBase64(zkp.ephemeralKeyPair)),
 		});
 	}
-
-	async sponsorTransaction({
```
hayes-mysten, line 315: we should mention this in a changeset

packages/enoki/src/wallet/index.ts
```diff
@@ -0,0 +1,335 @@
...
+	#provider: AuthProvider;
+	#clientId: string;
+	#redirectUrl: string | undefined;
+	#extraParams: Record<string, string> | undefined;
+	#network: EnokiNetwork;
```
hayes-mysten, line 44: I know this is how it worked in the version of the PR I created, and how EnokiFlow currently works, but it would be so much nicer if we didn't have to re-register wallets when the network changed, and could allow the wallet to work across multiple networks.

Skimming over the EnokiFlow class, I think it kinda has broken handling for networks (or more accurately assumes the network will always be the same, even though methods accept network as an argument). 

The hard part about this (today) is that we would need to provide a client for each network when registering wallets.  Internally we could either make EnokiFlow a little smarter about multiple networks, or just have seperate EnokiFlow instances for each network on a wallet. 

This isn't a blocker, but I suspect for non-react users who aren't just copying the example code this might be a footgun

packages/enoki/src/wallet/index.ts
```diff
@@ -0,0 +1,335 @@
...
+		this.#redirectUrl = redirectUrl;
+		this.#extraParams = extraParams;
+		this.#network = network;
+		this.#windowFeatures = windowFeatures;
+
+		this.#setAccount();
+	}
+
+	#signTransaction: SuiSignTransactionMethod = async ({ transaction }) => {
+		const parsedTransaction = Transaction.from(await transaction.toJSON());
```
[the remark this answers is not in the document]
hayes-mysten: I don't think dappkit or the wallet standard enforces that the provided chain matches a supported chain, so we might actually want to validate this here

packages/enoki/src/wallet/index.ts
```diff
@@ -0,0 +1,335 @@
...
+	#setAccount() {
+		const state = this.#flow.$zkLoginState.get();
+		if (state.address) {
+			this.#accounts = [
+				new ReadonlyWalletAccount({
+					address: state.address,
+					chains: this.chains,
+					features: Object.keys(this.features) as IdentifierArray,
+					// NOTE: zkLogin accounts don't have meaningful public keys, so we will leave this empty.
+					publicKey: new Uint8Array(),
```
hayes-mysten, line 203: this isn't actually true.  We should definitely be providing a publicKey here now

packages/enoki/src/wallet/index.ts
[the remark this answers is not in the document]
williamrobertson13: yeah good call, I was messing around with this a little yesterda

hayes-mysten (approved):

packages/enoki/src/wallet/state.ts
```diff
@@ -0,0 +1,306 @@
...
+	jwt?: string;
+	proof?: ZkLoginSignatureInputs;
+}
+
+const createStorageKeys = (apiKey: string, network: EnokiNetwork) => ({
+	STATE: `@enoki/flow/state/${apiKey}/${network}`,
+	SESSION: `@enoki/flow/session/${apiKey}/${network}`,
+});
+
+export class INTERNAL_EnokiFlow {
```
williamrobertson13, line 51: this is all copy pasta'd - the only change is tacking on `network` to the storage keys

packages/enoki/src/wallet/index.ts
hayes-mysten: sounds good to me

PR #125, opened 2025-03-20
wip fix CI

williamrobertson13:
## Description

## Test plan

---

### DIFF
--- a/packages/signers/package.json
+++ b/packages/signers/package.json
@@ -27,7 +27,8 @@
 		"aws",
 		"dist",
 		"gcp",
-		"src"
+		"src",
+		"webcrypto"
 	],
 	"scripts": {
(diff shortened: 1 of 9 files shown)


packages/signers/package.json
```diff
@@ -27,7 +27,8 @@
 		"aws",
 		"dist",
 		"gcp",
-		"src"
+		"src",
```
williamrobertson13, line 30: this is all auto-generated from running `pnpm run build`

hayes-mysten (approved):

packages/signers/package.json
williamrobertson13: eh I'll just rm it

hayes-mysten: I think its already release, and this should probably have been merged as part of the original PR

PR #134, opened 2025-03-26
[Seal] SDK v0.3

benr-ml:
## Description

## Test plan

---


packages/seal/CHANGELOG.md
```diff
@@ -0,0 +1,23 @@
+# @mysten/seal
+
+## 0.3.0
```
hayes-mysten, line 3: We should not be created/releasing changesets outside the changesets workflow.  It doesn't look like 0.3 was actually published, so this probably should not exist yet

packages/seal/README.md
```diff
@@ -0,0 +1,6 @@
+# `@mysten/seal`
+
+CONFIDENTIAL AND PROPRIETARY
```
hayes-mysten, line 3: Is this still confidential?

packages/seal/examples/encrypt.ts
```diff
@@ -0,0 +1,2 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
```
hayes-mysten, line 2: ?

packages/seal/package.json
```diff
@@ -0,0 +1,60 @@
+{
+	"name": "@mysten/seal",
+	"version": "0.3.0",
```
hayes-mysten, line 3: versions should be bumped only through changesets.  it looks like 0.2 was the last published version.  This should stay at 0.2 and have a changeset with a `minor`  change that will trigger the 0.3 release (if thats what we want)

packages/seal/package.json
```diff
@@ -0,0 +1,60 @@
...
+	},
+	"sideEffects": false,
+	"files": [
+		"CHANGELOG.md",
+		"dist"
+	],
+	"scripts": {
+		"clean": "rm -rf tsconfig.tsbuildinfo ./dist",
+		"build": "node genversion.mjs && build-package",
+		"codegen:version": "node genversion.mjs",
```
hayes-mysten, line 25: Is this used or should it be removed?

packages/seal/src/bcs.ts
```diff
@@ -0,0 +1,55 @@
...
+		aad: bcs.option(bcs.vector(bcs.U8)),
+	}),
+	Hmac256Ctr: bcs.struct('Hmac256Ctr', {
+		blob: bcs.vector(bcs.U8),
+		aad: bcs.option(bcs.vector(bcs.U8)),
+		mac: bcs.bytes(32),
+	}),
+	Plain: bcs.struct('Plain', {}),
+});
+export type CiphertextType = typeof Ciphertext.$inferInput;
```
hayes-mysten, line 28: We generally don't export these and use `typeof Ciphertext.$inferInput/.$inferType` as needed.  Even in this file you are inconsistent about if XType refers the the input or output type which is confusing

packages/seal/src/elgamal.ts
```diff
@@ -0,0 +1,54 @@
...
+ * The ciphertext is a pair of G1Elements (48 bytes).
+ */
+function decrypt(sk: Scalar, encryption: [G1Element, G1Element]): G1Element {
+	return encryption[1].subtract(encryption[0].multiply(sk));
+}
+
+// /**
+//  * Encrypt a message with a given public key. Both the public key and the message must a compressed G1Element (48 bytes).
+//  */
+// function elgamal_encrypt(pk: Uint8Array, message: Uint8Array): [Uint8Array, Uint8Array] {
```
hayes-mysten, line 28: whats the plan with these commented out methods? 

If we want to leave them, can we add a comment about why they are currently commented out?

packages/seal/src/session-key.ts
```diff
@@ -0,0 +1,137 @@
...
+		const message = `Accessing keys of package ${this.#packageId} for ${this.#ttlMin} mins from ${creationTimeUtc}, session key ${toBase64(this.#sessionKey.getPublicKey().toRawBytes())}`;
+		return new TextEncoder().encode(message);
+	}
+
+	async setPersonalMessageSignature(personalMessageSignature: string) {
+		try {
+			await verifyPersonalMessageSignature(this.getPersonalMessage(), personalMessageSignature, {
+				address: this.#address,
+				client: new SuiGraphQLClient({
+					url: 'https://sui-testnet.mystenlabs.com/graphql',
```
hayes-mysten, line 93: This will cause problems when we support mainnet

packages/seal/typedoc.json
```diff
@@ -0,0 +1,6 @@
+{
+	"entryPoints": ["src"],
```
hayes-mysten, line 2: I think you will also need to add the seal sdk as a dep for the docs package, and add seal to the typedoc config to get this to show up properly

packages/seal/vitest.config.mts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+	test: {
+		environment: 'happy-dom',
```
hayes-mysten, line 8: do we need this?

packages/seal/package.json
benr-ml: used

packages/seal/src/session-key.ts
benr-ml: Added a TODO to fix it (we have time before mainnet)

packages/seal/package.json
hayes-mysten: I couldn't find any reference to `codegen:version` in the PR, not a big deal either way

hayes-mysten (approved):

PR #147, opened 2025-03-28
[ci] Creating a dispatch workflow for seal ci

quepcompute:
## Description

see title

## Test plan

execute workflow from branch

---

### DIFF
--- a/.github/workflows/seal-ci.yml
+++ b/.github/workflows/seal-ci.yml
@@ -1,0 +1,28 @@
+---
+name: Seal Integration tests
+on:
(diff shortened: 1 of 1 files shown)


benr-ml (approved):

joyqvq:
do we need this ci? i think https://github.com/MystenLabs/ts-sdks/blob/8ca8df3fecc9d50c7cfc46c7c367e77e45c1ba2e/.github/workflows/turborepo.yml this already runs for all packages including /seal @quepcompute

PR #154, opened 2025-03-31
export EncryptedObject

joyqvq:
## Description

parsing for the EncrytedObject.id is useful in client code to construct txBytes for ids with nonce
## Test plan

---


.changeset/metal-dodos-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mysten/seal': minor
```
benr-ml, line 2: let's use a patch, not minor

benr-ml (approved):

PR #157, opened 2025-03-31
Parallel Executor: add additional signatures

vmc-crossmint:
## Description
The parallel and serial executors were using only the sponsor signatures to send transactions. But were not taking any param to pass additional signatures. This pr should address that.

## Test plan

---


hayes-mysten (approved):
Thanks for the PR!

I think we just need a changeset to get this released (which you can add either through the PR comment, or by running `pnpm changeset` at the root of the repo

PR #159, opened 2025-03-31
prototype client cache

hayes-mysten:
## Description

## Test plan

---


packages/typescript/src/experimental/cache.ts
```diff
@@ -0,0 +1,64 @@
...
+
+		if (this.#cache.has(cacheKey)) {
+			return this.#cache.get(cacheKey) as T;
+		}
+
+		const result = load();
+
+		this.#cache.set(cacheKey, result);
+
+		if (typeof result === 'object' && result !== null && 'then' in result) {
```
Jordan-Mysten, line 29: I assume this is just to avoid the additional microtask by caching the promise itself?

packages/typescript/src/experimental/cache.ts
```diff
@@ -0,0 +1,64 @@
...
+		}
+
+		for (const key of this.#cache.keys()) {
+			if (key.startsWith(prefixKey)) {
+				this.#cache.delete(key);
+			}
+		}
+	}
+
+	scope(prefix: string | string[]) {
```
Jordan-Mysten, line 58: Should we force folks to use this via scope? Or are there good reasons for global caches?

Jordan-Mysten (approved):
Super cool!

packages/typescript/src/experimental/cache.ts
hayes-mysten: yup

packages/typescript/src/experimental/cache.ts
hayes-mysten: I don't really know.  I don't think extensions will share caches, but I haven't thought through it

PR #160, opened 2025-03-31
walrus client wasm

hayes-mysten:
## Description

## Test plan

---


packages/walrus/src/storage-node/client.ts
```diff
@@ -232,7 +232,7 @@ export class StorageNodeClient {
 		let response: Response | undefined;
 
 		try {
-			response = await this.#fetch(`${nodeUrl}${path}`, {
+			response = await (0, this.#fetch)(`${nodeUrl}${path}`, {
```
Jordan-Mysten, line 235: Curious why you went this route

Jordan-Mysten (approved):

packages/walrus/src/storage-node/client.ts
hayes-mysten: fetch throws an error if its bound to not window

PR #167, opened 2025-04-01
Experimental package: metadata api

claykohut-mysten:
## Description

- Pull stashed wallet metadata from API
- Fallback to defaults
- Need to update url from localhost to actual url when set

Screenshot 2025-04-01 at 10 29 50 AM

## Test plan

---


packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -114,18 +118,32 @@ export class StashedWallet implements Wallet {
...
-		this.#origin = origin;
+		this.#origin = origin || DEFAULT_STASHED_ORIGIN;
 		this.#name = name;
+
+		this.#fetchWalletMetadata();
+	}
+
+	async #fetchWalletMetadata() {
+		try {
+			const response = await fetch(METADATA_API_URL);
```
hayes-mysten, line 132: This should probably be added as an option in the wallet constructor so that it can be overridden for local testing

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -116,16 +121,41 @@ export class StashedWallet implements Wallet {
...
+		useDynamicMetadata?: boolean;
 		chain?: SuiChain;
 	}) {
 		this.#accounts = [];
 		this.#events = mitt();
-		this.#origin = origin;
+		this.#origin = origin || DEFAULT_STASHED_ORIGIN;
 		this.#name = name;
+		this.#useDynamicMetadata = useDynamicMetadata;
+		if (this.#useDynamicMetadata) {
```
Jordan-Mysten, line 137: IMO these should be fetched as part of the `register` call so that we can disable the actual registration itself. We probably want to disable this by default, and only enable it when the HTTP call comes back (for the initial release of the wallet at least)

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -337,9 +347,21 @@ export function registerStashedWallet(
...
+			metadata = await response.json();
+		} catch (error) {
+			console.error('Error fetching metadata', error);
+		}
+	}
+
 	const stashedWalletInstance = new StashedWallet({
 		name,
 		origin,
+		...metadata,
```
claykohut-mysten, line 364: should I add a schema for this?

Jordan-Mysten: Probably, we should also probably chuck this into a property namespace and not spread it so that you only need to schematize the variables parts of the constructor (I like `metadata` FWIW)

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -337,9 +347,21 @@ export function registerStashedWallet(
 		// Ignore errors
 	}
 
+	let metadata = {};
+	if (useDynamicMetadata) {
+		try {
+			const response = await fetch(METADATA_API_URL);
```
Jordan-Mysten, line 353: Accept a URL override in the `registerStashedWallet` call.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -313,12 +321,14 @@ export class StashedWallet implements Wallet {
 	};
 }
 
-export function registerStashedWallet(
+export async function registerStashedWallet(
 	name: string,
 	{
 		origin,
+		useDynamicMetadata = true,
```
Jordan-Mysten, line 328: IMO for the initial release we should not let folks opt-out of this, since we know we'll be doing a day one update.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -337,9 +347,21 @@ export function registerStashedWallet(
 		// Ignore errors
 	}
 
+	let metadata = {};
+	if (useDynamicMetadata) {
+		try {
+			const response = await fetch(METADATA_API_URL);
+			if (!response.ok) throw new Error('Failed to fetch wallet metadata');
+			metadata = await response.json();
```
Jordan-Mysten, line 355: To control the visibility of this, I think one of the pieces of metadata should be `enabled: boolean` so that the response can dictate if the wallet registers or not.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -48,19 +48,22 @@ const getStashedSession = (): { accounts: StashedAccount[]; token: string } => {
 };
 
 const SUI_WALLET_EXTENSION_ID = 'com.sui.extension' as const;
-
+const METADATA_API_URL = ' http://localhost:3001/api/wallet/metadata';
 export class StashedWallet implements Wallet {
 	#events: Emitter<WalletEventsMap>;
 	#accounts: ReadonlyWalletAccount[];
 	#origin: string;
+	#walletName: string = STASHED_WALLET_NAME; // Default name
```
Jordan-Mysten, line 56: I _think_ for the initial release we can make dynamic config a requirement and follow-up with opt-outs + static fallbacks

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -48,19 +50,32 @@ const getStashedSession = (): { accounts: StashedAccount[]; token: string } => {
 };
 
 const SUI_WALLET_EXTENSION_ID = 'com.sui.extension' as const;
-
+const METADATA_API_URL = 'http://localhost:3001/api/wallet/metadata';
+
+const WalletMetadataSchema = object({
+	walletName: string('Wallet name is required'),
+	icon: custom<WalletIcon>((value) => {
```
claykohut-mysten, line 57: Is this check too restrictive? We could also just use a default icon if invalid url.

packages/zksend/src/experimental/wallet/index.ts
```diff
@@ -337,9 +365,17 @@ export function registerStashedWallet(
 		// Ignore errors
 	}
 
+	const metadata = await fetchMetadata(metadataApiUrl);
+
+	if (!metadata?.enabled) {
+		console.error('Stashed wallet is not currently enabled.');
```
Jordan-Mysten, line 371: Maybe just do `console.log` since this really isn't "scary"

Jordan-Mysten (approved):

Issue #181, opened 2025-04-06
Incorrect usage of systemObjectId as package in delete_blob Move call


daoauth:
https://github.com/MystenLabs/ts-sdks/blob/d54486981274ae3af89920fcc08a32d3763c053c/packages/walrus/src/client.ts#L990-L998

this.#packageConfig.systemObjectId is being passed to the package field, but this value is not a package ID. It leads to the following error when executing the transaction:

```
Object is not a package with ID 0x6c2547cbbc38025cf3adac45f63cb0a8d12ecf777cdc75a4971612bf97fdf6af
```

I believe this should instead refer to the actual package address that contains the system module.

And, It would also be helpful if the SDK exported a **MAINNET_WALRUS_PACKAGE_CONFIG** alongside **TESTNET_WALRUS_PACKAGE_CONFIG** for easier configuration.

hayes-mysten:
Fix has been merged, the release should be out in a few minutes!

PR #182, opened 2025-04-06
fix: delete blob tx with wrong pkg address

Jarekkkkk:
[related_issue](https://github.com/MystenLabs/ts-sdks/issues/181)

### DIFF
--- a/.changeset/late-planes-move.md
+++ b/.changeset/late-planes-move.md
@@ -1,0 +1,5 @@
+---
+'@mysten/walrus': patch
+---
+
+fix packageId in deleteBlob move call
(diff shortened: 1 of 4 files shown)


daoauth:
Using a **systemContract.delete_blob** instead of making direct Move calls seems more consistent.

https://github.com/MystenLabs/ts-sdks/blob/main/packages/walrus/src/contracts/system.ts

Jarekkkkk:
> Using a **systemContract.delete_blob** instead of making direct Move calls seems more consistent.
> 
> https://github.com/MystenLabs/ts-sdks/blob/main/packages/walrus/src/contracts/system.ts

just updated, plz check sir

hayes-mysten (approved):
Thanks for the fix!

PR #183, opened 2025-04-06
fix (docs: walrus): import package & writeBlob.option?

nidz-the-fact:
## Description

Edit From should be **@mysten/walrus** and **writeBlob({ deletable: `false` })**


hayes-mysten (approved):
thanks

PR #189, opened 2025-04-08
feat(docs): walrus fix install package

nidz-the-fact:
## Description

Should `--save` be removed?

### DIFF
--- a/packages/docs/content/walrus/index.mdx
+++ b/packages/docs/content/walrus/index.mdx
@@ -5,7 +5,7 @@ title: Walrus SDK
 ## Installation
 
 ```bash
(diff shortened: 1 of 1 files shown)


hayes-mysten:
`--save` will add it to you package.json, what is the issue you are running into?

hayes-mysten:
I don't really understand what you mean, I think for most users, installing in dependencies, not devDependencies is the right location, and If users are adding this for some sort of dev script, they should know to use --save-dev if needed.

Almost all documentation for packages uses `--save` (or `add` for yarn/pnpm) as the default recommendation in their docs

PR #192, opened 2025-04-09
init window-wallet-core package

pchrysochoidis:
## Description

* init window-wallet-code package thatwill be the place for the common functionality between window wallets
* adds `web-wallet-channel` that allows connecting dapps with wallets running on web
* `jwt-session` - wallet-connection/jwt utility functions for creating/decoding/verifying session jwts

example usage -> https://github.com/MystenLabs/ts-sdks/pull/83 and https://github.com/MystenLabs/ts-sdks/pull/225

## Test plan

---


packages/window-wallet-channel/src/requests.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as v from 'valibot';
+
+export type JsonData = string | number | boolean | null | { [key: string]: JsonData } | JsonData[];
+
+const JsonSchema: v.GenericSchema<JsonData> = v.lazy(() =>
```
hayes-mysten, line 8: Are these all exactly the same as the ones from the stashed package?  Is the plan to have separate copies, but keep them in sync until we have good reason for theme to diverge?

packages/window-wallet-channel/src/utils.ts
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export interface Resolvers<T = any> {
+	promise: Promise<T>;
+	reject: (error: Error) => void;
+	resolve: (value: T) => void;
+}
+
+export function withResolvers<T = any>(): Resolvers<T> {
```
hayes-mysten, line 10: We have this in several packages now.  At some point soon I want to have some shared utils, but I'm not sure where they should live yet.

hayes-mysten: I think maybe `@mysten/sdk-utils` might be good? I want to have something we can use for things like bcs that aren't directly tied to the `@mysten/sui` package (not a blocker for this PR)

packages/window-wallet-channel/src/index.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export { HostResponse } from './host-response.js';
+export { WindowRequest as PopupRequest } from './window-request.js';
```
hayes-mysten, line 5: Why are these called 2 different things?

hayes-mysten: In general, I don't really like any of the names of these exports. They aren't really requests/responses themselves, the are the 2 ends of a message channel, and aren't inherently limited to one message

hayes-mysten (commented):
Overall this looks reasonable, and very similar to the other package.  Is the plan to use this as shared logic for multiple implementations?

If so, are do we want to use this in the other package right away?

packages/window-wallet-channel/src/requests.ts
pchrysochoidis: Yes it's more or less the same with some extra fields like `extraRequestOptions`. I think we should use this in both stashed and enoki that's was my intension anyway to have only one implementation of this and reuse it

packages/window-wallet-channel/src/index.ts
pchrysochoidis: Yeah I don't like them either, let me know if you have any suggestions.

pchrysochoidis:
> Overall this looks reasonable, and very similar to the other package. Is the plan to use this as shared logic for multiple implementations?
> 

Yes that's the plan - still need to work on naming and the jwt session

> If so, are do we want to use this in the other package right away?

Probably yes, it's not a big change anyway

packages/wallet-core/package.json
```diff
@@ -0,0 +1,50 @@
+{
+	"name": "@mysten/wallet-core",
+	"version": "0.0.1-experimental",
```
pchrysochoidis, line 3: I made it public and added `experimental` to the version because `dapp-kit` changeset  snapshot was failing

packages/wallet-core/src/jwt-session/index.ts
```diff
@@ -0,0 +1,59 @@
...
+import type { JWTPayload } from 'jose';
+import { SignJWT, decodeJwt, jwtVerify } from 'jose';
+
+const AccountSchema = v.object({
+	address: v.string(),
+	publicKey: v.string(),
+	nickname: v.optional(v.string()),
+	icon: v.optional(v.string()),
+	chains: v.optional(v.array(v.pipe(v.string(), v.regex<`${string}:${string}`>(/^.+:.+$/)))),
+	features: v.optional(v.array(v.pipe(v.string(), v.regex<`${string}:${string}`>(/^.+:.+$/)))),
```
pchrysochoidis, line 14: Do we expect the account to have `features` or it should be something that wallet-standard should only handle?

packages/wallet-core/src/web-wallet-channel/requests.ts
```diff
@@ -0,0 +1,57 @@
...
+		v.null(),
+		v.record(v.string(), JsonSchema),
+		v.array(JsonSchema),
+	]),
+);
+
+export const RequestData = v.variant('type', [
+	v.object({
+		type: v.literal('connect'),
+		chain: v.string('`chain` is required'),
```
pchrysochoidis, line 22: added chain to connect as well and until we have it in the connect request we can default it with the network of the wallet-standard ie [here](https://github.com/MystenLabs/ts-sdks/pull/83/files#diff-c9f674dd4746d5e05a9340d7ac3a2f50287e2734ed0672b086ea19562b4b88ffR225)

Jordan-Mysten: Connect should be chain-agnostic?

pchrysochoidis: I added this mainly for zklogin that we usually create a jwt with ephemeral private key based on the selected network. But I think you are right chain is not required for connect. I will remove it

packages/wallet-core/src/jwt-session/index.ts
hayes-mysten: I think its okay to leave this off for now.  @Jordan-Mysten do we plan to expose read-only accounts through the wallet-standard?

I think that's the main use case here, if you want to be able to log in/connect without the ability to sign transactions

hayes-mysten: I *think* you can add read only accounts without owning the account, so it would probably be weird to allow connecting with that kind of account

hayes-mysten: This is probably fine as is.  I'm not sure how useful it is, but having it optional in the schema doesn't hurt

packages/wallet-core/src/web-wallet-channel/responses.ts
```diff
@@ -0,0 +1,53 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as v from 'valibot';
+
+export const ResponseData = v.variant('type', [
+	v.object({
+		type: v.literal('connect'),
+		session: v.string('`session` is required'),
```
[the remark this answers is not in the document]
hayes-mysten: Is that really what we want?

Session is going to be sent back and forth between the app and wallet on every message.  Do we want to embed all the information for all the accounts in every request?

packages/wallet-core/src/web-wallet-channel/wallet-post-message-channel.ts
```diff
@@ -0,0 +1,79 @@
...
+
+	async verifyJwtSession(secretKey: Uint8Array) {
+		if (!('session' in this.#request.payload)) {
+			return null;
+		}
+
+		const { session } = await verifyJwtSession(this.#request.payload.session, secretKey);
+
+		if (session.appOrigin !== new URL(this.#request.appUrl).origin) {
+			throw new Error('App and session origin mismatch');
```
hayes-mysten, line 48: should this be verified agains the opener?

packages/wallet-core/src/web-wallet-channel/wallet-post-message-channel.ts
```diff
@@ -0,0 +1,79 @@
...
+		}
+
+		return session;
+	}
+
+	sendMessage(payload: ResponsePayloadType) {
+		window.opener.postMessage(
+			{
+				id: this.#request.requestId,
+				source: 'web-wallet-channel',
```
hayes-mysten, line 65: should this be customizable (cc @claykohut-mysten). Not sure if we want to use the same name here for enoki connect vs stashed web

packages/wallet-core/src/web-wallet-channel/dapp-post-message-channel.ts
```diff
@@ -0,0 +1,126 @@
...
+		}, 1000);
+	}
+
+	send<T extends RequestDataType['type']>({
+		type,
+		...data
+	}: {
+		type: T;
+	} & Extract<RequestDataType, { type: T }>): Promise<ResponseTypes[T]> {
+		window.addEventListener('message', this.#listener);
```
hayes-mysten, line 73: We should probably prevent calling send multiple times, otherwise this could end up in a bad state

hayes-mysten: we should also make this we haven't closed when sending

packages/wallet-core/src/web-wallet-channel/utils.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export function getClientMetadata() {
+	return {
+		version: 'v1',
```
hayes-mysten, line 6: why not just `version: 1`

hayes-mysten: if we want something like semver, we should probably do that from the start

packages/wallet-core/src/web-wallet-channel/wallet-post-message-channel.ts
```diff
@@ -0,0 +1,79 @@
...
+
+		if (!addressInSession) {
+			throw new Error('Requested account not found in session');
+		}
+
+		return session;
+	}
+
+	sendMessage(payload: ResponsePayloadType) {
+		window.opener.postMessage(
```
hayes-mysten, line 62: We should probably enforce that only 1 message can be sent

packages/wallet-core/package.json
hayes-mysten: does this work with changesets? I thought you had to manage these kinds of tags through some extra state that changesets tracks

packages/wallet-core/src/jwt-session/index.ts
```diff
@@ -0,0 +1,59 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import * as v from 'valibot';
+import type { JWTPayload } from 'jose';
+import { SignJWT, decodeJwt, jwtVerify } from 'jose';
+
+const AccountSchema = v.object({
+	address: v.string(),
+	publicKey: v.string(),
+	nickname: v.optional(v.string()),
```
hayes, line 11: nickname I think should not be shared with the dapp

packages/wallet-core/src/jwt-session/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	icon: v.optional(v.string()),
```
hayes, line 12: what is this? I've never seen accounts with icons

packages/wallet-core/src/jwt-session/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	chains: v.optional(v.array(v.pipe(v.string(), v.regex<`${string}:${string}`>(/^.+:.+$/)))),
+	features: v.optional(v.array(v.pipe(v.string(), v.regex<`${string}:${string}`>(/^.+:.+$/)))),
+});
+
+const JwtSessionSchema = v.looseObject({
```
Jordan-Mysten, line 17: This is way too much info, let's trim down similar to what we did here: https://github.com/MystenLabs/wallet-next/pull/1210/files#diff-bc3253061283423c721c833f262624f2ba8b6073381c5122cc6671b500cfc743

packages/wallet-core/src/web-wallet-channel/responses.ts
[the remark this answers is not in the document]
hayes-mysten: I think if we trim it down a bit, this is probably fine

packages/wallet-core/src/jwt-session/index.ts
pchrysochoidis: Hmm yeah I had similar thoughts and in the end I added it. Worst case I think we just wont use it

packages/wallet-core/src/web-wallet-channel/responses.ts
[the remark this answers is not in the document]
pchrysochoidis: It might be a bit of an overhead but felt easier to have everything in jwt, to have a single source of truth mainly.

I assumed that data wont be that big and probably it will be fine/reasonably fast to store and serialize/deserialise.

Do you think it's better to have only account addresses in jwt and another field for all account info? I can do that

packages/wallet-core/src/web-wallet-channel/utils.ts
pchrysochoidis: It's the metadata version. I will change it to 1

pchrysochoidis:
I am updating this to the latest changes from [here](https://github.com/MystenLabs/wallet-next/pull/1210/files#)

packages/wallet-core/src/jwt-session/index.ts
pchrysochoidis: Why not? Since it's optional should I keep it or remove it?

packages/wallet-core/src/web-wallet-channel/wallet-post-message-channel.ts
pchrysochoidis: Now that I am thinking about it, I don't think it will make things more secure.
But I added it to test it.

packages/wallet-core/src/jwt-session/index.ts
pchrysochoidis: Remove this

packages/wallet-core/src/jwt-session/index.ts
pchrysochoidis: Remove this as well

packages/wallet-core/src/jwt-session/index.ts
```diff
@@ -0,0 +1,59 @@
...
+	session: v.object({
+		appName: v.string(),
+		appOrigin: v.pipe(v.string(), v.url()),
+		appUrl: v.pipe(v.string(), v.url()),
+		accounts: v.array(AccountSchema),
```
pchrysochoidis, line 22: keep only accounts

packages/wallet-core/src/jwt-session/index.ts
pchrysochoidis, line 17: Remove this if it does through with extra props. Same for all other schemas

pchrysochoidis: It doesn't throw it just removes unknown fields

packages/wallet-core/src/web-wallet-channel/responses.ts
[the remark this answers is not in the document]
pchrysochoidis: Discussed this and decided to keep only the jwt

packages/wallet-core/src/web-wallet-channel/wallet-post-message-channel.ts
pchrysochoidis: As discussed keep it hardcoded like this

pchrysochoidis:
I think I addressed all open issues. Please have another look!

hayes-mysten (approved):

PR #195, opened 2025-04-09
Add async thunks to tx.add

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/typescript/src/transactions/Transaction.ts
+++ b/packages/typescript/src/transactions/Transaction.ts
@@ -410,6 +410,9 @@ export class Transaction {
 
 	/** Destructively merges a transaction into the current transaction. The passed transaction will not be usable after merging. */
 	#merge(transaction: Transaction, replace?: number, resultIndex?: number) {
+		console.log(replace, resultIndex);
+		console.dir(transaction.#data.commands, { depth: null });
+		console.dir(this.#data.commands, { depth: null });
 		if (transaction.#lastPendingTask) {
 			throw new Error('Cannot merge transactions that have pending tasks');
 		}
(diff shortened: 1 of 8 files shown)


packages/typescript/src/transactions/Transaction.ts
```diff
@@ -383,15 +391,129 @@ export class Transaction {
...
+		transaction.#intentResolvers.forEach((resolver, intent) => {
+			this.addIntentResolver(intent, resolver);
+		});
+	}
+
 	/** Add a transaction to the transaction */
-	add<T = TransactionResult>(command: Command | ((tx: Transaction) => T)): T {
-		if (typeof command === 'function') {
-			return command(this);
+	add<T extends Command | ((tx: Transaction) => unknown) | AsyncThunk | Transaction>(
```
hayes-mysten, line 470: These types can probably be cleaned up a bit.

tx.add now does a LOT of different things

Jordan-Mysten: Probably easier just to use TS overloads so that we can just think of the cases distinctly?

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -135,6 +137,12 @@ export class Transaction {
 	#serializationPlugins: TransactionPlugin[];
 	#buildPlugins: TransactionPlugin[];
 	#intentResolvers = new Map<string, TransactionPlugin>();
+	#lastPendingTask: Promise<void> | undefined;
```
Jordan-Mysten, line 140: Promise chains are the GOAT

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -383,15 +391,129 @@ export class Transaction {
...
+		fork.#lastPendingTask = this.#lastPendingTask;
+		return fork;
+	}
+
+	#commit(transaction: Transaction) {
+		this.#data = transaction.#data;
+		this.#lastPendingTask = transaction.#lastPendingTask;
+	}
+
+	/** Destructively merges a transaction into the current transaction. The passed transaction will not be usable after merging. */
```
Jordan-Mysten, line 411: > The passed transaction will not be usable after merging

Outside of the scope of this PR, but at some point I really wanted to have a way to "poison" a transaction so that after you passed it to methods like signAndExecute it was unusable, to avoid issues with folks re-using the same TX object and accidentally equivocating objects. Would also be useful for a case like this where after merging we'd be able to mark the passed transaction as unusable.

PR #198, opened 2025-04-11
async thunk sections

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/typescript/src/transactions/Transaction.ts
+++ b/packages/typescript/src/transactions/Transaction.ts
@@ -497,17 +497,10 @@ export class Transaction {
 
 	#resolveArgument(arg: TransactionArgument): Argument {
 		if (typeof arg === 'function') {
-			const resolved = arg(this);
+			const resolved = this.add(arg as never);
 
 			if (typeof resolved === 'function') {
 				return this.#resolveArgument(resolved);
-			}
-
-			if (resolved && typeof resolved === 'object' && 'then' in resolved) {
-				return parse(
-					Argument,
-					this.add(() => resolved),
-				);
 			}
 
 			return parse(Argument, resolved);
--- a/packages/typescript/src/transactions/Transaction.ts
+++ b/packages/typescript/src/transactions/Transaction.ts
@@ -410,6 +410,9 @@ export class Transaction {
 
 	/** Destructively merges a transaction into the current transaction. The passed transaction will not be usable after merging. */
 	#merge(transaction: Transaction, replace?: number, resultIndex?: number) {
+		console.log(replace, resultIndex);
+		console.dir(transaction.#data.commands, { depth: null });
+		console.dir(this.#data.commands, { depth: null });
 		if (transaction.#lastPendingTask) {
 			throw new Error('Cannot merge transactions that have pending tasks');
 		}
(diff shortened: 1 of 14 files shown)


packages/typescript/test/unit/transactions/blob-registration.test.ts
```diff
@@ -0,0 +1,220 @@
...
+			      },
+			      "$kind": "$Intent",
+			    },
+			    {
+			      "$kind": "MoveCall",
+			      "MoveCall": {
+			        "arguments": [
+			          {
+			            "$kind": "Result",
+			            "Result": 2,
```
hayes-mysten, line 111: This is currently broken, it references the right result, but the commands aren't in the right order.  Need to figure out exactly why this case isn't working after getting some sleep

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -613,7 +699,89 @@ export class Transaction {
...
+		const unorderedInputs = this.#data.inputs;
+		const orderedCommands = this.#data.commands
+			.filter((cmd) => cmd.$Intent?.name !== 'AsyncTransactionThunk')
+			.sort((a, b) => {
+				const sectionA = this.#sectionMap.get(a);
+				const sectionB = this.#sectionMap.get(b);
+				const indexA = this.#data.commands.indexOf(a);
+				const indexB = this.#data.commands.indexOf(b);
+
+				return sectionA!.localeCompare(sectionB!) || indexA - indexB;
```
hayes-mysten, line 724: this sorting is pretty inefficient, I can clean this up if we decide we like the general approach

williamrobertson13 (approved):
lgtm!!I think the sections terminology makes sense, and the implementation is clean as well :)

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -383,15 +405,98 @@ export class Transaction {
 		return this.object(Inputs.SharedObjectRef(...args));
 	}
 
+	#fork() {
```
Jordan-Mysten, line 408: I like how light weight this ended up being

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -611,9 +722,92 @@ export class Transaction {
 		};
 
 		await createNext(0)();
+
+		this.#inputSection = this.#data.inputs;
+		this.#commandSection = this.#data.commands;
+	}
+
+	async #waitForPendingTasks() {
```
Jordan-Mysten, line 730: This is clever, never thought about doing it this way.

packages/typescript/src/transactions/Transaction.ts
```diff
@@ -611,9 +722,92 @@ export class Transaction {
...
+
+				return { ...arg, NestedResult: [updated, arg.NestedResult[1]] };
+			}
+
+			return arg;
+		});
 	}
 
 	async prepareForSerialization(options: SerializeTransactionOptions) {
+		await this.#waitForPendingTasks();
```
Jordan-Mysten, line 809: We could _technically_ save one microtask if we conditionally call / await this only when pending promises.length > 0.

packages/walrus/src/client.ts
```diff
@@ -652,51 +652,54 @@ export class WalrusClient {
 	 * tx.transferObjects([await client.createStorage({ size: 1000, epochs: 3 })], owner);
```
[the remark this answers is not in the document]
hayes-mysten: We haven't documented the low level transaction methods anywhere other than comments

packages/typescript/src/transactions/Transaction.ts
hayes-mysten: yeah... gonna leave it as is for now

packages/typescript/src/transactions/Transaction.ts
hayes-mysten: going to leave this as is for now

PR #206, opened 2025-04-14
[dapp-kit-next] add reactive wallets state

williamrobertson13:
## Description

This PR sets up the reactive state for tracking the registered wallets in an app. 

If you look closely in this PR, you'll see a new type in this PR called `UiWallet` which is from the wallet standard's newer `ui` package. This can be pretty confusing to wrap your head around at first, but there's a new concept of immutable handles called `UiWallet` and `UiWalletAccount` which hold a weak reference back to the underlying `Wallet` and `WalletAccount` objects. There are a few benefits to this, IMO one of the main benefits is that you can think of all actions in an app [happening on a specific account](https://github.com/anza-xyz/wallet-standard/pull/51#discussion_r1628427225) which has some authorized set of features/chains/etc. This means we also don't have to manually keep track of the current wallet state since it can be derived from the current account. https://github.com/anza-xyz/wallet-standard/pull/51 has more context (some of the rendering equality changes probably don't matter too much for us fwiw)

Follow-up PR for managing the connection state: https://github.com/MystenLabs/ts-sdks/pull/254

## Test plan
- Manual testing by registering new wallets and un-registering them as well as changing their properties
- Will set up Vitest browser tests 🔜


packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { WalletWithRequiredFeatures } from '@mysten/wallet-standard';
+import { getWallets, StandardEvents } from '@mysten/wallet-standard';
+import { onMount } from 'nanostores';
+import { isSuiWallet, getWalletUniqueIdentifier } from '../../utils/wallets.js';
+import type { DAppKitState } from '../state.js';
+
+export function syncRegisteredWallets({ $wallets }: DAppKitState) {
```
williamrobertson13, line 10: so the way we did this in the original version of dApp-kit was technically speaking kind of wrong; besides keeping track of when wallets get registered and unregistered, we need to make sure we are properly refreshing the state when the properties of a wallet change and clean-up those event listeners accordingly

this code is loosely modeled after https://github.com/wallet-standard/wallet-standard/blob/master/packages/react/core/src/useWallets_INTERNAL_ONLY_NOT_FOR_EXPORT.ts which covers the additional edge cases I mentioned above

packages/dapp-kit-next/src/store/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+type CreateDAppKitStoreOptions = void;
+
+export function createDAppKitStore(_: CreateDAppKitStoreOptions) {
+	const state = createState();
+
+	syncRegisteredWallets(state);
+
+	return {
+		$wallets: readonlyType(state.$wallets),
```
williamrobertson13, line 18: One other point on the `wallets` state, the concept of `preferredWallets` from dapp-kit v1 is purely related to the UI so I don't think that should be taken into account here. I think there's a similar case for `walletFilter` prop in that people should really filter the wallet list on their end or through props on the UI components

packages/dapp-kit-next/src/store/index.ts
```diff
@@ -0,0 +1,37 @@
...
+import { readonlyType } from 'nanostores';
+import { syncRegisteredWallets } from './initializers/registered-wallets.js';
+import { createState } from './state.js';
+
+export type DAppKitStore = ReturnType<typeof createDAppKitStore>;
+
+type CreateDAppKitStoreOptions = void;
+
+export function createDAppKitStore(_: CreateDAppKitStoreOptions) {
+	const state = createState();
```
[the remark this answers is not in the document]
williamrobertson13: nvm, map is the way to go

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,65 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { getWallets, StandardEvents } from '@mysten/wallet-standard';
+import { onMount } from 'nanostores';
+import type { DAppKitState } from '../state.js';
+
+import type { WalletWithRequiredFeatures } from '@mysten/wallet-standard';
+import { getWalletUniqueIdentifier, isSuiWallet } from '../../utils/wallets.js';
+
+export function syncRegisteredWallets($state: DAppKitState) {
```
williamrobertson13, line 11: so the way we did this in the original version of dApp-kit was technically speaking kind of wrong; besides keeping track of when wallets get registered and unregistered, we need to make sure we are properly refreshing the state when the properties of any wallets change and clean-up those event listeners accordingly

this code is loosely modeled after https://github.com/wallet-standard/wallet-standard/blob/master/packages/react/core/src/useWallets_INTERNAL_ONLY_NOT_FOR_EXPORT.ts which covers the additional edge cases I mentioned above

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { getWallets, StandardEvents } from '@mysten/wallet-standard';
+import { onMount } from 'nanostores';
+import type { DAppKitState } from '../state.js';
+
+import type { Wallet, WalletWithRequiredFeatures } from '@mysten/wallet-standard';
+import { getSuiWallets, isSuiWallet } from '../../utils/wallets.js';
+import { getOrCreateUiWalletForStandardWallet_DO_NOT_USE_OR_YOU_WILL_BE_FIRED as getOrCreateUiWalletForStandardWallet } from '@wallet-standard/ui-registry';
```
williamrobertson13, line 10: copied from my other PR:

the DO_NOT_USE_OR_YOU_WILL_BE_FIRED naming is because these functions are only meant to be used by wallet standard UI libraries and not external applications, the naming isn't changed to reflect that yet tho. Since we are really a library on top of the wallet standard, it seems fair game :)

(plz don't fire me)

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,61 @@
...
+			removedWallets.filter(isSuiWallet).forEach((wallet) => {
+				const unsubscribeFromChange = unsubscribeCallbacksByWallet.get(wallet);
+				if (unsubscribeFromChange) {
+					unsubscribeCallbacksByWallet.delete(wallet);
+					unsubscribeFromChange();
+				}
+			});
+
+			const suiWallets = getSuiWallets().map(getOrCreateUiWalletForStandardWallet);
+			$state.setKey('wallets', suiWallets);
```
hayes-mysten, line 46: should we move this to a function that's re-used on `change`

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export function syncRegisteredWallets({ $state }: DAppKitState) {
+	onMount($state, () => {
+		const walletsApi = getWallets();
+		const unsubscribeCallbacksByWallet = new Map<Wallet, () => void>();
+
+		const subscribeToWalletEvents = (wallet: WalletWithRequiredFeatures) => {
+			const unsubscribeFromChange = wallet.features[StandardEvents].on('change', () => {
+				const suiWallets = getSuiWallets().map(getOrCreateUiWalletForStandardWallet);
+				$state.setKey('wallets', suiWallets);
```
hayes-mysten, line 23: We are re-setting the wallets here.  I'm curious what this actually does, and how this impacts our unsubscribeCallbacksByWallet set. 

If the instances are different, then won't this cause issues when we try to look them up later on unregister? If the instance doesn't change, why are we setting them again here

williamrobertson13: Good question! The wallet-standard recycles the array of `Wallet` entities even when properties change, so prevWalletA and nextWalletA will always be referentially equal. That means it is safe to use the `Wallet` object as a key for the unsubscribe callback map throughout this code.

Now for the actual `wallets` state we are setting with Nanostores, we are mapping the underlying wallet instance to an immutable `UiWallet` handle on line 22. These maintain a pointer to the underlying wallet and will be referentially different when property changes occur. So any updates to a wallet or their accounts should result in a state update and will cause a re-render in UI code.

(I realize this isn't very intuitive, maybe let me throw in some comments 😆?)

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,65 @@
...
+		const unsubscribeCallbacksByWallet = new Map<Wallet, () => void>();
+
+		const onWalletsChanged = () => {
+			const suiWallets = getSuiWallets().map(getOrCreateUiWalletForStandardWallet);
+			$state.setKey('wallets', suiWallets);
+		};
+
+		const subscribeToWalletEvents = (wallet: WalletWithRequiredFeatures) => {
+			const unsubscribeFromChange = wallet.features[StandardEvents].on('change', () => {
+				onWalletsChanged();
```
Jordan-Mysten, line 27: Just curious why you need to invoke this when the underlying wallet change?

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
```diff
@@ -0,0 +1,65 @@
...
+	onMount($state, () => {
+		const walletsApi = getWallets();
+		const unsubscribeCallbacksByWallet = new Map<Wallet, () => void>();
+
+		const onWalletsChanged = () => {
+			const suiWallets = getSuiWallets().map(getOrCreateUiWalletForStandardWallet);
+			$state.setKey('wallets', suiWallets);
+		};
+
+		const subscribeToWalletEvents = (wallet: WalletWithRequiredFeatures) => {
```
Jordan-Mysten, line 25: You could also subscribe to changes on $state.wallets and do this there if you wanted to separate these things out (wallets changing vs wallets being added / removed), but if not this is also cool

packages/dapp-kit-next/src/store/initializers/registered-wallets.ts
williamrobertson13: Say you have wallets A, B, and C registered in your app. Wallet B supports `sui:mainnet` and `sui:testnet`, but then all of a sudden the wallet properties change so that now only `solana:testnet` is supported. We don't want to keep that wallet in our `wallets` state anymore because it's not a valid wallet that can be used in a Sui dApp. So we want to make sure our wallets state gets refreshed in this scenario

(This is completely contrived but you get the point 😆 )

PR #214, opened 2025-04-16
[DO NOT MERGE YET][faucet] Update faucet routes

stefan-mysten:
## Description

The faucet will use the new `/v2/gas` routes very soon. This PR updates it in a few places, but ideally I would need @hayes-mysten to have a look to make sure we're covering it everywhere.

Please note that both `/gas` and `/v1/gas` will be deprecated, and an error response will be returned to use `/v2/gas` instead (which is heavily rate limited). Also, the response has changed. See https://github.com/MystenLabs/sui/issues/21775

!!! Not sure how the CI picks up changes from main `sui` repo, but probably this PR: https://github.com/MystenLabs/sui/pull/21865 needs to be merged there first.

## Test plan

---

### DIFF
--- a/packages/typescript/src/faucet/faucet.ts
+++ b/packages/typescript/src/faucet/faucet.ts
@@ -48,7 +48,7 @@ type FaucetRequest = {
 	method: 'GET' | 'POST';
 };
 
-async function faucetRequest({ host, path, body, headers, method }: FaucetRequest) {
+async function faucetRequest<T>({ host, path, body, headers, method }: FaucetRequest): Promise<T> {
 	const endpoint = new URL(path, host).toString();
 	const res = await fetch(endpoint, {
 		method,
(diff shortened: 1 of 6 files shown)


hayes-mysten (approved):
This just needs a changeset, and then it should be good to go

packages/typescript/src/faucet/index.ts
```diff
@@ -4,6 +4,7 @@
 export {
 	requestSuiFromFaucetV0,
 	requestSuiFromFaucetV1,
+	requestSuiFromFaucetV2,
 	getFaucetRequestStatus,
```
hayes-mysten, line 8: you deleted this in the file above, was that intentional?

stefan-mysten: Ah yes, we need to keep it for now, but it should be deprecated later because there's no more `/v1/status` after the new routes are updated.

stefan-mysten: Re-added

packages/typescript/src/faucet/faucet.ts
```diff
@@ -103,6 +150,27 @@ export async function requestSuiFromFaucetV1(input: {
 	});
 }
 
+export async function requestSuiFromFaucetV2(input: {
+	host: string;
+	recipient: string;
+	headers?: HeadersInit;
+}): Promise<FaucetResponseV2> {
+	return faucetRequestV2({
```
hayes-mysten, line 158: Is there a reason to have this extra method here?  We can probably just flatten this right?

packages/typescript/src/faucet/faucet.ts
```diff
@@ -67,6 +78,39 @@ async function faucetRequest({ host, path, body, headers, method }: FaucetReques
 	}
 }
 
+async function faucetRequestV2({ host, path, body, headers, method }: FaucetRequest) {
```
hayes-mysten, line 81: Is this any different than the one above?  We should be able to reuse the same method for both versions

packages/typescript/src/faucet/faucet.ts
hayes-mysten: ignore this, we should just use faucetRequest right?

packages/typescript/src/faucet/faucet.ts
stefan-mysten: Yes, error code is different. There's no more "error" field in the response.

PR #224, opened 2025-04-17
Update passkey.mdx

v1xingyue:
Recover passkey bug . provider should be passed as second param.


hayes-mysten (commented):
thanks for the fix!

PR #227, opened 2025-04-17
wip - add client middleware stack

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/typescript/src/experimental/middleware.ts
+++ b/packages/typescript/src/experimental/middleware.ts
@@ -1,0 +1,83 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { SuiClient } from '../client/index.js';
+import type { Experimental_SuiClientTypes } from './types.js';
+
+type MiddlewareFunction = <Args extends unknown[], Return extends unknown>(
+	scope: string,
+	method: string,
+	args: Args,
+	next: (...args: Args) => Promise<Return>,
+) => Promise<Return>;
+
+export class ClientMiddlewareStack {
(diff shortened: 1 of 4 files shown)


packages/typescript/src/experimental/transports/jsonRPC.ts
```diff
@@ -23,37 +23,41 @@ export class JSONRpcTransport extends Experimental_CoreClient {
...
-		for (const batch of batches) {
-			const objects = await this.#jsonRpcClient.multiGetObjects({
-				ids: batch,
-				options: {
-					showOwner: true,
-					showType: true,
-					showBcs: true,
-				},
-			});
+	getObjects = this.root.middleware.wrap(
```
hayes-mysten, line 30: I don't love making all of these instance properties... We could also do

```ts
getObjects(options) {
  return this.root.middleware.run('core', 'getObjects', this, [options], (options) => {
     // contents
  })
}
```

packages/typescript/src/experimental/middleware.ts
```diff
@@ -0,0 +1,83 @@
...
+		const stack = this;
+		return function <T>(this: T, ...args: Args) {
+			return stack.run(scope, name, fn, this, args);
+		};
+	}
+
+	run<Args extends unknown[], Return extends unknown>(
+		scope: string,
+		name: string,
+		fn: (...args: Args) => Promise<Return>,
```
hayes-mysten, line 38: this should be the last arg

packages/typescript/src/experimental/transports/jsonRPC.ts
Jordan-Mysten: Yeah neither do I, I'd probably do it this way

PR #244, opened 2025-04-23
feat(seal): add import and export session key

joyqvq:
## Description

add feat to import and export session key, useful to persist in web storage e.g. indexed db

## Test plan

unit test

---

### DIFF
--- a/packages/seal/src/session-key.ts
+++ b/packages/seal/src/session-key.ts
@@ -65,6 +65,9 @@ export class SessionKey {
 			throw new UserError(`Invalid TTL ${ttlMin}, must be between 1 and 30`);
 		}
 
+		if (signer && signer.getPublicKey().toSuiAddress() !== address) {
+			throw new UserError('Signer address does not match session key address');
+		}
 		this.#address = address;
 		this.#packageId = packageId;
 		this.#creationTimeMs = Date.now();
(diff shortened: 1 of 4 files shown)


packages/seal/src/session-key.ts
```diff
@@ -135,4 +143,47 @@ export class SessionKey {
...
+			},
+		});
+		return obj;
+	}
+
+	/**
+	 * Restore a SessionKey instance for the given object.
+	 * @returns A new SessionKey instance with restored state
+	 */
+	static import(data: SessionKeyType): SessionKey {
```
hayes-mysten, line 173: We should probably accept the signer (and any other constructor args as a second optional options argument here).  Then we can verify the imported address/signature against the signer if its provided

packages/seal/src/session-key.ts
```diff
@@ -30,6 +29,15 @@ export type Certificate = {
 	signature: string;
 };
 
+export type SessionKeyType = {
+	address: string;
+	packageId: string;
+	creationTimeMs: number;
+	ttlMin: number;
+	personalMessageSignature?: string;
+	sessionKey: string;
```
hayes-mysten, line 38: We should add a sessionKey scheme for forwards compatibility.  It can just be typed as 'ED25519' for now, and we can throw in import if it's anything else

joyqvq: looks like the getSecretKey im using already encode the flag/scheme into it. should we just reuse that for decoding? https://github.com/MystenLabs/ts-sdks/blob/b0d3074f4af8665b21795794880c423af3c09868/packages/typescript/src/keypairs/ed25519/keypair.ts#L131

https://github.com/MystenLabs/ts-sdks/blob/03975f4ade68ae5f4206c2b0818bafe8a7de071e/packages/typescript/src/cryptography/keypair.ts#L125

packages/seal/src/session-key.ts
joyqvq: do we expect the signer to be used? iiuc this is just for testing, for a dapp it should rather use the setPersonalMessageSignature after getting it from wallet hook.

> any other constructor args as a second optional options argument

personalMessageSignature is already optional, you mean, making creationTimeMs also optional? but it should always have it, right? 

the personalMessageSignature is verified when setting it, so its checked to be consitent with the creationTimeMs, ttlMin, packageId and session key pk

hayes-mysten:
```ts
static import(data: SessionKeyType, { signer }: { signer?: Signer }): SessionKey {
```

I think we do want people to use signers for imported session keys.  There are a few cases for this in dapps:
* zklogin users
* dapps using an enoki signer
* passkey signers
* We will eventually have a wallet signer that implements the signer interface using the wallet standard

Signers allow a session key to refresh itself, and imort/export allow sessions to be persisted.  I could see an argument for not needing to persist sessions if you can automatically refresh it with the signer, but for things like enoki this will reduce the number of API calls required, and for the wallet signer it would reduce the number of times a user is prompted for a wallet interaction

joyqvq: added!

hayes-mysten (approved):

PR #246, opened 2025-04-23
[faucet] Move to the latest faucet v2

stefan-mysten:
## Description

This PR updates the test suite to use the latest faucet v2. This new version includes a simplified local faucet which now uses `/v2/gas` as the standard route.

Before this can be merged/used, we need to also update the `sui-tools` image to refer to a commit that has the new local faucet changes in the main `sui-repo`.

## Test plan

---


hayes-mysten (approved):

stefan-mysten:
Thanks @hayes-mysten for fixing it and merging the changes.

PR #252, opened 2025-04-25
Check key server version

jonas-lj:
## Description

Add a minimal key server version, and compare with the version from the response header.

## Test plan

Unit tests.

---

### DIFF
--- a/packages/seal/package.json
+++ b/packages/seal/package.json
@@ -47,6 +47,7 @@
 		"@hey-api/openapi-ts": "^0.66.1",
 		"@mysten/build-scripts": "workspace:*",
 		"@types/node": "^22.13.17",
+		"@types/semver": "^7.7.0",
 		"typescript": "^5.8.2",
 		"vitest": "^3.1.1"
 	},
@@ -55,6 +56,7 @@
 		"@mysten/sui": "workspace:*",
 		"@noble/curves": "^1.8.1",
 		"@noble/hashes": "^1.7.1",
+		"semver": "^7.7.0",
 		"shamir-secret-sharing": "0.0.3"
 	}
 }
--- a/packages/seal/package.json
+++ b/packages/seal/package.json
@@ -47,7 +47,6 @@
 		"@hey-api/openapi-ts": "^0.66.1",
 		"@mysten/build-scripts": "workspace:*",
 		"@types/node": "^22.13.17",
-		"@types/semver": "^7.7.0",
 		"typescript": "^5.8.2",
 		"vitest": "^3.1.1"
 	},
@@ -56,7 +55,6 @@
 		"@mysten/sui": "workspace:*",
 		"@noble/curves": "^1.8.1",
 		"@noble/hashes": "^1.7.1",
-		"semver": "^7.7.0",
 		"shamir-secret-sharing": "0.0.3"
 	}
 }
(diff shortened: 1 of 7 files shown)


packages/seal/package.json
```diff
@@ -55,6 +56,7 @@
 		"@mysten/sui": "workspace:*",
 		"@noble/curves": "^1.8.1",
 		"@noble/hashes": "^1.7.1",
+		"semver": "^7.7.0",
```
hayes-mysten, line 59: We have been working towards eliminating as many runtime dependencies as we can.  I think the use-case here is simple enough that it would be better to just write a simple custom utility function to do this for us instead

packages/seal/src/key-server.ts
```diff
@@ -115,3 +120,18 @@ export async function verifyKeyServer(server: KeyServer, timeout: number): Promi
...
+ *
+ * @param response - The response from the key server.
+ */
+export function verifyKeyServerVersion(response: Response) {
+	const keyServerVersion = response.headers.get('X-KeyServer-Version');
+	if (
+		keyServerVersion == null ||
+		new Version(keyServerVersion).older_than(SERVER_VERSION_REQUIREMENT)
+	) {
+		throw new InvalidKeyServerError('Key server version not found');
```
benr-ml, line 135: also in case of old version, maybe change to
```suggestion
		throw new InvalidKeyServerError('Key server version not supported');
```

packages/seal/src/utils.ts
```diff
@@ -37,3 +37,33 @@ export function createFullId(dst: Uint8Array, packageId: string, innerId: string
...
+	// Compare this version with another version. True if this version is older than the other version.
+	older_than(other: Version): boolean {
+		if (this.major !== other.major) {
+			return this.major < other.major;
+		} else if (this.minor !== other.minor) {
+			return this.minor < other.minor;
+		}
+		return this.patch < other.patch;
+	}
+}
```
benr-ml, line 69: let's add tests that:
- 0.1.0.older_than(SERVER_VERSION_REQUIREMENT) and the opposite is false
- test per major, minor, patch

packages/seal/src/key-server.ts
```diff
@@ -6,13 +6,15 @@ import { bls12_381 } from '@noble/curves/bls12-381';
 import { KeyServerMove } from './bcs.js';
 import {
 	InvalidGetObjectError,
+	InvalidKeyServerError,
```
benr-ml, line 9: InvalidKeyServerError is currently returned when the user does not set the key servers correctly, let's add a different error InvalidKeyServerVersionError

packages/seal/test/unit/key-server.test.ts
```diff
@@ -111,4 +115,24 @@ describe('key-server tests', () => {
...
+		expect(version2.older_than(version1)).toBe(false);
+		expect(version1.older_than(version1)).toBe(false);
+		expect(version1.older_than(version3)).toBe(true);
+		expect(version3.older_than(version1)).toBe(false);
+		expect(version1.older_than(version4)).toBe(true);
+		expect(version4.older_than(version1)).toBe(false);
+
+		// Update when SERVER_VERSION_REQUIREMENT changes
+		expect(version0.older_than(SERVER_VERSION_REQUIREMENT)).toBe(true);
+		expect(version1.older_than(SERVER_VERSION_REQUIREMENT)).toBe(false);
```
benr-ml, line 136:
```suggestion
		expect(SERVER_VERSION_REQUIREMENT.older_than(version0)).toBe(false);
		expect(SERVER_VERSION_REQUIREMENT.older_than(SERVER_VERSION_REQUIREMENT)).toBe(false);
```

benr-ml (approved):

PR #254, opened 2025-04-25
[dapp-kit-next] add functionality for managing connection state

williamrobertson13:
## Description

Follow-up on https://github.com/MystenLabs/ts-sdks/pull/206 for supporting connecting, disconnecting, and switching accounts. There are some tricky edge cases here to handle - auto-connect functionality will be in a follow-up PR!

## Test plan
- Tested that you are disconnected if your wallet gets unregistered from the UI
- Tested that you are disconnected if the properties (`chains`) on your wallet change in such a way that make it incompatible
- Tested that you are disconnected if the properties on your ONLY account change in such a way that make it incompatible
- Tested that you are defaulted to the first account in your wallet if the properties on your selected account change in such a way that make in incompatible
- Tested connecting, disconnecting, switching accounts under the usual scenarios

---


packages/dapp-kit-next/src/store/actions/connect-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import type { DAppKitState } from '../state.js';
+import { task } from 'nanostores';
+import type { UiWallet } from '@wallet-standard/ui';
+import type { StandardConnectInput } from '@mysten/wallet-standard';
+import type { StandardConnectFeature } from '@mysten/wallet-standard';
+import { isSuiChain, StandardConnect } from '@mysten/wallet-standard';
+import { getWalletFeature } from '@wallet-standard/ui';
+import {
+	getOrCreateUiWalletAccountForStandardWalletAccount_DO_NOT_USE_OR_YOU_WILL_BE_FIRED as getOrCreateUiWalletAccountForStandardWalletAccount,
```
williamrobertson13, line 12: the `DO_NOT_USE_OR_YOU_WILL_BE_FIRED` naming is because these functions are only meant to be used by wallet standard UI libraries and not external applications, the naming isn't changed to reflect that yet tho

packages/dapp-kit-next/src/store/index.ts
```diff
@@ -4,19 +4,55 @@
...
+						supportedIntents: state.supportedIntents,
+						connectionStatus: state.connectionStatus,
+						isConnected: true,
+						isConnecting: false,
+						isDisconnected: false,
+					} as const;
+				case 'connecting':
+					return {
+						account: state.currentAccount,
+						supportedIntents: state.supportedIntents,
```
hayes-mysten, line 37: do we have this while connecting?

packages/dapp-kit-next/src/store/actions/disconnect-wallet.ts
```diff
@@ -0,0 +1,43 @@
...
+				throw new WalletNotConnectedError('No wallet is connected.');
+			}
+
+			try {
+				const { disconnect } = getWalletFeature(
+					currentAccount,
+					StandardDisconnect,
+				) as StandardDisconnectFeature[typeof StandardDisconnect];
+
+				await disconnect(...standardDisconnectArgs);
```
hayes-mysten, line 32: should we have a disconnecting state?

packages/dapp-kit-next/src/store/actions/connect-wallet.ts
```diff
@@ -0,0 +1,61 @@
...
+export type ConnectWalletArgs = {
+	/** The wallet to connect to. */
+	wallet: UiWallet;
+} & Omit<StandardConnectInput, 'silent'>;
+
+export function connectWalletCreator({ $state }: DAppKitState) {
+	/**
+	 * Prompts the specified wallet to connect and authorize new accounts for the current domain.
+	 */
+	return async function connectWallet({ wallet, ...standardConnectArgs }: ConnectWalletArgs) {
```
hayes-mysten, line 26: should our connect support a preferred account/address as an alternative to selecting the first account?

packages/dapp-kit-next/src/store/actions/disconnect-wallet.ts
hayes-mysten: probably not worth it, since unlike connect there isn't a user interaction it might be waiting on

hayes-mysten (approved):

packages/dapp-kit-next/src/store/index.ts
williamrobertson13: the types are inferred so I just used the variables instead of manually specifying `null`

image

packages/dapp-kit-next/src/store/actions/disconnect-wallet.ts
williamrobertson13: Good thought, I think you could make a case just for completeness but yeah tbh since disconnecting in 99% of cases is nearly instant I think leaving it out for now is okay. I'm not strongly opinionated either way

PR #255, opened 2025-04-25
fix: use passed in graphql client

joyqvq:
## Description

to pass in graphql client optionally, if the personal message sig needs to be verified as zklogin.

## Test plan

---


packages/seal/src/session-key.ts
```diff
@@ -96,14 +96,11 @@ export class SessionKey {
 		return new TextEncoder().encode(message);
 	}
 
-	async setPersonalMessageSignature(personalMessageSignature: string) {
+	async setPersonalMessageSignature(personalMessageSignature: string, client?: SuiGraphQLClient) {
```
hayes-mysten, line 99: You can now update this to use https://github.com/MystenLabs/ts-sdks/blob/main/packages/typescript/src/zklogin/publickey.ts#L19 but I think you need to export it here first: https://github.com/MystenLabs/ts-sdks/blob/main/packages/typescript/src/zklogin/index.ts#L14

hayes-mysten: I think it would probably be better to pass the client in the constructor instead

joyqvq: revised! PTAL. 

i patched sui sdk and seal sdk in one changeset, checking if this is fine cc @hayes-mysten

hayes-mysten (approved):

PR #263, opened 2025-04-28
[seal] batched key server requests

notmatical:
## Description

Deduplicates the passed object of key server object ID's so they are only fetched one time, and then returns the preserved array including the fetched content.

## Test plan

unit tests

---

### DIFF
--- a/.changeset/slimy-plants-add.md
+++ b/.changeset/slimy-plants-add.md
@@ -2,4 +2,4 @@
 '@mysten/seal': patch
 ---
 
-introduced deduplication of key server object fetching
+introduced batched requests for key server objects


notmatical:
looks like `validateEncryptionServices` integration test is failing, i am doing some investigation as to why now.

packages/seal/src/client.ts
```diff
@@ -321,9 +327,11 @@ export class SealClient {
 					receivedIds.size === expectedIds.size &&
 					[...receivedIds].every((id) => expectedIds.has(id));
 
-				// Return early if the completed servers is more than threshold.
+				// Count each occurrence of this servers objectId from the original keyServers array.
 				if (hasAllKeys) {
-					completedServerCount++;
+					const occurrences = keyServers.filter((ks) => ks.objectId === server.objectId).length;
+					completedServerCount += occurrences;
```
hayes-mysten, line 333: this doesn't seem right

hayes-mysten: This might be correct if the comment above is correct that we want to could duplicate servers towards the threshold

packages/seal/src/key-server.ts
```diff
@@ -83,6 +86,14 @@ export async function retrieveKeyServers({
 			};
 		}),
 	);
+
+	return objectIds.map((objectId) => {
+		if (!fetchedServers[objectId]) {
+			throw new InvalidGetObjectError(`KeyServer ${objectId} not found`);
+		}
+
+		return fetchedServers[objectId];
```
hayes-mysten, line 95: rather than making all the changes to to other file, you could keep this change more contained, and just create separate server objects here

hayes-mysten (changes requested):
I think this overall seems worth while, but we probably want to add some test coverage for these changes

.changeset/slimy-plants-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mysten/seal': minor
```
benr-ml, line 2: let's use a patch

notmatical: Resolved, just got back from vacation.

jonas-lj:
Duplicates is not an issue after #293 has been merged, since `retrieveKeyServers` is now called without any duplicates among the object ID's. Duplicates are instead handled by adding weights to key servers. So I don't think this PR is needed anymore.

notmatical:
Correct, your PR is a solid approach. However, this should still be viable in terms of batching the rpc requests for object data. Previously we used getObject for each passed ID, now we take all the passed ID's and make use of getObjects

jonas-lj:
Ah, yes. Then this still makes sense, but you don't need to use a Set to avoid duplicates.

notmatical:
@jonas-lj latest should reflect the changes upstream from #293.

PR #265, opened 2025-04-28
Align with errors from Seal key server

jonas-lj:
## Description

Add new errors from version alignment between sdk and server.

Also reverts an accidental change to pnpm-lock.yaml from https://github.com/MystenLabs/seal/pull/92.

## Test plan

N/A

---

### DIFF
--- a/packages/seal/src/error.ts
+++ b/packages/seal/src/error.ts
@@ -95,7 +95,7 @@ export class InvalidSessionKeySignatureError extends SealAPIError {
 	}
 }
 
-// Errors returned by the Seal server that indicate that the SDK version is invalid or deprecated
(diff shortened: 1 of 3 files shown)


packages/seal/src/error.ts
```diff
@@ -91,6 +95,20 @@ export class InvalidSessionKeySignatureError extends SealAPIError {
 	}
 }
 
+// Errors returned by the Seal server that indicate that the SDK version is invalid or deprecated
```
benr-ml, line 98:
```suggestion
// Errors returned by the Seal server that indicate that the SDK version is invalid (implying that HTTP headers used by the SDK are being removed) or deprecated (implying that the SDK should be upgraded).
```

benr-ml (approved):

PR #271, opened 2025-04-29
Correctly handle byteOffset in Uint8Arrays when reading bcs bytes

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/bcs/src/reader.ts
+++ b/packages/bcs/src/reader.ts
@@ -41,7 +41,7 @@ export class BcsReader {
 	 * @param {Uint8Array} data Data to use as a buffer.
 	 */
 	constructor(data: Uint8Array) {
-		this.dataView = new DataView(data.buffer);
+		this.dataView = new DataView(data.buffer, data.byteOffset, data.byteLength);
 	}
 	/**
 	 * Shift current cursor position by `bytes`.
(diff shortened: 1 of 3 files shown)


Jordan-Mysten (approved):
Good catch, didn't even know this existed

packages/bcs/src/reader.ts
```diff
@@ -42,6 +42,7 @@ export class BcsReader {
 	 */
 	constructor(data: Uint8Array) {
 		this.dataView = new DataView(data.buffer);
+		this.bytePosition = data.byteOffset;
```
williamrobertson13, line 45: should we change the `start` calc in `readBytes` and `readUleb` to not factor in the dataView byteOffset since we're intentionally initializing the data view with a clean slate? nbd either way just thought it might be more clear

williamrobertson13 (approved):
lgtm!

packages/bcs/src/reader.ts
hayes-mysten: I made dataView work like the uint8array by setting the byteOffset and byteLength on the view directly.  That way bytePosition doesn't change and is 0 indexed regardless of the buffer, and everything else stays the same

PR #275, opened 2025-04-30
[dapp-kit-next] add auto-connect functionality

williamrobertson13:
## Description

This PR wires up auto-connect functionality and fixes some minor issues with `disconnect` (this was targeting accounts not the wallet which was wrong) and `connect` (need to introduce a new `reconnecting` state since you can authorize accounts multiple times). 

## Test plan
- Tested manually


packages/dapp-kit-next/src/store/actions/connect-wallet.ts
```diff
@@ -36,7 +36,7 @@ export function connectWalletCreator({ $state }: DAppKitState) {
 	}: ConnectWalletArgs) {
 		return await task(async () => {
 			try {
-				$state.setKey('connectionStatus', 'connecting');
+				$state.setKey('connection.status', 'connecting');
```
williamrobertson13, line 39: primer: I switched the state to a `deepMap` because doing this is a bad pattern as it'll cause listeners to fire on specific keys that haven't really changed. I chose this over separating out the `wallets` state and `connection` state into two separate atoms because then there's a potential footgun where we add new state but accidentally forget to add it to whatever computed state and then the UI components don't re-render properly

```
$state.set({
  ..$state.get(),
  relatedKey1: 'abc',
  relatedKey2: 'def',
});
```

packages/dapp-kit-next/src/store/actions/disconnect-wallet.ts
```diff
@@ -7,33 +7,39 @@ import type { DAppKitState } from '../state.js';
...
+				: null;
+
+			if (!currentWallet) {
 				throw new WalletNotConnectedError('No wallet is connected.');
 			}
 
 			try {
 				const { disconnect } = getWalletFeature(
-					currentAccount,
+					currentWallet,
```
williamrobertson13, line 33: disconnecting should happen on the wallet and not the app - this was an oversight

packages/dapp-kit-next/src/store/actions/disconnect-wallet.ts
```diff
@@ -7,33 +7,39 @@ import type { DAppKitState } from '../state.js';
...
 					StandardDisconnect,
 				) as StandardDisconnectFeature[typeof StandardDisconnect];
 
 				await disconnect(...standardDisconnectArgs);
+			} catch (error) {
```
williamrobertson13, line 38: misunderstood how `try` / `finally` works 😅

packages/dapp-kit-next/src/store/index.ts
```diff
@@ -5,52 +5,91 @@ import { computed, readonlyType } from 'nanostores';
...
 	return {
 		...actions,
-		$state: readonlyType(state.$state),
-		$wallets: computed(state.$state, (state) => state.wallets),
-		$currentWallet: state.$currentWallet,
-		$currentAccount: computed([state.$state], (state) => {
-			switch (state.connectionStatus) {
+		$state: readonlyType($state),
+		$wallets: computed($state, (state) => state.wallets),
+		$currentAccount: computed([$state], ({ connection, wallets }) => {
```
williamrobertson13, line 59: I don't know what the best API is to expose to folks for the connection info. In the current version of dApp Kit, we have all the connection status fields on `currentWallet` and the `currentAccount` hook just maybe returns you the account. I flipped this when I added the initial connection state, but I think it would probably be best to combine these so you get the same level of type-safety. Idk if `currentAccount` / `currentWallet` / `connection` is a better name, I'm kind of thinking (2) or (3) since the wallet owns the account

packages/dapp-kit-next/src/store/initializers/autoconnect-wallet.ts
```diff
@@ -0,0 +1,82 @@
...
+
+import { listenKeys, onMount, task } from 'nanostores';
+import type { DAppKitState } from '../state.js';
+import type { StateStorage } from '../../utils/storage.js';
+import type { UiWallet } from '@wallet-standard/ui';
+
+/**
+ * Attempts to connect to a previously authorized wallet account on mount and when new wallets are registered.
+ */
+export function autoConnectWallet({
```
williamrobertson13, line 12: Okay so I played around with making the code for managing your connection + auto-connect functionality more declarative, but idk if this is better or worse than doing the equivalent logic imperatively in `syncRegisteredWallets`. The pro with this is that `syncRegisteredWallets` become this huge state manager for the whole library, and the con is that the code could maybe be a little more un-predictable / we have to throw in some extra checks

packages/dapp-kit-next/src/store/initializers/autoconnect-wallet.ts
```diff
@@ -0,0 +1,75 @@
...
+					storageKey,
+					wallets,
+				});
+			});
+
+			if (savedWalletAccount) {
+				$state.setKey('connection', {
+					status: 'connected',
+					currentAccount: savedWalletAccount,
+					supportedIntents: [],
```
williamrobertson13, line 38: In a world where we don't call `connect` during auto-connect and rely on wallets pre-authorizing accounts, I think we need to figure out what to do with `supportedIntents` here:
1. Get rid of supported intents from dApp Kit and worry about it later when we have a practical use case come up
2. If we do want to keep supported intents, should intents be specific to accounts or the wallet itself?

Jordan-Mysten: Easy answer is to just shove it on a property of the wallet or as a feature.

packages/dapp-kit-next/src/store/initializers/autoconnect-wallet.ts
```diff
@@ -0,0 +1,75 @@
...
+		return null;
+	}
+
+	const [savedWalletName, savedAccountAddress] = savedWalletNameAndAddress.split(':');
+	if (!savedWalletName || !savedAccountAddress) {
+		return null;
+	}
+
+	for (const wallet of wallets) {
+		if (wallet.name === savedWalletName) {
```
williamrobertson13, line 65: JFYI: I know we introduced an optional `id` field on the wallet interface in our wallet-standard, uuuuh idk I was kind of sanity checking if that is still necessary since I couldn't think of a practical scenario where you'd want to register two "Sign in with Google" Enoki wallets 🤔 

It's ez to change this around if we like, was just trying to simplify

Jordan-Mysten: Yeah I'm not sure, I think the ID is hypothetically more stable but also the slush wallet has different IDs and different names between the platforms so IDK.

Jordan-Mysten (approved):

packages/dapp-kit-next/src/utils/storage.ts
```diff
@@ -0,0 +1,40 @@
...
+		setItem(key: string, value: string) {
+			store.set(key, value);
+		},
+		removeItem(key: string) {
+			store.delete(key);
+		},
+	};
+}
+
+function isLocalStorageAvailable() {
```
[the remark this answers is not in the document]
williamrobertson13: I was trying to make this a little more robust (followed https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#testing_for_availability)

packages/dapp-kit-next/src/utils/storage.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export type StateStorage = {
+	getItem: (name: string) => string | null | Promise<string | null>;
+	setItem: (name: string, value: string) => unknown | Promise<unknown>;
+	removeItem: (name: string) => unknown | Promise<unknown>;
+};
+
+export const DEFAULT_STORAGE_KEY = 'sui-dapp-kit:wallet-connection-info';
```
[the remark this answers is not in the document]
williamrobertson13: picked a new key!

packages/dapp-kit-next/src/store/initializers/sync-state-to-storage.ts
```diff
@@ -0,0 +1,30 @@
...
+	$state,
+	storage,
+	storageKey,
+}: {
+	$state: DAppKitState;
+	storage: StateStorage;
+	storageKey: string;
+}) {
+	onMount($state, () => {
+		return listenKeys($state, ['connection', 'connection.currentAccount'], ({ connection }) => {
```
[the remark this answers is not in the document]
williamrobertson13: Yeah we do, it's kinda weird. I'll double check this one more time

packages/dapp-kit-next/src/store/initializers/sync-state-to-storage.ts
```diff
@@ -0,0 +1,30 @@
...
+			if (connection.currentAccount) {
```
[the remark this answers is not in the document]
williamrobertson13: Sorry might not fully understand this. With this version of auto-connecting, there isn't really a need (or at least I don't think there's a need) to keep a separate status for auto-connecting since we don't asynchronously call the `connect` feature. Getting the values from storage is technically async but in 99% of use cases it'll be instant.

So mount (disconnected) -> register wallet -> auto-connect logic activates -> instantly connected

PR #276, opened 2025-05-01
[seal] Add getter to cached keys

jonas-lj:
## Description

Add a getter to the cached keys. This is needed for on-chain decryption.

## Test plan

Unit test.

---

### DIFF
--- a/packages/seal/src/client.ts
+++ b/packages/seal/src/client.ts
@@ -140,6 +140,7 @@ export class SealClient {
 		txBytes: Uint8Array;
 	}) {
 		const encryptedObject = EncryptedObject.parse(data);
+
 		this.#validateEncryptionServices(
 			encryptedObject.services.map((s) => s[0]),
 			encryptedObject.threshold,
@@ -151,6 +152,7 @@ export class SealClient {
 			sessionKey,
 			threshold: encryptedObject.threshold,
 		});
+
 		return decrypt({ encryptedObject, keys: this.#cachedKeys });
 	}
 
--- a/packages/seal/src/client.ts
+++ b/packages/seal/src/client.ts
@@ -345,4 +345,17 @@ export class SealClient {
 			throw toMajorityError(errors);
 		}
 	}
+
+	/**
+	 * Return the derived key for a given id and key server if it is in the cache.
+	 *
+	 * @param packageId The id of the package.
+	 * @param id The id of the derived key.
+	 * @param serverObjectId The object id for the key server.
+	 * @returns The derived key for the given id and server if it in the cache (see fetchKey) and `undefined` otherwise.
+	 */
+	getKey(packageId: string, id: string, serverObjectId: string): G1Element | undefined {
+		const fullId = createFullId(DST, packageId, id);
+		return this.#cachedKeys.get(`${fullId}:${serverObjectId}`);
+	}
 }
--- a/packages/seal/src/client.ts
+++ b/packages/seal/src/client.ts
@@ -358,7 +358,20 @@ export class SealClient {
 		return encryptedObject;
 	}
 
-	async getKemKeys({
+	/**
+	 * Get the derived keys needed to decrypt the encrypted object.
+	 *
+	 * Calls fetchKeys in case one or more of the required keys is not cached yet.
+	 * The function throws an error if the client's key servers are not a subset of
+	 * the encrypted object's key servers (including the same weights) or if the
+	 * threshold cannot be met.
+	 *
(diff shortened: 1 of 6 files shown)


packages/seal/src/client.ts
```diff
@@ -345,4 +345,17 @@ export class SealClient {
 			throw toMajorityError(errors);
 		}
 	}
+
```
benr-ml, line 348: It'd be hard to use this API since `fetchKeys` returns when it has enough derived keys, not necessarily all keys or the one requested here.

some options:
- I believe that in all the use cases we have in mind, we should be able to decrypt also in the sdk, right? if so, maybe this should be an extension of `decrypt`, i.e., the caller can call `decrypt` and also get the used derived keys? for that we would need also the encryption though
- expose something similar to decrypt, e.g.,` getKemKeys(sessionKey, txBytes, threshold) ` which returns keys from cache or calls `fetchKeys`

jonas-lj: Yea, I see your point. The user could call `fetchKeys` with t = n before get keys, but it's indeed inconvenient.

Isn't the `getKemKeys` idea basically just returning the keys from `fetchKeys`? Or more precisely, writing a wrapper around `fetchKeys` that does this.

packages/seal/src/client.ts
```diff
@@ -345,4 +345,46 @@ export class SealClient {
...
+	 * @param id - The id of the encrypted object.
+	 * @param services - The object ids of the key servers to use.
+	 * @param txBytes - The transaction bytes to use (that calls seal_approve* functions).
+	 * @param sessionKey - The session key to use.
+	 * @param threshold - The threshold.
+	 * @returns - Derived keys for the given services that are in the cache as a Service -> Key map. If the call is succesful, at least threshold keys will be returned.
+	 */
+	async getDerivedKeys({
+		id,
+		services,
```
benr-ml, line 361: I think we can just use the client's services, no?

packages/seal/src/client.ts
```diff
@@ -345,4 +345,46 @@ export class SealClient {
...
+		txBytes,
+		sessionKey,
+		threshold,
+	}: {
+		id: string;
+		services: string[];
+		txBytes: Uint8Array;
+		sessionKey: SessionKey;
+		threshold: number;
+	}): Promise<Map<string, G1Element>> {
```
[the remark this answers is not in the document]
jonas-lj: You're right. The return value is fixed now. The client logic is hard-coded to G1Element, but we can refactor that when needed.

packages/seal/src/client.ts
```diff
@@ -345,4 +350,57 @@ export class SealClient {
...
+		sessionKey,
+		threshold,
+	}: {
+		id: string;
+		txBytes: Uint8Array;
+		sessionKey: SessionKey;
+		threshold: number;
+	}): Promise<Map<string, DerivedKey>> {
+		const keyServers = await this.getKeyServers();
+		this.#validateEncryptionServices(
```
benr-ml, line 375: shouldn't this be 
`if (threshold > this.#serverObjectIds.length) { .. }`

packages/seal/src/client.ts
```diff
@@ -345,4 +350,57 @@ export class SealClient {
...
+	/**
+	 * Get derived keys from the given services.
+	 *
+	 * @param id - The id of the encrypted object.
+	 * @param txBytes - The transaction bytes to use (that calls seal_approve* functions).
+	 * @param sessionKey - The session key to use.
+	 * @param threshold - The threshold.
+	 * @returns - Derived keys for the given services that are in the cache as a "service object ID" -> derived key map. If the call is succesful, exactly threshold keys will be returned.
+	 */
+	async getDerivedKeys({
```
benr-ml, line 363: should it also receive 
`kemType = KemType.BonehFranklinBLS12381DemCCA,`
?

packages/seal/src/client.ts
```diff
@@ -345,4 +350,57 @@ export class SealClient {
...
+		id,
+		txBytes,
+		sessionKey,
+		threshold,
+	}: {
+		id: string;
+		txBytes: Uint8Array;
+		sessionKey: SessionKey;
+		threshold: number;
+	}): Promise<Map<string, DerivedKey>> {
```
benr-ml, line 373: this will break once we add pq kem, no?

jonas-lj: Replaced with an opaque type.

benr-ml (approved):

PR #277, opened 2025-05-05
[dapp-kit-next] some housekeeping and renaming to get the API closer to the final state

williamrobertson13:
## Description

This PR slightly changes how dApp Kit instances get created, namely when you call `createDAppKit` the instance will get attached to the global on the first invocation and then subsequent invocations will just return you a newly store.

As part of this, I decided to do some renaming to maybe make naming and concepts easier. I think it was a little confusing that a `nanostore` store is an atom, but our dApp Kit store is a collection of many stores.
  - `/store` -> `/core` just for generality
  - `createDAppKitStore` -> `createDAppKit`

## Test plan
---
- Tested in Storybook + the example app


packages/dapp-kit-next/src/components/dapp-kit-connect-modal.ts
```diff
@@ -18,10 +18,10 @@ export class DAppKitConnectModal extends ScopedRegistryHost(LitElement) {
 	};
 
 	@storeProperty()
-	store?: DAppKitStore;
+	store?: DAppKit;
 
-	override willUpdate() {
```
williamrobertson13, line 23: didn't realize that properties are binded in `connectedCallback` so this feels less bad

packages/dapp-kit-next/src/components/dapp-kit-connect-modal.ts
```diff
@@ -18,10 +18,10 @@ export class DAppKitConnectModal extends ScopedRegistryHost(LitElement) {
...
-		this.store ||= getDefaultStore();
+	override connectedCallback() {
+		this.store ||= createDAppKit();
```
williamrobertson13, line 24: oh wait this is wrong

packages/dapp-kit-next/src/core/index.ts
```diff
@@ -5,11 +5,37 @@ import { computed, readonlyType } from 'nanostores';
...
+let defaultInstance: DAppKit | undefined;
+
+export function createDAppKit(options: CreateDAppKitOptions) {
+	const dAppKit = createDAppKitInstance(options);
+
+	if (!defaultInstance) {
+		defaultInstance = dAppKit;
+
+		globalThis.__DEFAULT_DAPP_KIT_INSTANCE__ ||= defaultInstance;
+		if (globalThis.__DEFAULT_DAPP_KIT_INSTANCE__ !== defaultInstance) {
```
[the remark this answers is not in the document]
Jordan-Mysten: Should we also warn if defaultInstance is defined and you call this again?

packages/dapp-kit-next/src/core/index.ts
```diff
@@ -34,11 +35,35 @@ type CreateDAppKitStoreOptions = {
...
+			console.warn('Detected multiple dApp-kit instances. This may cause un-expected behavior.');
+		}
+	}
+
+	return dAppKit;
+}
+
+export function getDefaultInstance() {
+	if (!defaultInstance) {
```
Jordan-Mysten, line 56: Shouldn't this read the global too in case there's multiple dapp kit vesions?

Jordan-Mysten (approved):

packages/dapp-kit-next/src/core/index.ts
williamrobertson13: uuuuuh yeah I reckon you're correct, will fix

packages/dapp-kit-next/src/core/index.ts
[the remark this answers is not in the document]
williamrobertson13: Thought on this for a bit, I guess I don't really need `defaultInstance` at all if we just want `getDefaultInstance` to resolve to whatever is attached to the global. The instances are actually technically isolated (well assuming you specified different storage keys), but the intention is that you only ever create one instance period so yeah we can warn more generally if you call `createDAppKit` and there's already an instance on the global

PR #281, opened 2025-05-06
Metadata fallback for Slush

claykohut-mysten:
## Description

Adds hardcoded metadata fallback for Slush wallet


packages/slush-wallet/src/wallet/index.ts
```diff
@@ -42,6 +42,14 @@ export const SLUSH_WALLET_NAME = 'Slush' as const;
 const SUI_WALLET_EXTENSION_ID = 'com.mystenlabs.suiwallet' as const;
 const METADATA_API_URL = 'https://api.slush.app/api/wallet/metadata';
 
+const FALLBACK_METADATA = {
+	id: 'com.mystenlabs.suiwallet.web',
+	walletName: 'Slush Wallet',
```
Jordan-Mysten, line 47: Should just be "Slush"

packages/slush-wallet/src/wallet/index.ts
```diff
@@ -339,6 +347,7 @@ export async function registerSlushWallet(
 		metadata = await fetchMetadata(metadataApiUrl);
 	} catch (error) {
 		console.error('Error fetching metadata', error);
+		metadata = FALLBACK_METADATA;
```
Jordan-Mysten, line 350: I wonder if we should fetch in the background so that we can register with the static metadata first and have this metadata fetched async and used to update the already-registered wallet (or unregister if enabled: false).

We don't have to do it this way either just spitballing

claykohut-mysten: @Jordan-Mysten updated to use this approach - still need to test

packages/slush-wallet/src/wallet/index.ts
```diff
@@ -334,25 +348,26 @@ export async function registerSlushWallet(
...
+		metadata: FALLBACK_METADATA,
 	});
-
 	const unregister = wallets.register(slushWalletInstance);
 
+	fetchMetadata(metadataApiUrl)
+		.then((metadata) => {
+			if (!metadata.enabled) {
+				console.log('Slush wallet is not currently enabled.');
+				unregister();
```
claykohut-mysten, line 362: May need to check if wallet has already been unregistered (to handle edge case)

packages/slush-wallet/src/wallet/index.ts
```diff
@@ -306,6 +314,12 @@ export class SlushWallet implements Wallet {
...
 	}
+
+	updateMetadata(metadata: WalletMetadata) {
+		this.#id = metadata.id;
+		this.#walletName = metadata.walletName;
+		this.#icon = metadata.icon as WalletIcon;
+	}
 }
 
 async function fetchMetadata(metadataApiUrl: string): Promise<WalletMetadata> {
```
Jordan-Mysten, line 325: It doesn't look like this needs to be async anymore?

packages/slush-wallet/src/wallet/index.ts
```diff
@@ -334,25 +348,26 @@ export async function registerSlushWallet(
...
 	const slushWalletInstance = new SlushWallet({
 		name,
 		origin,
-		metadata,
+		metadata: FALLBACK_METADATA,
 	});
-
 	const unregister = wallets.register(slushWalletInstance);
 
+	fetchMetadata(metadataApiUrl)
```
Jordan-Mysten, line 358: Given we're no longer awaiting this, does the parent function still need to be async?

claykohut-mysten: no, updated.

packages/slush-wallet/src/wallet/index.ts
claykohut-mysten: looks like checking whether it's unregistered already isn't necessary. Correct me if I'm wrong @Jordan-Mysten

Jordan-Mysten: Looks like it's safe to call multiple times

Jordan-Mysten (approved):

PR #282, opened 2025-05-06
Remove object and balance changes

Jordan-Mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/zksend/src/links/get-sent-transactions.ts
+++ b/packages/zksend/src/links/get-sent-transactions.ts
@@ -16,6 +16,7 @@ export async function getSentTransactionsWithLinks({
 	network = 'mainnet',
 	contract = getContractIds(network),
 	client = new SuiClient({ url: getFullnodeUrl(network) }),
+	loadAssets = true,
 	loadClaimedAssets = false,
 	options,
 	...linkOptions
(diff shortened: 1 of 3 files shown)


claykohut-mysten (approved):

packages/zksend/src/links/get-sent-transactions.ts
```diff
@@ -140,7 +147,7 @@ export async function getSentTransactionsWithLinks({
 							...linkOptions,
 						});
 
-						await link.loadAssets({ loadClaimedAssets });
+						await link.loadAssets({ loadAssets, loadClaimedAssets });
```
williamrobertson13, line 150: I'm still kinda confused, the `loadAssets` boolean param doesn't do anything in `loadAssets` function (which is kind of also un-intuitive). do we mean to just conditionally call this method?

williamrobertson13: this naming is confusing but this seems fineeee from tracing the code, should we just only call `loadAssets` when the `loadAssets` param option is true?

williamrobertson13 (approved):
no context on the outage but yeah this seems reasonable

packages/zksend/src/links/get-sent-transactions.ts
Jordan-Mysten: This still does the `loadBagObject` which we probably want

PR #286, opened 2025-05-08
[walrus] clean up write from wallet example

daniellam258:
## Description

- Remove unnecessary `getFundedKeypair()` in write from wallet in Walrus SDK
- use `setSender()` to ensure the sender is not blank when submitting transaction

---


Jordan-Mysten (approved):
Thanks so much for this!

PR #290, opened 2025-05-09
[seal] Clean ups and simplifications

jonas-lj:
## Description

Cleaning up seal codebase. Simplifies some flows in the client, but there should be no functional changes.

## Test plan

N/A

---

### DIFF
--- a/packages/seal/src/client.ts
+++ b/packages/seal/src/client.ts
@@ -164,7 +164,7 @@ export class SealClient {
 	#validateEncryptionServices(services: string[], threshold: number) {
 		// Check that the client's key servers are a subset of the encrypted object's key servers.
 		if (
-			this.#serverObjectIds.some(
+			services.some(
 				(objectId) => count(this.#serverObjectIds, objectId) < count(services, objectId),
 			)
 		) {
--- a/packages/seal/src/client.ts
+++ b/packages/seal/src/client.ts
@@ -163,10 +163,12 @@ export class SealClient {
 
 	#validateEncryptionServices(services: string[], threshold: number) {
 		// Check that the client's key servers are a subset of the encrypted object's key servers.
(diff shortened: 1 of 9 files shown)


packages/seal/src/client.ts
```diff
@@ -158,20 +158,14 @@ export class SealClient {
...
-		}
-		for (const [objectId, count] of serverObjectIdsMap) {
-			if (servicesMap.get(objectId) !== count) {
-				throw new InconsistentKeyServersError(
-					`Client's key servers must be a subset of the encrypted object's key servers`,
-				);
-			}
+		if (
+			this.#serverObjectIds.some(
+				(objectId) => count(this.#serverObjectIds, objectId) < count(services, objectId),
```
jonas-lj, line 163: This was an inequality check, but iiuc checking that there are at least as many should be sufficient.

packages/seal/src/decrypt.ts
```diff
@@ -106,11 +99,6 @@ async function combine(shares: { index: number; share: Uint8Array }[]): Promise<
 
 	// The Shamir secret sharing library expects the index/x-coordinate to be at the end of the share
 	return externalCombine(
-		shares.map(({ index, share }) => {
-			const packedShare = new Uint8Array(share.length + 1);
-			packedShare.set(share, 0);
-			packedShare[share.length] = index;
-			return packedShare;
-		}),
+		shares.map(({ index, share }) => flatten([share, new Uint8Array([index])])),
```
benr-ml, line 102: is `share` fixed size?

jonas-lj: Yes, it's a share of the base key which is fixed fo 32 bytes.

benr-ml (approved):

PR #300, opened 2025-05-14
Add signAndExecute to Signer and use core API for transaction resolution

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/typescript/src/experimental/transports/json-rpc-resolver.ts
+++ b/packages/typescript/src/experimental/transports/json-rpc-resolver.ts
@@ -3,16 +3,19 @@
 
 import { parse } from 'valibot';
 
-import type { BcsType } from '../bcs/index.js';
-import { bcs } from '../bcs/index.js';
-import type { SuiClient } from '../client/client.js';
-import { normalizeSuiAddress, normalizeSuiObjectId, SUI_TYPE_ARG } from '../utils/index.js';
-import { ObjectRef } from './data/internal.js';
-import type { Argument, CallArg, Command, OpenMoveTypeSignature } from './data/internal.js';
-import { Inputs } from './Inputs.js';
-import { getPureBcsSchema, isTxContext, normalizedTypeToMoveTypeSignature } from './serializer.js';
-import type { TransactionDataBuilder } from './TransactionData.js';
+import { normalizeSuiAddress, normalizeSuiObjectId, SUI_TYPE_ARG } from '../../utils/index.js';
+import { ObjectRef } from '../../transactions/data/internal.js';
+import type { CallArg, Command, OpenMoveTypeSignature } from '../../transactions/data/internal.js';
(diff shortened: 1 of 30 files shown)


packages/typescript/src/experimental/transports/json-rpc-resolver.ts
```diff
@@ -21,56 +24,37 @@ const MAX_OBJECTS_PER_FETCH = 50;
...
+	return async function resolveTransactionData(
+		transactionData: TransactionDataBuilder,
+		options: BuildTransactionOptions,
+		next: () => Promise<void>,
+	) {
+		await normalizeInputs(transactionData, client);
+		await resolveObjectReferences(transactionData, client);
+
+		if (!options.onlyTransactionKind) {
+			await setGasPrice(transactionData, client);
```
Jordan-Mysten, line 37: I know this is just moved code but should these be under a Promise.all?

hayes-mysten: I think they are all dependent on each other.  I think dryRun (budget) depends on the gas price.  I think setting the payment used to account for the budget, but doesn't anymore so we could do that in parallel.  If we are going to change that, I think I'd rather do it separately

williamrobertson13 (approved):

packages/typescript/src/experimental/types.ts
```diff
@@ -192,10 +192,14 @@ export namespace Experimental_SuiClientTypes {
...
 		digest: string;
 		signatures: string[];
+		epoch: string | null;
 		effects: TransactionEffects;
-		// TODO: Return parsed data:
-		// We need structured representations events, and transaction data
-		bcs: Uint8Array;
+		objectTypes: PromiseLike<Record<string, string>>;
+		transaction: {
+			data: SerializedTransactionDataV2;
```
hayes-mysten, line 199: we could flatten this so transaction is data + bcs, which is sortof what we do for effects.  It's slightly different because I am re-using an existing type here

hayes-mysten: decided to flatten it

packages/typescript/src/experimental/transports/jsonRPC.ts
```diff
@@ -186,15 +186,27 @@ export class JSONRpcTransport extends Experimental_CoreClient {
...
 			transaction: {
 				digest: await tx.getDigest(),
-				effects: parseTransactionEffectsJson({
-					effects: result.effects,
-					objectChanges: result.objectChanges,
-				}),
+				epoch: null,
+				effects,
+				objectTypes: {
+					get then() {
```
Jordan-Mysten, line 200: Any reason why you can't just do this statically?

hayes-mysten: Nope, just didn't update it after refactoring... I had originally tried to do this a bit differently

PR #301, opened 2025-05-15
Use MVR in Deepbook SDK

tonylee08:
## Description

Use MVR in Deepbook SDK

## Test plan

Tested SDK functionality on testnet


hayes-mysten:
This looks like it will break if you don't add the mvr plugin to each transaction right?  We should be careful about how this gets rolled out so it doesn't break existing apps without warning

tonylee08:
> This looks like it will break if you don't add the mvr plugin to each transaction right? We should be careful about how this gets rolled out so it doesn't break existing apps without warning

@hayes-mysten Yes that's right. We're going to make this a major version change when merged, and add in docs to show how to properly use the new version

0xaslan (approved):

tonylee08:
> IMO we shouldn't start using MVR in our core libraries until it's enabled by default. If we feel comfortable doing that then we can start to go down that route.

Agree won't be merging this change until we've had discussions

tonylee08:
> How urgent is this change? We are working on new documentation for SDKs to standardize how SDKs should be built, which will include a lot of new patterns that will require some breaking changes.
> 
> I haven't looked into mvr patterns yet, but that should probably be part of those recommendation. I think it would result in less ecosystem churn if we could try to do all of that together.
> 
> I am also a little concerned about how users will be expected to configure mvr caches from multiple SDKs.
> 
> If this doesn't need to go out immediately, I think we would be better off combining some of these changes so we only have one major release with breaking changes, and ensuring we have a better story around how mvr resolution for SDKs should actually work.
> 
> Manually adding the plugin to each transaction is pretty painful. There is support for global transaction plugins, which might be a simple solution to that specific problem, but I'd want to see what this looks like for users already using mvr, and we at least have a good story for combining multiple mvr caches

This is definitely not urgent, and optional. It’s more to set an example for external partners on how to use MVR in SDKs effectively. Agree that bundling this with standardized SDK patterns would make more sense. Happy to hold off and revisit

PR #302, opened 2025-05-15
Add wallet-signer

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/dapp-kit-next/src/current-account-signer.ts
+++ b/packages/dapp-kit-next/src/current-account-signer.ts
@@ -3,16 +3,16 @@
 
 import type { PublicKey, SignatureScheme } from '@mysten/sui/cryptography';
 import { SIGNATURE_FLAG_TO_SCHEME, Signer } from '@mysten/sui/cryptography';
-import type { DAppKitStore } from './store/index.js';
+import type { DAppKit } from './core/index.js';
 import type { Transaction } from '@mysten/sui/transactions';
 import type { Experimental_SuiClientTypes } from '@mysten/sui/experimental';
 import { parseTransactionBcs, parseTransactionEffectsBcs } from '@mysten/sui/experimental';
 import { toBase64, fromBase64 } from '@mysten/sui/utils';
 
 export class CurrentAccountSigner extends Signer {
-	#store: DAppKitStore;
+	#store: DAppKit;
 
-	constructor(store: DAppKitStore) {
+	constructor(store: DAppKit) {
 		super();
 		this.#store = store;
 	}
--- a/packages/dapp-kit-next/src/current-account-signer.ts
+++ b/packages/dapp-kit-next/src/current-account-signer.ts
@@ -1,0 +1,80 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { PublicKey, SignatureScheme } from '@mysten/sui/cryptography';
+import { SIGNATURE_FLAG_TO_SCHEME, Signer } from '@mysten/sui/cryptography';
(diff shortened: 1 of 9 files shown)


packages/typescript/src/zklogin/publickey.ts
```diff
@@ -56,6 +56,47 @@ export class ZkLoginPublicIdentifier extends PublicKey {
 		}
 	}
 
+	static fromBytes(
```
hayes-mysten, line 59: This is a little nasty, but we get the (normalized) publicKey bytes + address from the wallet standard.  To get a proper publcKey instance for a zklogin address, we need both of these, since the bytes are always the non-legacy bytes.

packages/dapp-kit-next/src/store/index.ts
```diff
@@ -56,6 +57,13 @@ export function createDAppKitStore({
 		...actions,
 		$state: readonlyType($state),
 		$wallets: computed($state, (state) => state.wallets),
+		$publicKey: computed($state, (state) =>
```
hayes-mysten, line 60: I'm not sure where this actually belongs, but we need a memoized way to get the publicKey because it's a little expensive to derive

packages/dapp-kit-next/src/current-account-signer.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { PublicKey, SignatureScheme } from '@mysten/sui/cryptography';
+import { SIGNATURE_FLAG_TO_SCHEME, Signer } from '@mysten/sui/cryptography';
+import type { DAppKitStore } from './store/index.js';
+import type { Transaction } from '@mysten/sui/transactions';
+import type { Experimental_SuiClientTypes } from '@mysten/sui/experimental';
+import { parseTransactionBcs, parseTransactionEffectsBcs } from '@mysten/sui/experimental';
+import { toBase64, fromBase64 } from '@mysten/sui/utils';
+
+export class CurrentAccountSigner extends Signer {
+	#store: DAppKitStore;
```
williamrobertson13, line 13: I think you need a rebase on main, some of the naming has changed (e.g. just `DAppKit` now as opposed to `DAppKitStore`

packages/dapp-kit-next/src/current-account-signer.ts
```diff
@@ -0,0 +1,80 @@
...
+		return publicKey;
+	}
+
+	sign(_data: Uint8Array): never {
+		throw new Error(
+			'WalletSigner does not support signing directly. Use signTransaction or signPersonalMessage instead',
+		);
+	}
+
+	async signTransaction(bytes: Uint8Array) {
```
williamrobertson13, line 42: Oh in my head, I thought this was going to be inverted so that dApp Kit used an internal signer to implement signXXX methods and then folks could plug their own signer instance. This makes sense though for being able to compose signers

packages/dapp-kit-next/src/store/index.ts
williamrobertson13: I'm not _opposed_ to this being here since we can return any computed state we find useful for consumers. Other options would be to just memo-ize the function call in the account signer OR tack this onto the `account` object in the `connection` state

Not sure which is best, if it's expensive to derive maybe this or just memo-izing the function call in the signer

hayes-mysten: going to leave it for now, and then it can be moved around as needed as we add derived state for things like currentAccount?

packages/dapp-kit-next/src/current-account-signer.ts
hayes-mysten: I originally was thinking that too, but I think it will be simpler in the direction

PR #304, opened 2025-05-16
[dapp-kit-next] hook up the basic dialog UI

williamrobertson13:
## Description

This PR adds the base modal UI logic and fixes up the Storybook a bit. It turns out building a dialog even with a `dialog` element is hard LOL. The modal implementation is ripped (inspired) by https://material-web.dev/components/dialog. I still probably need to look into accessibility but that'll come later

## Test plan

image

---


packages/dapp-kit-next/src/components/dapp-kit-connect-modal.ts
```diff
@@ -4,15 +4,16 @@
...
 import { customElement } from 'lit/decorators.js';
 import { storeProperty } from '../utils/lit.js';
 import { WalletList } from './internal/wallet-list.js';
 import { getDefaultInstance } from '../core/index.js';
 import type { DAppKit } from '../core/index.js';
+import { BaseModal } from './internal/base-modal.js';
 
 @customElement('mysten-dapp-kit-connect-modal')
-export class DAppKitConnectModal extends ScopedRegistryHost(LitElement) {
+export class DAppKitConnectModal extends ScopedRegistryHost(BaseModal) {
```
williamrobertson13, line 16: this worked better than composing base modal component because I ran into some weird issues trying to query the inner modal from this component - probably fine for now

packages/dapp-kit-next/src/components/internal/base-modal.ts
```diff
@@ -0,0 +1,156 @@
...
+			this.setAttribute('open', '');
+			this.show();
+		} else {
+			this.removeAttribute('open');
+			this.close();
+		}
+	}
+
+	@query('dialog')
+	private readonly _dialog!: HTMLDialogElement;
```
williamrobertson13, line 46: TS decorators don't support `#` wa wa

packages/dapp-kit-next/src/components/internal/base-modal.ts
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { LitElement } from 'lit';
+import { property, query } from 'lit/decorators.js';
+
+export class BaseModal extends LitElement {
```
williamrobertson13, line 7: this is fully compliant with https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog

packages/dapp-kit-next/src/components/dapp-kit-connect-button.ts
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import '@webcomponents/scoped-custom-element-registry';
```
williamrobertson13, line 4: I'm not sure the best way to expose this polyfill for consumers, right now it's just imported in this component and the modal component

packages/dapp-kit-next/src/components/dapp-kit-connect-button.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { DAppKit } from '../core/index.js';
+import type { DAppKitConnectModal } from './dapp-kit-connect-modal.js';
+
+@customElement('mysten-dapp-kit-connect-button')
+export class DAppKitConnectButton extends LitElement {
+	@storeProperty()
+	store?: DAppKit;
+
+	@query('mysten-dapp-kit-connect-modal')
+	private readonly _modal!: DAppKitConnectModal;
```
Jordan-Mysten, line 19: Just curious can we not use true privates?

packages/dapp-kit-next/src/components/internal/base-modal.ts
Jordan-Mysten: I _think_ you're wrong if you use spec decorators instead of the experimental decorators. I think you have this project setup to use the experimental ones, but lit also supports spec ones (you might just need to add a few `accessor` prefixes to properties)

packages/dapp-kit-next/src/components/internal/base-modal.ts
```diff
@@ -0,0 +1,156 @@
...
+	#isOpen = false;
+	#isOpening = false;
+
+	#resolveIsConnectedPromise = () => {};
+	#isConnectedPromise = this.#getIsConnectedPromise();
```
Jordan-Mysten, line 12: Could also use the withResolvers stuff we do elsewhere

Jordan-Mysten (approved):

packages/dapp-kit-next/src/components/internal/base-modal.ts
williamrobertson13: yah Lit recommends experimental decorators for prod (https://lit.dev/docs/components/decorators/)

image

packages/dapp-kit-next/src/components/internal/base-modal.ts
williamrobertson13: I tried but LitElement super classes can't be abstract unfortunately. I could prob re-write this as a mixin or get the has-a composition to work later at some point

image

packages/dapp-kit-next/src/components/dapp-kit-connect-button.ts
```diff
@@ -0,0 +1,42 @@
...
+import '@webcomponents/scoped-custom-element-registry';
+
+import { html, LitElement } from 'lit';
+import { customElement, query } from 'lit/decorators.js';
+import { storeProperty } from '../utils/lit.js';
+import { getDefaultInstance } from '../core/index.js';
+import type { DAppKit } from '../core/index.js';
+import type { DAppKitConnectModal } from './dapp-kit-connect-modal.js';
+
+@customElement('mysten-dapp-kit-connect-button')
```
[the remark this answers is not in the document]
williamrobertson13: yeah @hayes-mysten brought up this Q in my original scaffold PR and I thought it wouldn't hurt to be extra explicit

TO BE FAIR though, people will eventually be able to use scoped registries to self register these web components (I need to export them) so I don't think it really matters if we prefix it with `mysten-` or not

PR #305, opened 2025-05-16
[dapp-kit-next] re-organize the state to be bottom-up and keep networks <> wallets in sync

williamrobertson13:
## Description

I realized there were a few issues with the organization of the Nanostore state, so this PR is attempt to remedy those issues:
  - The list of wallets you can interact with really depends on the configured networks your dApp supports, so we shouldn't let people connect to wallets (i.e. testnet only wallets) for apps that only support mainnet. This is more relevant if your app supports switching networks
  - This uncovers another (not practical at all) issue where the way I'm tracking wallet state is incorrect; you could have non-Sui wallets change properties and technically become Sui wallets 😆. So we really have a concept of `registered` wallets from the standard and `compatible` wallets in the dApp which include a minimum feature set + support the networks configured by the dApp
  - There are places where I want to access derived state from within actions like `connectWallet` that was only being returned from the main dApp kit instance. Having `createStores` return all the computed data up front is nice
  - The deepMap API in nanostores is a little non-ideal in that it has trouble handling generics and listening for key changes on nested state is a un-intuitive (i.e., need to listen to connection and connection.currentAccount to see when the account changes)

## Test plan
- 👁️ in a test dApp


packages/dapp-kit-next/src/core/initializers/autoconnect-wallet.ts
```diff
@@ -11,17 +11,19 @@ import { getWalletUniqueIdentifier } from '../../utils/wallets.js';
...
-	state: DAppKitState;
+	stores: DAppKitStores;
 	storage: StateStorage;
 	storageKey: string;
 }) {
-	onMount($state, () => {
-		return listenKeys($state, ['wallets'], async ({ connection, wallets }, oldValue) => {
-			if (oldValue.wallets.length > wallets.length) return;
+	onMount($compatibleWallets, () => {
+		return $compatibleWallets.listen(async (wallets, oldWallets) => {
```
Jordan-Mysten, line 23: Just confirming: we went this direction because this means if this mounts before the wallet is registered that the autoconnect still works, right? Some dapps actually ran into that as a bug so glad we're fixing that in the next version!

packages/dapp-kit-next/src/core/store.ts
```diff
@@ -0,0 +1,118 @@
...
+import { atom, computed, map } from 'nanostores';
+import { getChain } from '../utils/networks.js';
+import type { Networks } from '../utils/networks.js';
+import { getAssociatedWalletOrThrow, requiredWalletFeatures } from '../utils/wallets.js';
+import { publicKeyFromSuiBytes } from '@mysten/sui/verify';
+import type { Experimental_BaseClient } from '@mysten/sui/experimental';
+
+type WalletConnection =
+	| {
+			status: 'disconnected' | 'connecting';
```
Jordan-Mysten, line 14: Does this mean that to switch wallets the flow is forced to be:
`connected` -> `disconnected` -> `connecting` -> `connected`

packages/dapp-kit-next/src/core/store.ts
```diff
@@ -0,0 +1,118 @@
...
+			currentAccount: null;
+	  }
+	| {
+			status: 'reconnecting' | 'connected';
```
Jordan-Mysten, line 18: What does reconnecting mean?

packages/dapp-kit-next/src/core/store.ts
```diff
@@ -0,0 +1,118 @@
...
+>;
+
+export function createStores<TNetworks extends Networks>({
+	defaultNetwork,
+	getClient,
+}: {
+	defaultNetwork: TNetworks[number];
+	getClient: (network: TNetworks[number]) => Experimental_BaseClient;
+}) {
+	const $baseConnection = map<WalletConnection>({
```
Jordan-Mysten, line 33: The only downside to having multiple stores is that it's a _little bit_ easier to have state drift if something unexpected happens (method throws). In practice this just means we need to think about these cases and test for them to ensure this doesn't happen, so not really a concern for this PR but just something to keep in mind since atom state changes can be disconnected now.

Jordan-Mysten (approved):

packages/dapp-kit-next/src/core/initializers/autoconnect-wallet.ts
williamrobertson13: yep exactly!

packages/dapp-kit-next/src/core/store.ts
williamrobertson13: yep

packages/dapp-kit-next/src/core/store.ts
williamrobertson13: I snuck this in 🎃. So based on the convo we had about providing a "manage connection" sort of UI that just calls `connect` again, I found it better to explicitly distinguish "I'm connecting to a wallet for the first time" and "I'm re-connecting (authorizing more) accounts from the same wallet"

Jordan-Mysten: okie fine with me!

PR #307, opened 2025-05-16
[dapp-kit-next] add support for signing/executing transactions

williamrobertson13:
## Description

This PR adds support for signing and executing transactions. This is kind of just the basic version, still a bit TBD on how we might change the APIs here to play nicely with sponsorship and different execution models and whatnot.

## Test plan
- Tested manually


packages/dapp-kit-next/src/core/actions/sign-and-execute-transaction.ts
```diff
@@ -0,0 +1,59 @@
...
+		const suiClient = $currentClient.get();
+		const chain = getChain(suiClient.network);
+
+		const signAndExecuteTransactionFeature = getAccountFeature({
+			account,
+			chain,
+			featureName: SuiSignAndExecuteTransaction,
+		}) as SuiSignAndExecuteTransactionFeature[typeof SuiSignAndExecuteTransaction];
+
+		return await signAndExecuteTransactionFeature.signAndExecuteTransaction({
```
hayes-mysten, line 42: We should probably still support signAndExecuteTransactionBlock.  Previously we used the signAndExecuteTransaction function exported from the wallet-standard package to automatically fallback to the best supported feature

packages/dapp-kit-next/src/core/actions/sign-transaction.ts
```diff
@@ -0,0 +1,53 @@
...
+		return await signTransactionFeature.signTransaction({
+			...standardArgs,
+			transaction: {
+				toJSON: async () => {
+					if (typeof transaction === 'string') {
+						return transaction;
+					}
+
+					// TODO: Fix passing through the client and supported intents for plugins.
+					transaction.setSenderIfNotSet(account.address);
```
Jordan-Mysten, line 45: This behavior is net new right? (it's good / correct, just confirming)

williamrobertson13: yep this is new (at your request)!

packages/dapp-kit-next/src/core/actions/sign-and-execute-transaction.ts
hayes-mysten: @Jordan-Mysten curious about your take here.  If all wallets are updated it's probably fine 🤷

williamrobertson13: Yah I might revisit this before we release the library, just want to get this landed for now since it's been open forever. It's like < 20 lines of code so I really don't care either way 😆

PR #312, opened 2025-05-22
[seal] CCA secure protocol

jonas-lj:
## Description

Update to the new CCA secure protocol.
Integration tests currently fail because these are breaking changes and the CI servers are not yet updated.

## Test plan

---


packages/seal/src/client.ts
```diff
@@ -340,12 +339,11 @@ export class SealClient {
 						continue;
 					}
 					this.#cachedKeys.set(`${fullId}:${server.objectId}`, keyElement);
-					receivedIds.add(fullId);
 				}
 
 				// Check if all the receivedIds are consistent with the requested fullIds.
 				// If so, consider the key server got all keys and mark as completed.
-				if (fullIds.every((fullId) => receivedIds.has(fullId))) {
+				if (fullIds.every((fullId) => this.#cachedKeys.has(`${fullId}:${server.objectId}`))) {
```
jonas-lj, line 346: This is unrelated to the CCA sec changes, but is to avoid the redundant `receivedIds` map.

benr-ml (approved):

PR #315, opened 2025-05-22
ts codegen

hayes-mysten:
## Description

Example generated code: https://github.com/MystenLabs/ts-sdks/pull/317

## Test plan

---


packages/codegen/src/index.ts
```diff
@@ -0,0 +1,64 @@
...
+			await mkdir(`./tests/generated/deps/${moduleAddress}`, { recursive: true });
+			await writeFile(
+				`./tests/generated/deps/${moduleAddress}/${moduleName}.ts`,
+				builder.toString('./', `./deps/${moduleAddress}/${moduleName}.ts`),
+			);
+		}
+	}
+}
+
+Promise.all(
```
williamrobertson13, line 60: I think we forgot this in here haha

packages/codegen/src/render-types.ts
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	const filteredTypeParameters = typeParameters.filter(
+		(_type, i) => !handle.type_parameters[i].is_phantom,
+	);
+
+	switch (options.format) {
+		case 'typescriptArg':
+			return 'string';
+		// return typeParameters.length === 0
```
williamrobertson13, line 262: can we remove this commented out code?

williamrobertson13 (approved):
lgtm - all looks pretty sensible and organized to me

packages/codegen/src/render-types.ts
hayes-mysten: yes I think so.  I'm not entirely sure what the plan here was.  In think eventually it would be nice to have stricter types for results of other commands, but that will require some more work

PR #322, opened 2025-05-27
fix(seal): require suiClient in Session Key constructor

benr-ml:
## Description

## Test plan

---


packages/seal/src/client.ts
```diff
@@ -430,4 +432,37 @@ export class SealClient {
...
+
+	/**
+	 * Create a new session key with defined Sui Client.
+	 * @param address - The sender adress
+	 * @param packageId - The package ID for Seal approval
+	 * @param ttlMin - The TTL of the session key.
+	 * @param signer - The optional signer to use.
+	 * @returns The session key.
+	 */
+	createSessionKey(
```
benr-ml, line 444: this only saves passing the suiclient, no? people could still use the direct new SessionKey and import

packages/seal/src/types.ts
```diff
@@ -1,9 +1,17 @@
...
+	Experimental_CoreClient,
+	Experimental_SuiClientTypes,
+} from '@mysten/sui/experimental';
 
 export type KeyCacheKey = `${string}:${string}`;
 export type SealCompatibleClient = ClientWithExtensions<{
-	core: Experimental_CoreClient;
+	core: Experimental_CoreClient & {
+		verifyZkLoginSignature: NonNullable<
+			Experimental_SuiClientTypes.TransportMethods['verifyZkLoginSignature']
```
benr-ml, line 14: what is the implication of this? today we can use the standard sui client

packages/seal/src/client.ts
joyqvq: correct

packages/seal/src/types.ts
joyqvq: yes you can use the standard sui client. this change is needed to make SealCompatibleClient compatible with ZkLoginCompatibleClient

packages/seal/src/client.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import type { Signer } from '@mysten/sui/dist/cjs/cryptography/keypair.js';
```
hayes-mysten, line 4: @mysten/sui/cryptography
```suggestion
import type { Signer } from '@mysten/sui/cryptography';
```

hayes-mysten (approved):

benr-ml (commented):
Looks good, let's update the documentation (UsingSeal)

PR #326, opened 2025-05-28
fix(seal): Use versioned key server registry

joyqvq:
## Description

a future upgrade to key server should look like: 

1. upgrade key_server.move contract with a new struct KeyServerV2 with additional fields. and update functions to: create_v2, create_and_transfer_v2
2. all key servers can call create_and_transfer_v2, can also be newly onboarded key servers starts with v2 (last version).
3. new sdk releases with KeyServerMoveV2 and use new parser KeyServerMoveV2.parse()

note that the object ids in getAllowlistedKeyServers do not need to change. older sdk will continue to work with existing V1 objects. 

## Test plan

---


packages/seal/src/key-server.ts
```diff
@@ -73,18 +75,51 @@ export async function retrieveKeyServers({
...
+			let resVersionedKs;
+			try {
+				resVersionedKs = await client.core.getDynamicField({
+					parentId: objectId,
+					name: {
+						type: 'u64',
+						bcs: bcs.u64().serialize(EXPECTED_SERVER_VERSION).toBytes(),
+					},
+				});
+			} catch (e) {
```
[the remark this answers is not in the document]
joyqvq: good point, removed all the catch block since they never should happen

packages/seal/src/key-server.ts
```diff
@@ -73,18 +75,51 @@ export async function retrieveKeyServers({
 			} catch (e) {
 				throw new InvalidGetObjectError(`KeyServer ${objectId} not found; ${(e as Error).message}`);
 			}
-
 			const ks = KeyServerMove.parse(res.object.content);
-			if (ks.keyType !== 0) {
-				throw new UnsupportedFeatureError(`Unsupported key type ${ks.keyType}`);
+
+			if (ks.firstVersion !== EXPECTED_SERVER_VERSION.toString()) {
```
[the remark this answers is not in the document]
joyqvq: here it should be a range check, but before parsing it needs the exactly version. i moved the check into a if block right when parsing below to be clear and extensible in future.

.changeset/dark-rooms-play.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mysten/seal': minor
```
benr-ml, line 2: let's use a patch

packages/seal/src/key-server.ts
```diff
@@ -62,30 +63,48 @@ export async function retrieveKeyServers({
...
+			// Then fetch the expected versioned object and parse it.
+			const resVersionedKs = await client.core.getDynamicField({
+				parentId: objectId,
+				name: {
+					type: 'u64',
+					bcs: bcs.u64().serialize(EXPECTED_SERVER_VERSION).toBytes(),
+				},
+			});
+
+			if (ks.firstVersion === '1') {
```
benr-ml, line 92: do we need this if?

packages/seal/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@mysten/seal",
-	"version": "0.4.5",
+	"version": "0.4.6",
```
hayes-mysten, line 3: we shouldn't manually bump versions

joyqvq: yes, this is temporary so tests can run

packages/seal/src/key-server.ts
```diff
@@ -62,29 +58,40 @@ export async function retrieveKeyServers({
...
+			// First get the KeyServer object and validate it.
+			const res = await client.core.getObject({
+				objectId,
+			});
 			const ks = KeyServerMove.parse(res.object.content);
-			if (ks.keyType !== 0) {
-				throw new UnsupportedFeatureError(`Unsupported key type ${ks.keyType}`);
+
+			if (
+				Number(ks.lastVersion) > EXPECTED_SERVER_VERSION ||
```
benr-ml, line 70: hmm, this condition is wrong, no?
should be if (last < expected || first > expected)

benr-ml: let's add a test to catch those

joyqvq: we agreed in the past that first_version = latest version. e.g. 2 (last), 3, 4, 5 (first) so if expected > 5 or expected < 2 throws error. 

we can still flip it though if you would like to think first_version = oldest

benr-ml: oh, i didn't remember that, it's a bit confusing that last<=first, no? do we depend on that elsewhere?

joyqvq: yes i agreed to that! we can flip it here if we want to. it's not used anywhere since we only initialized with 1 for both values.

packages/seal/test/unit/key-server.test.ts
```diff
@@ -18,59 +18,24 @@ import { Version } from '../../src/utils.js';
 const pk = fromBase64(
 	'oEC1VIuwQo+6FZiVwHCAy/3HbvAbuIyiztXIWwd4LgmXCh9WhOKg3T0+Mb62y9fqAsSaN5SybG09n/3JnkmEzJgdDXLpM8KvMwkha/cBHp6Cx7aCdogvGLoOp/RadyHb',
 );
-const id = '0xb35a7228d8cf224ad1e828c0217c95a5153bafc2906d6f9c178197dce26fbcf8';
+const id = '0x73d05d62c18d9374e3ea529e8e0ed6161da1a141a94d3f76ae3fe4e99356db75';
 const keyType = 0;
-const url = 'https://seal-key-server-testnet-1.mystenlabs.com';
-const name = 'mysten-testnet-1';
+const url = 'https://seal-key-server-testnet-v1-1.mystenlabs.com';
```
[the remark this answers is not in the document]
joyqvq: no we don't

packages/seal/src/key-server.ts
benr-ml: let's switch

hayes-mysten (approved):

PR #327, opened 2025-05-28
feat(seal): add optional api key to server request

joyqvq:
## Description

this also include the issue with scripts that commits the seal package version diff. 

## Test plan

---


packages/seal/src/client.ts
```diff
@@ -45,10 +45,12 @@ export interface SealClientExtensionOptions {
 
 export interface SealClientOptions extends SealClientExtensionOptions {
 	suiClient: SealCompatibleClient;
+	apiKey?: string;
 }
 
 export class SealClient {
 	#suiClient: SealCompatibleClient;
+	#apiKey?: string;
```
benr-ml, line 53: it should be per key server, and let the user define to header name.
perhaps let's replace serverObjectIds with something like KeyServerConfig[] where
class KeyServerConfig { objectId, weight, ?apiKeyHeader, ?apiKey }

packages/seal/src/client.ts
```diff
@@ -38,18 +38,25 @@ import { createFullId, count } from './utils.js';
  * @property timeout: Timeout in milliseconds for network requests. Defaults to 10 seconds.
  */
 export interface SealClientExtensionOptions {
-	serverObjectIds: [string, number][];
+	serverConfigs: KeyServerConfig[];
```
benr-ml, line 41: pls update the doc above

packages/seal/src/client.ts
```diff
@@ -38,18 +38,25 @@ import { createFullId, count } from './utils.js';
...
+}
+
 export interface SealClientOptions extends SealClientExtensionOptions {
 	suiClient: SealCompatibleClient;
 }
 
 export class SealClient {
 	#suiClient: SealCompatibleClient;
-	#weights: Map<string, number>;
+	#configs: Map<string, KeyServerConfig>;
```
benr-ml, line 59: let's check below also that either apiKeyName&&apiKey or !apiKeyName&&!apiKey

packages/seal/src/client.ts
```diff
@@ -30,26 +30,32 @@ import { createFullId, count } from './utils.js';
...
  * @property timeout: Timeout in milliseconds for network requests. Defaults to 10 seconds.
  */
 export interface SealClientExtensionOptions {
-	serverObjectIds: [string, number][];
+	serverConfigs: KeyServerConfig[];
 	verifyKeyServers?: boolean;
 	timeout?: number;
 }
 
+export interface KeyServerConfig {
```
benr-ml, line 45: need to expose in index.ts?

benr-ml (approved):
Need to expose

package.json
```diff
@@ -8,7 +8,7 @@
 		"icons": "pnpm --filter ./apps/icons",
 		"wallet": "turbo --filter ./apps/wallet",
 		"bcs": "turbo --filter ./packages/bcs",
-		"changeset-version": "pnpm changeset version && pnpm --filter @mysten/sui codegen:version",
+		"changeset-version": "pnpm changeset version && pnpm codegen:version",
```
hayes-mysten, line 11: I think you need to add a `-r` to run this across all workspace packages, or `--filter @mysten/*`

PR #328, opened 2025-05-28
[dapp-kit-next] build UI for the connect button and modal

williamrobertson13:
## Description

This PR implements the UI for the connect button and modal using shadcn theming variables. There are two parts of this that I'm just going to get put in follow-up PRs (the "you have no wallets installed" UI + the connected account dropdown which is a PITA). This could still use a good coat of polish, but it's *good enough* for now IMO. 

One other PS is that I'll do the documentation and auto-code doc analyzer stuff for the components in a separate PR.

| Theme         | Image |
|---------------|-------|
| Default theme |  |
| Default modal |  |
| Loading |  |
| Error |  |
| Custom theme  |  |

## Test plan
---
- Tested a variety of wallets, cancelling/going back while connecting, error retries
- Tested focus trap behavior, auto-focus behavior, keyboard navigation, mobile, etc.


packages/dapp-kit-next/src/core/store.ts
```diff
@@ -60,13 +60,17 @@ export function createStores<TNetworks extends Networks>({
...
-				? publicKeyFromSuiBytes(new Uint8Array(currentAccount.publicKey), {
-						address: currentAccount.address,
-					})
-				: null,
-		),
+		$publicKey: computed($baseConnection, ({ currentAccount }) => {
+			if (!currentAccount) return null;
+
+			try {
+				return publicKeyFromSuiBytes(new Uint8Array(currentAccount.publicKey), {
```
williamrobertson13, line 67: this was erroring (er breaking everything) because some wallets don't return public keys, I'll probably pull this out of an atom tbh but will do that later

packages/dapp-kit-next/src/core/initializers/autoconnect-wallet.ts
```diff
@@ -20,27 +20,29 @@ export function autoConnectWallet({
 	storageKey: string;
 }) {
 	onMount($compatibleWallets, () => {
-		return $compatibleWallets.listen(async (wallets, oldWallets) => {
-			if (oldWallets.length > wallets.length) return;
+		return $compatibleWallets.listen(
+			async (wallets, oldWallets: readonly UiWallet[] | undefined) => {
```
williamrobertson13, line 24: since this is computed now, `oldWallets` is undefined initially so just adding a check to guard here

packages/dapp-kit-next/src/core/store.ts
hayes-mysten: hmmm, thats not good.  If you know what wallets cause that, we should try to get them to fix this

williamrobertson13: I know Clay just landed a fix for zkLogin accounts in Slush, I think Nightly implements several things against the standard incorrectly though as well. I'll double check the offenders here

PR #332, opened 2025-05-29
fix(seal): check package ID version = 1

joyqvq:
## Description

check if the package is first version. 

## Test plan

---

### DIFF
--- a/.changeset/tame-pigs-watch.md
+++ b/.changeset/tame-pigs-watch.md
@@ -1,0 +1,5 @@
+---
+'@mysten/seal': patch
+---
+
+fix(seal): throw error early if package is not first version
--- a/packages/seal/src/session-key.ts
+++ b/packages/seal/src/session-key.ts
@@ -14,7 +14,7 @@ import {
 	InvalidPersonalMessageSignatureError,
 	UserError,
 } from './error.js';
-import type { SuiClient } from '@mysten/sui/client';
+import type { SealCompatibleClient } from './types.js';
 
 export const RequestFormat = bcs.struct('RequestFormat', {
 	ptb: bcs.vector(bcs.U8),
@@ -50,7 +50,7 @@ export class SessionKey {
 	#sessionKey: Ed25519Keypair;
 	#personalMessageSignature?: string;
 	#signer?: Signer;
-	#suiClient: SuiClient;
+	#suiClient: SealCompatibleClient;
 
 	constructor({
 		address,
@@ -65,7 +65,7 @@ export class SessionKey {
 		mvrName?: string;
 		ttlMin: number;
 		signer?: Signer;
-		suiClient: SuiClient;
+		suiClient: SealCompatibleClient;
 	}) {
 		if (mvrName && !isValidNamedPackage(mvrName)) {
 			throw new UserError(`Invalid package name ${mvrName}`);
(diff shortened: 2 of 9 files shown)


packages/seal/src/session-key.ts
```diff
@@ -177,12 +203,12 @@ export class SessionKey {
 	 * Restore a SessionKey instance for the given object.
 	 * @returns A new SessionKey instance with restored state
 	 */
-	static import(
+	static async import(
```
benr-ml, line 206: this one does not need to check the package again

benr-ml: both checks are sanity in case of user error, and worse case it will be detected when calling the server.

packages/seal/src/session-key.ts
```diff
@@ -177,11 +203,11 @@ export class SessionKey {
...
```
benr-ml, line 206: does it need to be async?

joyqvq: revised!

packages/seal/src/session-key.ts
```diff
@@ -46,9 +47,9 @@ export class SessionKey {
 	#sessionKey: Ed25519Keypair;
 	#personalMessageSignature?: string;
 	#signer?: Signer;
-	#suiClient: ZkLoginCompatibleClient;
+	#suiClient: SuiClient;
 
-	constructor({
+	private constructor({
```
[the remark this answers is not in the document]
joyqvq: sounds good, leaving it public for now.

packages/seal/test/unit/session-key.test.ts
```diff
@@ -4,15 +4,15 @@
...
+import { getFullnodeUrl, SuiClient } from '@mysten/sui/client';
 
 describe('Session key tests', () => {
 	const TESTNET_PACKAGE_ID = '0x9709d4ee371488c2bc09f508e98e881bd1d5335e0805d7e6a99edd54a7027954';
 	it('import and export session key', async () => {
 		const kp = Ed25519Keypair.generate();
-		const suiClient = new SuiGraphQLClient({ url: 'https://sui-testnet.mystenlabs.com/graphql' });
-		const sessionKey = new SessionKey({
+		const suiClient = new SuiClient({ url: getFullnodeUrl('testnet') });
+		const sessionKey = await SessionKey.create({
```
benr-ml, line 15: can remove

packages/seal/test/unit/session-key.test.ts
```diff
@@ -22,7 +22,7 @@ describe('Session key tests', () => {
 		await sessionKey.setPersonalMessageSignature(sig.signature);
 
 		const exportedSessionKey = sessionKey.export();
-		const restoredSessionKey = SessionKey.import(exportedSessionKey, suiClient);
+		const restoredSessionKey = await SessionKey.import(exportedSessionKey, suiClient);
```
benr-ml, line 25: can remove

benr-ml (approved):

.changeset/tame-pigs-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mysten/seal': patch
+---
+
+fix(seal): throw error early if package is not first version
```
hayes-mysten, line 5: you'll need a changeset for the graphql client change to ensure it also gets released

hayes-mysten (approved):

packages/seal/src/session-key.ts
```diff
@@ -49,7 +50,7 @@ export class SessionKey {
 	#sessionKey: Ed25519Keypair;
 	#personalMessageSignature?: string;
 	#signer?: Signer;
-	#suiClient: ZkLoginCompatibleClient;
+	#suiClient: SealCompatibleClient;
```
benr-ml, line 54: perhaps let's add something like 
/**
 * @deprecated Use `create()` instead.
 */

benr-ml (approved):

PR #335, opened 2025-06-02
[seal] Allow MVR names in signed message

jonas-lj:
## Description

Allow using MVR names for the signed message.

## Test plan

Unit tests (tbd)

---

### DIFF
--- a/.changeset/bumpy-coats-rhyme.md
+++ b/.changeset/bumpy-coats-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@mysten/seal': patch
+---
+
+Allow MVR names in signed message
--- a/.changeset/cuddly-badgers-leave.md
+++ b/.changeset/cuddly-badgers-leave.md
@@ -1,0 +1,5 @@
+---
+'@mysten/seal': minor
+---
+
+Allow MVR names in signed message
--- a/.changeset/cuddly-badgers-leave.md
+++ b/.changeset/cuddly-badgers-leave.md
@@ -1,5 +1,0 @@
----
-'@mysten/seal': minor
----
-
-Allow MVR names in signed message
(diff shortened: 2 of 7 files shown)


benr-ml (commented):
The main purpose of the MVR name in Seal is for the user signature, but developers would still need to use `packageId` for encrypt, decrypt, etc. If so, I think that SessionKey should receive optional `mvrName` (in addition to the non optional `packageId`), and simply use it in `getPersonalMessage` when given.

jonas-lj:
> > > The main purpose of the MVR name in Seal is for the user signature, but developers would still need to use `packageId` for encrypt, decrypt, etc. If so, I think that SessionKey should receive optional `mvrName` (in addition to the non optional `packageId`), and simply use it in `getPersonalMessage` when given.
> > 
> > 
> > Perhaps I misunderstand what you're saying, but isn't this exactly what's implemented? The `Package` holds an optional mvr_name along with the package_id. Or is it adding this `Package` struct that you oppose to?
> 
> Oh, I mistakingly thought it's either... but yes, why do we need this extra layer of Package?



> > > The main purpose of the MVR name in Seal is for the user signature, but developers would still need to use `packageId` for encrypt, decrypt, etc. If so, I think that SessionKey should receive optional `mvrName` (in addition to the non optional `packageId`), and simply use it in `getPersonalMessage` when given.
> > 
> > 
> > Perhaps I misunderstand what you're saying, but isn't this exactly what's implemented? The `Package` holds an optional mvr_name along with the package_id. Or is it adding this `Package` struct that you oppose to?
> 
> Oh, I mistakingly thought it's either... but yes, why do we need this extra layer of Package?

Well, I won't claim it's needed per se. The thought was to put the the address and MVR name in the same struct because they're so tightly connected, but if it's a source of confusion just adding an optional mvr name will be just as fine.

benr-ml (approved):

PR #338, opened 2025-06-02
[seal] Add nonce check

jonas-lj:
## Description

Add nonce check needed for CCA security.

## Test plan

Unit tests.

---


benr-ml (approved):
let's add tests that catch this and similar bugs in another pr

PR #339, opened 2025-06-02
Add MVR to core client

hayes-mysten:
## Description

## Test plan

---


packages/typescript/src/experimental/suins.ts
```diff
@@ -0,0 +1,179 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { chunk } from '@mysten/utils';
```
hayes-mysten, line 4: This is all just move from the current mvr plugin

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -70,116 +64,25 @@ export const namedPackagesPlugin = ({
...
-				[...getFirstLevelNamedTypes(names.types)].filter((x) => !cache.types[x]),
-				url,
-				pageSize,
-			),
-		]);
-
-		// save first-level mappings to cache.
-		Object.assign(cache.packages, packages);
-		Object.assign(cache.types, types);
+		const client = getClient(buildOptions);
```
hayes-mysten, line 74: this should be updated so we only get the client if there are types or packages that need to be resolved

packages/typescript/src/client/client.ts
```diff
@@ -4,13 +4,14 @@ import { fromBase58, toBase64, toHex } from '@mysten/bcs';
 
 import type { Signer } from '../cryptography/index.js';
 import { Experimental_BaseClient } from '../experimental/client.js';
+// eslint-disable-next-line import/no-cycle
```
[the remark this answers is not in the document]
hayes-mysten: I think the way to avoid the cycle is to not have the plugin use the SuiClient, and just call another method to resolve names instead.  I'll look into it

manolisliolios (approved):
LGTM!

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
```diff
@@ -40,9 +35,14 @@ export type NamedPackagesPluginOptions = {
...
 	 *
 	 */
-	overrides?: NamedPackagesPluginCache;
+	overrides?: NamedPackagesOverrides;
 };
 
+// The original versions of the mvr plugin cached lookups by mutating overrides.
+// We don't want to mutate the options, but we can link our cache to the provided overrides object
+// This preserves the caching across transactions while removing the mutation side effects
+const cacheMap = new WeakMap<object, ClientCache>();
```
williamrobertson13, line 44: I mean you could have nuked the plugin in terms of backward compatibility since it's experimental, but this seems like a win-win haha

packages/typescript/src/experimental/core.ts
```diff
@@ -153,4 +171,19 @@ export abstract class Experimental_CoreClient
 			}
 		}
 	}
+
+	resolveNamedPackage({ name }: { name: string }): Promise<string> {
```
williamrobertson13, line 175: `resolveRegistryPackage`, `resolveMoveRegistryPackage`, `resolveNamedPackage`, eh

williamrobertson13 (approved):
lgtm from a glance, sorry that I don't have more valuable feedback for what you asked about haha 😅

packages/typescript/src/experimental/mvr.ts
```diff
@@ -0,0 +1,446 @@
...
+	#url?: string;
+	#pageSize: number;
+	#overrides: {
+		packages?: Record<string, string>;
+		types?: Record<string, string>;
+	};
+
+	constructor({ cache, url, pageSize = 50, overrides }: MvrClientOptions) {
+		this.#cache = cache;
+		this.#url = url;
```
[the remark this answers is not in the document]
hayes-mysten: yup.  We will probably also just have a default based on the network soon

packages/typescript/src/experimental/mvr.ts
```diff
@@ -0,0 +1,446 @@
...
+	normalizeSuiAddress,
+	parseStructTag,
+} from '../utils/sui-types.js';
+import type { ClientCache } from './cache.js';
+import type { TransactionDataBuilder } from '../transactions/TransactionData.js';
+import { PACKAGE_VERSION } from '../version.js';
+
+const NAME_SEPARATOR = '/';
+const MVR_API_HEADER = {
+	'Mvr-Source': `@mysten/sui@${PACKAGE_VERSION}`,
```
Jordan-Mysten, line 19: Just confirming this returns the correct CORS headers to not make this a problem

packages/typescript/src/experimental/mvr.ts
```diff
@@ -0,0 +1,446 @@
...
+ */
+export function findNamesInTransaction(builder: TransactionDataBuilder): {
+	packages: string[];
+	types: string[];
+} {
+	const packages: Set<string> = new Set();
+	const types: Set<string> = new Set();
+
+	for (const command of builder.commands) {
+		if (command.MakeMoveVec?.type) {
```
Jordan-Mysten, line 362: I personally like switches over the kind but I get why you don't do it given almost all of the cases are the default case.

packages/typescript/src/experimental/mvr.ts
```diff
@@ -0,0 +1,446 @@
...
+		this.#overrides = {
+			packages: overrides?.packages,
+			types: overrides?.types,
+		};
+
+		validateOverrides(this.#overrides);
+	}
+
+	#mvrPackageDataLoader(url = this.#url) {
+		return this.#cache.readSync(['#mvrPackageDataLoader', url ?? ''], () => {
```
Jordan-Mysten, line 54: Oh this is an interesting way to create cached loaders...

Jordan-Mysten: Why not just chuck this onto the instance? Not saying this is bad just curious

packages/typescript/src/experimental/mvr.ts
```diff
@@ -0,0 +1,446 @@
...
+
+		return results;
+	}
+
+	async #fetch<T>(url: string, body: Record<string, unknown>): Promise<T> {
+		if (!this.#url) {
+			throw new Error('MVR Api URL is not set for the current client');
+		}
+
+		const response = await fetch(`${this.#url}${url}`, {
```
Jordan-Mysten, line 164: Any reason we'd ever need to customize the fetch function? Mostly just thinking about patterns from things like the walrus SDK

packages/typescript/src/transactions/plugins/NamedPackagesPlugin.ts
Jordan-Mysten: Yeah I'm also fine nuking it but it's neat you could make it compatible

packages/typescript/src/experimental/mvr.ts
hayes-mysten: I want to be able to always do something like `client.reset()` and have it clear all caches

packages/utils/src/dataloader.ts
```diff
@@ -0,0 +1,518 @@
+// Copyright (c) Mysten Labs, Inc.
```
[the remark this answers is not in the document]
hayes-mysten: just removing unaudited 3rd party deps.  dataloader is probably pretty safe, but just having as few deps as possible makes sense.  I'd much rather just have a file in our repo over a single file dependency that can be updated by 3rd parties

packages/typescript/src/experimental/mvr.ts
hayes-mysten: I should test this.  I was told it should, but I didn't actually confirm myself

packages/typescript/src/experimental/mvr.ts
hayes-mysten: I thought about this, but couldn't think of a good reason today.  It's not hard to add it in the future, but harder to remove, so I omitted it for now

packages/typescript/src/experimental/mvr.ts
hayes-mysten: yeah same.  This was just moved from the old implementation, but will update

packages/typescript/src/experimental/mvr.ts
hayes-mysten: This was also creating a different loader for each url for a previous iteration of how this was used in the tx plugin.  I've updated it so thats not needed anymore, and is now just a getter

PR #340, opened 2025-06-03
[experimental client] add method to resolve name service names to the core client

williamrobertson13:
## Description

In the web components for dApp Kit, we need to be able to resolve name service names for the connected account menu across our standardized client transports. This PR adds a method for resolving SuiNS names + a follow-up PR will change the return type of `createClient` in dApp Kit to return `ClientWithCoreApi` instances.

## Test plan
- Will test that this works in dApp Kit before landing

### DIFF
--- a/packages/typescript/src/experimental/core.ts
+++ b/packages/typescript/src/experimental/core.ts
@@ -71,6 +71,10 @@ export abstract class Experimental_CoreClient
 		options: Experimental_SuiClientTypes.GetDynamicFieldsOptions,
 	): Promise<Experimental_SuiClientTypes.GetDynamicFieldsResponse>;
 
+	abstract resolveNameServiceNames?(
+		options: Experimental_SuiClientTypes.ResolveNameServiceNamesOptions,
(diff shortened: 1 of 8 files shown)


packages/typescript/src/experimental/core.ts
```diff
@@ -71,6 +71,10 @@ export abstract class Experimental_CoreClient
 		options: Experimental_SuiClientTypes.GetDynamicFieldsOptions,
 	): Promise<Experimental_SuiClientTypes.GetDynamicFieldsResponse>;
 
+	abstract resolveNameServiceNames?(
```
hayes-mysten, line 74: I think adding this as abstract means every implementation needs to define it.  Can this be defined as `resolveNameServiceNames: undefined` or something like that? I just want to make sure it doesn't cause issues

hayes-mysten: I think I ended up just omitting it from here for the zklogin signature verification method, but can't recall for sure

hayes-mysten (approved):

PR #361, opened 2025-06-09
[dapp-kit-next] add React bindings

williamrobertson13:
## Description

This PR adds React bindings for the new dApp Kit! I decided to ship a `Provider` component because it makes testing easier and in case folks ever need to create the kit instance inside render code (basically following suit with pretty much every popular React library, https://tkdodo.eu/blog/zustand-and-react-context for some more reading). 

The hooks and components are setup so you can get full type safety by either using the provider with declaration merging or explicitly passing the dApp Kit instance into the hook, which is nice.

(there's probably some more stuff to add here that'd be helpful, but this seems equivalent to what the current library ships today)

## Test plan
- Manual testing

image


packages/dapp-kit-next/packages/dapp-kit-react/src/hooks/useSuiClient.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import type { DAppKit, RegisteredDAppKit } from '@mysten/dapp-kit-core';
+import { useStore } from '@nanostores/react';
+import { useDAppKit } from './useDAppKit.js';
+
+export type UseSuiClientOptions<TDAppKit extends DAppKit> = {
+	dAppKit?: TDAppKit;
+};
+
+export function useSuiClient<TDAppKit extends DAppKit = RegisteredDAppKit>({
```
hayes-mysten, line 12: dumb semantic question (cc @Jordan-Mysten) do we want to use `suiClient` as a term for our new extendable client, or should we have just `useClient` (which will probably make imports and llms confused).

Pretty sure suiClient is fine, but wanted to double check

hayes-mysten (approved):
I love how simple this is

packages/dapp-kit-next/packages/dapp-kit-react/src/hooks/useSuiClient.ts
williamrobertson13: I was 50/50 on this too, eh gonna land for now to get the last bit of feature work in and will re-work this with whatever we decide on naming + re-structuring with the hook factory pattern

PR #363, opened 2025-06-10
[seal] Get rid of dependency for Shamir's secret sharing

jonas-lj:
## Description

Get rid of dependency for Shamir's secret sharing.

## Test plan

Unit tests.

---


packages/seal/src/shamir.ts
```diff
@@ -0,0 +1,372 @@
...
+		) {
+			this.coefficients.pop();
+		}
+	}
+
+	static fromBytes(bytes: Uint8Array): Polynomial {
+		return new Polynomial(Array.from(bytes, (b) => new GF256(b)));
+	}
+
+	degree(): number {
```
benr-ml, line 133: deg would be -1 for zero()

benr-ml (approved):
let's make sure we tested all operations

PR #382, opened 2025-06-17
[enoki] return info about the active session from EnokiGetMetadata

williamrobertson13:
## Description

Since we lean into the wallet standard now for handling Enoki wallets, it isn't possible to retrieve the underlying session information such as the JWT which can expose useful information if you provide extra auth scopes.

This PR updates the `EnokiGetMetadata` feature to expose the `decodedJwt` on the **active session**. In the near future, the Enoki wallet will support multiple sessions (one per supported network) so I'm trying to save a breaking change here (see https://github.com/MystenLabs/ts-sdks/compare/main...wrobertson/port_enoki_demo#diff-bb15b729a6aefd6a2be396041c3e799055e2cea9ec9774729819d37c76ea7facR83 for a WIP implementation).

Pattern with the current version of dApp Kit:

```
registerEnokiWallets({
  providers: {
    google: {
      extraParams: {
        scope: 'email profile',
      },
     clientId: 'abc',
},

...

function getSignedInEmail(enokiWallet: Wallet) {
  const metadata = getWalletMetadata(enokiWallet);
  const decodedJwt = metadata.activeSession?.decodedJwt;

  return decodedJwt && 'email' in decodedJwt ? decodedJwt.email : null;
}

......

const enokiWallet = useWallets().find(isEnokiWallet);
const email = enokiWallet ? getSignedInEmail(enokiWallet) : null;

console.log('email!', email);
```

## Test plan
Tested in the demo dApp:

image

### DIFF
--- a/packages/enoki/src/wallet/index.ts
+++ b/packages/enoki/src/wallet/index.ts
@@ -39,6 +39,7 @@ import { INTERNAL_ONLY_EnokiFlow } from './state.js';
 import type { RegisterEnokiWalletsOptions, WalletEventsMap } from './types.js';
 import type { EnokiGetMetadataFeature, EnokiGetMetadataMethod } from './feature.js';
 import { EnokiGetMetadata } from './feature.js';
+import { decodeJwt } from '@mysten/sui/zklogin';
 
 export class EnokiWallet implements Wallet {
 	#events: Emitter<WalletEventsMap>;
@@ -294,8 +295,12 @@ export class EnokiWallet implements Wallet {
 	};
 
 	#getMetadata: EnokiGetMetadataMethod = () => {
+		const session = this.#flow.$zkLoginSession.get();
+		const decodedJwt = session.value?.jwt ? decodeJwt(session.value.jwt) : undefined;
+
 		return {
 			provider: this.#provider,
+			activeSession: decodedJwt ? { decodedJwt } : undefined,
 		};
 	};
 
--- a/packages/enoki/src/wallet/index.ts
+++ b/packages/enoki/src/wallet/index.ts
@@ -227,12 +227,7 @@ export class EnokiWallet implements Wallet {
 					address: state.address,
 					chains: this.chains,
 					icon: this.icon,
-					features: [
-						EnokiGetMetadata,
-						SuiSignPersonalMessage,
-						SuiSignTransaction,
-						SuiSignAndExecuteTransaction,
-					],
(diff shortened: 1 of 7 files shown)


packages/enoki/src/wallet/utils.ts
```diff
@@ -19,21 +19,17 @@ export function isEnokiWallet(wallet: Wallet | UiWallet) {
...
+			enokiWallet,
+			EnokiGetMetadata,
+		) as EnokiGetMetadataFeature[typeof EnokiGetMetadata];
+
+		return getMetadata();
 	} else if (EnokiGetMetadata in enokiWallet.features) {
 		const walletWithFeature = enokiWallet as WalletWithFeatures<EnokiGetMetadataFeature>;
 		return walletWithFeature.features[EnokiGetMetadata].getMetadata();
 	}
-	return null;
```
williamrobertson13, line 36: this is technically a breaking change but I just added it yesterday and don't think anyone is using it. Not sure if returning null is better or throwing an error, I guess the expectation is really that you only call this on wallets you know are Enoki wallets though

packages/enoki/src/wallet/state.ts
```diff
@@ -118,9 +118,7 @@ export class INTERNAL_ONLY_EnokiFlow {
 			scope: [
 				'openid',
 				// Merge the requested scopes in with the required openid scopes:
-				...(input.extraParams && 'scope' in input.extraParams
-					? (input.extraParams.scope as string[])
-					: []),
+				...(input.extraParams?.scope ? input.extraParams.scope.split(' ') : []),
```
williamrobertson13, line 121: this was broken because `registerEnokiWallets` was expecting a record of strings for the extra parameters so you couldn't do `scope: [...]`. I did the easy thing and just made it so you specify scope like `scope: 'email profile ...'`

packages/enoki/src/wallet/index.ts
```diff
@@ -294,8 +295,12 @@ export class EnokiWallet implements Wallet {
 	};
 
 	#getMetadata: EnokiGetMetadataMethod = () => {
+		const session = this.#flow.$zkLoginSession.get();
+		const decodedJwt = session.value?.jwt ? decodeJwt(session.value.jwt) : undefined;
```
williamrobertson13, line 299: I could add a new feature for each specific thing, but meeeeh idk? Decoding the JWT takes < .5ms

Jordan-Mysten: We could also provide the raw JWT? But this is fine to me

Jordan-Mysten (approved):

PR #383, opened 2025-06-17
scaffold suins-v2 package

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/suins-v2/package.json
+++ b/packages/suins-v2/package.json
@@ -43,8 +43,8 @@
 		"node": ">=16"
 	},
 	"dependencies": {
-		"@mysten/sui": "workspace:*",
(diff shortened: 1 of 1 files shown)


packages/codegen/src/generate-utils.ts
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+export const utilsContent = /* ts */ `
```
hayes-mysten, line 4: this will probably eventually just go into mysten/sui, but going to just render it out for now

Jordan-Mysten (approved):

PR #393, opened 2025-06-20
Wrobertson/enoki multi network

williamrobertson13:
## Description

This PR modifies the Enoki wallet standard interface so that is supports multiple networks and supports auto-connection since we'll now persist the address/public key in local storage as opposed to session storage. There are a bunch of changes required to make this work and it isn't the cleanest, but I didn't want to spend forever trying to get this in any better shape.

Some notable changes here:
  - We now track the `zkLoginState` in local storage (only address + public key) as opposed to session storage
  - zkLogin session state is now segmented by `network`. So for one wallet, you could have N sessions created
  - Enoki wallets used to share state across providers which was incorrect and would end up in scenarios where you have a Google wallet account authorized for a Twitch or Facebook wallet. This is fixed now
  - We now have an `enokiWalletsInitializer` that can be plugged into dApp Kit v2
  - This change is backward compatible and should work with the old version of dApp Kit

## Test plan
- New test demo app and the old one
- Tested auto-connect functionality, switching networks, and transaction signing


packages/enoki/src/wallet/initializer.ts
```diff
@@ -0,0 +1,29 @@
...
+		async initialize({
+			networks,
+			getClient,
+		}: {
+			networks: readonly Experimental_SuiClientTypes.Network[];
+			getClient: (network?: Experimental_SuiClientTypes.Network) => ClientWithCoreApi;
+		}) {
+			const { unregister } = registerEnokiWallets({
+				...options,
+				getCurrentNetwork: () => getClient().network,
```
williamrobertson13, line 22: passing in a function to get the current network is kind of weird but required because when you connect you need to know which network to try to make the connection against

I could probably skirt this by ordering the clients so the first one is the active network?

PR #395, opened 2025-06-23
Add calls to suins v2 client

hayes-mysten:
## Description

## Test plan

---


packages/codegen/src/index.ts
```diff
@@ -20,6 +20,14 @@ export async function generateFromPackageSummary({
 }) {
 	const summaryDir = join(source, 'package_summaries');
 
+	let packageName = name;
+	try {
+		const packageToml = await readFile(join(source, 'Move.toml'), 'utf-8');
+		packageName = packageToml.match(/name\s*=\s*"(.*)"/)?.[1] ?? name;
```
hayes-mysten, line 26: I think this is fine for now.  We can pull in a toml parser later if needed

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
...
+	#couponHouseContract: ReturnType<typeof initCouponHouseContract>;
+
+	constructor(options: SuiNsCallsOptions) {
+		this.#paymentContract = initPaymentContract(options.packageIds.suins);
+		this.#paymentsContract = initPaymentsContract(options.packageIds.payments);
+		this.#couponHouseContract = initCouponHouseContract(options.packageIds.coupons);
+	}
+
+	get register() {
+		throw new Error('Not implemented');
```
hayes-mysten, line 32: These unimplemented These are methods that exist in the current suins client, but do more than just wrap the move call

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { init as initPaymentsContract } from './contracts/payments/payments.js';
+import { init as initPaymentContract } from './contracts/suins/payment.js';
+import { init as initCouponHouseContract } from './contracts/coupons/coupon_house.js';
```
hayes-mysten, line 6: I don't love having to alias these, but I don't have a great alternative... we could do something like `intit${capitalize(moduleName)}` instead? You might still need aliases if you have the same module name in multiple packages

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
...
+
+// TODO: Default these to MVR names and make them optional
+export interface SuiNsPackageIds {
+	suins: string;
```
hayes-mysten, line 10: When defining ids for packages, do we want to combine packages and objects?

Eg:

```ts
{
  suiNs: {
    package: '@suins/core',
    systemState: '0x123...',
  }
}
```

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
...
+	coupons: string;
+	discounts: string;
+}
+
+export interface SuiNsCallsOptions {
+	packageIds: SuiNsPackageIds;
+}
+
+export class SuiNsCalls {
+	#paymentContract: ReturnType<typeof initPaymentContract>;
```
hayes-mysten, line 21: should I export a type for this for codegen? It would have the same naming issue mentioned above

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
...
+	get register() {
+		throw new Error('Not implemented');
+	}
+
+	get renew() {
+		throw new Error('Not implemented');
+	}
+
+	get initRegistration() {
+		return this.#paymentContract.init_registration;
```
hayes-mysten, line 40: How do we want SDKs to actually do this?

* Do we want to rename methods (camel case)
* Do we like having a flat had-picked set of calls on the client or do we want to nest things?
  * Fully nesting (eg `client.suins.calls.payments.paymentes.init_registration` with both package and module names feels like too much)

packages/codegen/src/index.ts
Jordan-Mysten: I think this is fine but also a toml parser seems easy to pull in especially given this is a CLI and not shipped as part of the client

packages/suins-v2/package.json
```diff
@@ -29,7 +29,7 @@
 	"scripts": {
 		"clean": "rm -rf tsconfig.tsbuildinfo ./dist",
 		"build": "build-package",
-		"codegen": "pnpm sui-ts-codegen generate -s ../../../suins-contracts/packages/suins -o src/contracts && pnpm lint:fix",
+		"codegen": "pnpm sui-ts-codegen generate -s ../../../suins-contracts/packages/suins -s ../../../suins-contracts/packages/payments -s ../../../suins-contracts/packages/coupons -s ../../../suins-contracts/packages/discounts -s ../../../suins-contracts/packages/temp_subdomain_proxy -o src/contracts && pnpm lint:fix",
```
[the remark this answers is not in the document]
Jordan-Mysten: Why no just recursively read the directory? Why do this at all? I think globs are easy too though so you can just do *.move or whatever

packages/suins-v2/src/client.ts
```diff
@@ -2,17 +2,22 @@
...
+	packageIds: SuiNsPackageIds;
+}
 
 export interface SuiNsClientOptions extends SuiNsClientExtensionOptions {
 	client: SuiNsCompatibleClient;
 }
 
 export class SuiNsClient {
 	#client: SuiNsCompatibleClient;
+	calls: SuiNsCalls;
```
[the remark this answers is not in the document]
Jordan-Mysten: I'm fine with `calls` yeah. Other options I don't completely hate:
- `fns`
- `commands`
- `ptbs`
- `call` (non-plural but confusing because JS)
- `methods`
- `public` (probably terrible)
- `api`

packages/suins-v2/src/client.ts
```diff
@@ -2,17 +2,22 @@
...
 
 import type { ClientWithCoreApi, SuiClientRegistration } from '@mysten/sui/src/experimental';
+import { SuiNsCalls } from './calls.js';
+import type { SuiNsPackageIds } from './calls.js';
 
 export interface SuiNsCompatibleClient extends ClientWithCoreApi {}
 
-export interface SuiNsClientExtensionOptions {}
+export interface SuiNsClientExtensionOptions {
+	packageIds: SuiNsPackageIds;
```
Jordan-Mysten, line 11: Required until deep MVR support or?

packages/suins-v2/src/calls.ts
Jordan-Mysten: I think changing the export to `initModuleName` is probably sane, or default export?

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { init as initPaymentsContract } from './contracts/payments/payments.js';
+import { init as initPaymentContract } from './contracts/suins/payment.js';
+import { init as initCouponHouseContract } from './contracts/coupons/coupon_house.js';
+
+// TODO: Default these to MVR names and make them optional
```
Jordan-Mysten, line 8: Okay cool this answers my question below

packages/suins-v2/src/calls.ts
Jordan-Mysten: What if you export a namespace like `ModuleName` and then you can do `ModuleName.init` and `ModuleName` can have all the types itself (or do something like `ModuleName.$type` if that doesn't work for whatever reason.

packages/suins-v2/src/calls.ts
Jordan-Mysten: I assume these are like PTB compositions? Would kinda be cool to codegen these too but out of scope for now

packages/suins-v2/src/calls.ts
Jordan-Mysten: - I'm going back and forth on this. I think camelCase is more common in JS but I also kinda like mapping it closer to the chain so that it's obvious what you're doing? No super strong preference tho.
- I think not nesting is good, the API should be curated (ideally eventually code-gen'd itself).

packages/suins-v2/src/calls.ts
```diff
@@ -0,0 +1,130 @@
...
+	get finalizeRenew() {
+		return this.#paymentContract.renew;
+	}
+
+	get calculatePriceAfterDiscount() {
+		return this.#paymentsContract.calculate_price_after_discount;
+	}
+
+	get generateReceipt() {
+		throw new Error('Not implemented');
```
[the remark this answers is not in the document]
Jordan-Mysten: I think for well known system stuff (clock, randomness, system state), yes. We can basically treat them like the ctx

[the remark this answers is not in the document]
hayes-mysten: done

packages/suins-v2/src/calls.ts
hayes-mysten: probably a couple of them, but it's also branching to call the right method depending on an arg

packages/suins-v2/src/client.ts
hayes-mysten: yes, but I think this is fixed now!

PR #404, opened 2025-06-27
Redacted + improved privacy of private keys

Jordan-Mysten:
## Description

Taking inspiration from how Effect handles Redacted values, this introduces our own `redacted` function, which creates a value that is stored via a weakmap on a global registry. This has some neat properties:
- You need a reference to the redacted to get it out of the registry, so this should be plenty safe (weakmaps don't allow enumeration, or reading values without the keys).
- Logging the redacted identifier is completely safe, because the actual underlying value is stored completely distinctly
- You can kinda improve debugging of these values since we can add the debug tags that denote that this is a redacted value when it does get logged.

I originally wanted this to be exposed in a create-only way, so that outside of the package you couldn't read reacted values, but I realized that we probably want methods like `getSecretKey` to return redacted values by default.

While I was here, I realized we were using fake private for the data in keypairs, which kind of scares the shit out of me because if you log a keypair (which IMO seems like an operation that should be plenty safe) it means that all of the secret values are exposed:
Screenshot 2025-06-27 at 12 35 11 PM

I made this a true private which _shouldn't_ be a breaking change unless people are depending on undocumented properties. I thought about changing this to store as a redacted internally, but just using true privates felt like a better solution here.

## Test plan

I will test it + update docs before merging.

---

### DIFF
--- a/packages/typescript/src/cryptography/keypair.ts
+++ b/packages/typescript/src/cryptography/keypair.ts
@@ -13,6 +13,7 @@ import type { SignatureScheme } from './signature-scheme.js';
 import { toSerializedSignature } from './signature.js';
 import type { Transaction } from '../transactions/Transaction.js';
 import type { ClientWithCoreApi, Experimental_SuiClientTypes } from '../experimental/index.js';
+import { Redacted } from '../utils/redacted.js';
 
 export const PRIVATE_KEY_SIZE = 32;
 export const LEGACY_PRIVATE_KEY_SIZE = 64;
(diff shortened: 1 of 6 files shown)


packages/typescript/src/cryptography/keypair.ts
```diff
@@ -112,8 +113,15 @@ export abstract class Signer {
...
 	 * This returns the Bech32 secret key string for this keypair.
+	 *
+	 * We recommend using `getSecretKeyRedacted()` instead, to ensure the secret key is not logged.
 	 */
 	abstract getSecretKey(): string;
+
+	/**
+	 * Get the secret key for this keypair, with the value redacted so that it cannot be logged.
+	 */
+	abstract getSecretKeyRedacted(): Redacted<string>;
```
hayes-mysten, line 124: this can just be a concrete implementation right?

PR #408, opened 2025-07-01
update move-bytecode-template wasm build and instructions

hayes-mysten:
## Description

## Test plan

---


packages/move-bytecode-template/README.md
```diff
@@ -106,74 +106,70 @@ guide to deploy the contract.
...
-
-console.assert(json == bytes, '(de)serialization failed!');
-```
+- `init` is an exported function that initializes the wasm module in browser environments (a no-op
+  function is exported when running in node so you can use this anywhere)
+- `/web/walrus_wasm_bg.wasm` is an export of the wasm module itself, which may be needed in some
+  bundlers to know where to import the wasm module from
 
-## Using with Vite
+### Initializing in webpack and nodejs
```
williamrobertson13, line 120: not really important at all but why'd we change to webpack? haha

williamrobertson13 (approved):

packages/move-bytecode-template/README.md
hayes-mysten: There is a vite section below, its just webpack in theory has less setup.  so docs are in order of how many things you need to do to get it to work

williamrobertson13: fml this is the second time I've mis-read something simple in PRs today 😆

PR #411, opened 2025-07-01
add docs and small fixes for codegen

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/codegen/README.md
+++ b/packages/codegen/README.md
@@ -28,7 +28,7 @@ const config: SuiCodegenConfig = {
 export default config;
 ```
 
-The `package` field should be the MVR name for your move package. If you have not registered you
+The `package` field should be the MVR name for your move package. If you have not registered your
 package on MVR yet, you can use the `@local-pkg` scope, and set up an override in your `SuiClient`
 to resolve it to the correct address.
 


williamrobertson13 (approved):
lgtm!

pyncz:
Hey! I was about to start writing codegen for my project, but then discovered you guys are developing it rn - seems very handy, thank you thank you!

I have a question tho - is there a way to generate methods for std/sui packages, to abstract all the coin/balance/uq etc calls? I guess not, until onchain packages (without `path` specified) in config are supported, right?

packages/codegen/README.md
```diff
@@ -1 +1,193 @@
...
+	});
+
+	return digest;
+}
+
+async function incrementCount(id: string) {
+	const tx = tx.add(
+		counter.increment({
+			// Arguments can be passed in by name as an object, or as an array of values
+			arguments: {
```
Jordan-Mysten, line 153: I'm curious why we didn't just make this function take the arguments as the first param

Jordan-Mysten (approved):

packages/codegen/README.md
hayes-mysten: consistency with functions that take type_parameters, although that might still be the better option.  We could do something like

`module.func(arguments, { typeArguments, package }) or something` and expect most people will only ever pass arguments

Jordan-Mysten: I'm fine with either, I think your justification makes sense

PR #412, opened 2025-07-03
update schemas and support ConsensusAddressOwner

hayes-mysten:
## Description

## Test plan

---


packages/sui-grpc/src/core.ts
```diff
@@ -167,6 +176,9 @@ export class GrpcCoreClient extends Experimental_CoreClient {
 				bcs: {
 					value: fromBase64(signature),
 				},
+				signature: {
```
hayes-mysten, line 179: @bmwill Is this the expected way to send a signature when executing a transaction?  The generated types expect signature to be set, but I think we only want to send the bcs format?

PR #420, opened 2025-07-07
[enoki-sdk] wallet support custom auth providers

pchrysochoidis:
## Description

## Test plan

---


packages/enoki/src/wallet/register.ts
```diff
@@ -28,13 +27,28 @@ export function registerEnokiWallets({
 		'clients' in config ? config.getCurrentNetwork : () => clients[0].network;
 
 	const walletsApi = getWallets();
-	const wallets: Partial<Record<AuthProvider, EnokiWallet>> = {};
+	const wallets: Record<string, EnokiWallet> = {};
 
-	for (const { name, icon, provider } of ENOKI_PROVIDER_WALLETS_INFO) {
+	for (const { name, icon, provider, authenticationUrl, extraParams } of [
+		...(config.additionalProvidersWalletInfo ?? []),
+		...ENOKI_PROVIDER_WALLETS_INFO,
```
williamrobertson13, line 34: Hmmm so root issue is that there's drift between the Enoki API types and the types in our `ts-sdks` repo as we don't auto-generate the API types and there can be skew either way.

Oh man I don't have a great solution off the top of my head, but I'd rather almost just manually update the auth provider type + hardcode the set of provider metadata in `ENOKI_PROVIDER_WALLETS_INFO` than provide this stopgap option to patch around the issue. This has to be stored or maintained _somewhere_ (API/some shared package/here), the question is where 😆 

cc @hayes-mysten if you have any thoughts

pchrysochoidis: Yeah I added one championship [here](https://github.com/MystenLabs/ts-sdks/pull/427), so I will close this. I was thinking that with this `additionalProvidersWalletInfo` people could easily add any providers that are not in the wallet yet. But adding it in the wallet is pretty quick and less complex.

pchrysochoidis:
closing in favour of https://github.com/MystenLabs/ts-sdks/pull/427

PR #422, opened 2025-07-08
fix(codegen): move to next arg when resolved sui std arg in `normalizeMoveArguments`

pyncz:
## Description

Small fix for the codegen's `normalizeMoveArguments` util to fix args' resolution, when a Sui std arg object is present (i.e. `Clock`, `Random`, `DenyList` or `SuiSystemState`).

Previously, the function would not move on to the next argType after pushing the correct value into the `normalizedArgs` array for that Sui type, but would continue the iteration, trying to process the arg as if it were an arbitrary arg, which would eventually result in errors like "Parameter clock is required".

## Test plan

- Added test cases
- ensured they were failing for the previous code, indicating the issue
- added the fix
- made sure the tests were passing after.

### DIFF
--- a/.changeset/real-wolves-hide.md
+++ b/.changeset/real-wolves-hide.md
@@ -1,0 +1,6 @@
+---
+'@mysten/codegen': patch
+'@mysten/walrus': patch
+---
+
+Update codegen args' normalization
--- a/packages/codegen/package.json
+++ b/packages/codegen/package.json
@@ -30,6 +30,7 @@
 		"build": "build-package",
 		"codegen": "pnpm tsx src/bin/cli.ts generate && pnpm lint:fix",
 		"prepublishOnly": "pnpm turbo build",
+		"test": "vitest run",
 		"test:watch": "vitest",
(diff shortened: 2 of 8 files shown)


hayes-mysten (approved):
Thanks for this fix! Sorry about the bug here, this functionality hasn't really been tested yet since the packages we've been using this on so far don't use any of these objects.

Just pushed up some test changes, and updated the package.json so the new tests actually run as part of CI

pyncz:
Yep, thanks! I get it's just 0.x.x and not everything is 100% checked yet, no problem.

About the `package.json` btw. I've noticed the linked repo there is `sui` and not `ts-sdks`. Idk if intentionally, I guess not, so wanted to let you know.

pyncz:
@hayes-mysten Great, love the type-safety part of the roadmap.
In my experience, so far so good! Will get a better grasp with time ofc, especially after that clock fix is released.

Although, there's one suggestion, I think. Described it in this issue:
https://github.com/MystenLabs/ts-sdks/issues/426

PR #423, opened 2025-07-08
[walrus] update linear tip config

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/walrus/src/fan-out-proxy/client.ts
+++ b/packages/walrus/src/fan-out-proxy/client.ts
@@ -8,7 +8,6 @@ import type {
 	EncodingType,
 	FanOutTipConfig,
 	ProtocolMessageCertificate,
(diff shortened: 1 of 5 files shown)


packages/walrus/src/fan-out-proxy/client.ts
```diff
@@ -76,11 +75,35 @@ export class FanOutProxyClient {
...
+				},
+			};
+		}
+
+		return {
+			address: data.send_tip.address,
+			kind: {
+				linear: {
+					base: data.send_tip.kind.base,
+					perKb: data.send_tip.kind.encoded_size_mul_per_kb,
```
hayes-mysten, line 103: @Jordan-Mysten thoughts on how closely we should match the config shape here? I kinda hate `encoded_size_mul_per_kb`

PR #429, opened 2025-07-09
wip - improve names of bcs types so they can be used as struct types

hayes-mysten:
## Description

## Test plan

---

### DIFF
--- a/packages/bcs/src/bcs.ts
+++ b/packages/bcs/src/bcs.ts
@@ -3,6 +3,9 @@
 
 import type { BcsTypeOptions } from './bcs-type.js';
 import {
+	BcsEnum,
+	BcsStruct,
+	BcsTuple,
 	BcsType,
 	bigUIntBcsType,
 	dynamicSizeBcsType,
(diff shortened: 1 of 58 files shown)


packages/bcs/src/bcs.ts
```diff
@@ -11,9 +14,178 @@ import {
...
+import type {
+	EnumInputShape,
+	EnumOutputShape,
+	InferBcsInput,
+	InferBcsType,
+	JoinString,
+} from './types.js';
 import { ulebEncode } from './uleb.js';
 
+function fixedArray<T extends BcsType<any>, Name extends string = string>(
```
hayes-mysten, line 26: I am using function overloads here for backwards compatibility.  The first definition is the new version that is preferred

packages/walrus/src/utils/index.ts
```diff
@@ -137,7 +137,7 @@ export function signersToBitmap(signers: number[], committeeSize: number): Uint8
 	return bitmap;
 }
 
-export function getShardIndicesByNodeId(committee: InferBcsType<ReturnType<typeof Committee>>) {
```
hayes-mysten, line 140: Not having to use ReturnType is soooo much better

PR #438, opened 2025-07-11
fix(codegen): update arg fields' naming when an std arg is featured

pyncz:
## Description

Fixes a mismatch in generated function arguments' naming.

## Test plan

Sorry, I didn't add tests because there's no separate util for just arg building to unit-test it, and I didn't want to refactor the whole generator obviously.

But basically the easiest way to check the diff is to add a function like this in the tested move contracts (e.g. in `./tests/move/wal/sources/wal.move`):

```move
public fun do_something(
  clock: &sui::clock::Clock,
  value: u8
) {
  // ...
}
```

Before the fix, codegen would generate the following args:
```ts
export interface DoSomethingArguments {
    clock: RawTransactionArgument<number>;
}
```
(but actually it's `value` expected there and not `clock`)

After the fix, it's as expected:
```ts
export interface DoSomethingArguments {
    value: RawTransactionArgument<number>;
}
```

---

cc @hayes-mysten

### DIFF
--- a/.changeset/tasty-bikes-smoke.md
+++ b/.changeset/tasty-bikes-smoke.md
@@ -1,0 +1,5 @@
+---
+'@mysten/codegen': patch
+---
+
+Update arg fields' naming when an std arg is featured


hayes-mysten (approved):
thanks so much!

PR #446, opened 2025-07-17
walrus file api

hayes-mysten:
## Description

## Test plan

---


packages/walrus/src/client.ts
```diff
@@ -1926,6 +1934,50 @@ export class WalrusClient {
...
+		blobs: {
+			contents: Uint8Array;
+			identifier: string;
+			tags?: Record<string, string>;
+		}[];
+	}) {
+		const systemState = await this.systemState();
+		const encoded = encodeQuilt({
+			blobs,
+			numShards: systemState.committee.n_shards,
```
hayes-mysten, line 1975: should we find a better name than `numShards`?

packages/walrus/src/files/blob.ts
```diff
@@ -0,0 +1,95 @@
...
+		this.#client = client;
+	}
+
+	// Get the blob as a file (i.e. do not use Quilt encoding)
+	asFile() {
+		return new WalrusFile({ reader: this.#reader });
+	}
+
+	async blobId() {
+		// TODO: implement for local blobs
```
hayes-mysten, line 25: Not sure if we will actually have a concept of local blobs, I think local files is probably good enough for now

packages/walrus/src/files/blob.ts
```diff
@@ -0,0 +1,95 @@
...
+
+	async #blobStatus() {
+		if (!this.#status) {
+			this.#status = this.#client.getVerifiedBlobStatus({ blobId: await this.blobId() });
+		}
+
+		try {
+			return await this.#status;
+		} catch (error) {
+			this.#status = null;
```
hayes-mysten, line 74: I think this has the potential to create a race condition where we could load status multiple times in parallel if things are sequenced exactly the right way.  This should probably just use a `ClientCache`

packages/walrus/src/files/readers/blob.ts
```diff
@@ -0,0 +1,124 @@
...
+	#isQuilt?: boolean;
+
+	constructor({ client, blobId, bytes }: BlobReaderOptions) {
+		this.#client = client;
+		this.blobId = blobId;
+		this.#blobBytes = bytes ?? null;
+	}
+
+	async getIdentifier() {
+		return null;
```
hayes-mysten, line 34: should we return the blobId instead?

packages/walrus/src/files/readers/blob.ts
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// TODO: this is currently async incase we want to do some validation or check to ensure this is actually a quilt
+	async getQuiltReader() {
+		return new QuiltReader({ blob: this });
+	}
+
+	async getBytes() {
+		if (!this.#blobBytes) {
```
hayes-mysten, line 47: use clientCache

packages/walrus/src/files/readers/blob.ts
```diff
@@ -0,0 +1,124 @@
...
+			}
+
+			const metadata = await this.getMetadata();
+			return Number(metadata.metadata.V1.unencoded_length);
+		});
+	}
+
+	async getSizes() {
+		const { committee } = await this.#client.systemState();
+		return getSizes(await this.getSize(), committee.n_shards);
```
hayes-mysten, line 96: blobSize is a little expensive to calculate in some cases.  We need row/column sizes but not blobSize when  reading the quilt.

This might be worth breaking out or simplifying so we don't get the blobSize when its not needed

packages/walrus/src/files/readers/blob.ts
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	// TODO: We should handle retries and epoch changes
```
hayes-mysten, line 99: I think this might be easy to do using the clients existing retry logic

packages/walrus/src/files/readers/quilt.ts
```diff
@@ -0,0 +1,235 @@
...
+
+		const nSlivers = Math.ceil(length / columnSize);
+		const slivers = new Array(nSlivers)
+			.fill(0)
+			.map((_, i) => this.#blob.getSecondarySliver({ sliverIndex: sliver + columnOffset + i }));
+
+		// ignore errors from slivers that are not consumed below
+		slivers.forEach((p) => p.catch(() => {}));
+
+		for await (const sliver of slivers) {
```
hayes-mysten, line 52: TIL: you can use for await when you have an iterables of promises, not just for async iterables

packages/walrus/src/files/readers/quilt.ts
```diff
@@ -0,0 +1,235 @@
...
+	async #readBytesFromBlob(startColumn: number, length: number, offset = 0) {
+		const result = new Uint8Array(length);
+
+		if (!length) {
+			return result;
+		}
+
+		const [blob, { rowSize, symbolSize }] = await Promise.all([
+			this.#blob.getBytes(),
+			this.#blob.getSizes(),
```
hayes-mysten, line 79: if we don't refactor how sizes works above, it would be more efficient to call this after loading bytes is complete

packages/walrus/src/utils/index.ts
```diff
@@ -199,3 +220,30 @@ export function urlSafeBase64(bytes: Uint8Array): string {
 export function fromUrlSafeBase64(base64: string): Uint8Array {
 	return fromBase64(base64.replaceAll('-', '+').replaceAll('_', '/'));
 }
+
+export function getSizes(blobSize: number, numShards: number) {
+	const encodedBlobSize = encodedSliverSize(blobSize, numShards);
+	const { primarySymbols, secondarySymbols } = getSourceSymbols(numShards);
+	const totalSymbols = primarySymbols * secondarySymbols;
+	const totalSymbols2 = (primarySymbols + secondarySymbols) * numShards;
```
hayes-mysten, line 228: this was left over from some debugging, should be cleaned up

Jordan-Mysten (approved):
Generally LGTM, less concerned about the reader APIs since they are internal

packages/walrus/src/files/readers/blob.ts
hayes-mysten: its just the identifier, so I'll leave it nullable

PR #447, opened 2025-07-17
fix(seal): deprecations and improvements

joyqvq:
## Description

## Test plan

---


packages/seal/src/key-server.ts
```diff
@@ -170,3 +185,90 @@ export class BonehFranklinBLS12381DerivedKey implements DerivedKey {
...
+ * Options for fetching keys from the key server.
+ */
+export interface FetchKeysOptions {
+	/** The URL of the key server. */
+	url: string;
+	/** The Base64 string of request signature. */
+	requestSig: string;
+	/** The transaction bytes. */
+	txBytes: Uint8Array;
+	/** The ephemeral public key. */
```
benr-ml, line 199:
```suggestion
	/** The ephemeral secret key. */
```

packages/seal/src/shamir.ts
```diff
@@ -241,9 +241,11 @@ export class Polynomial {
 
 	/** Evaluate the polynomial at x. */
 	evaluate(x: GF256): GF256 {
-		return this.coefficients
-			.toReversed()
-			.reduce((sum, coefficient) => sum.mul(x).add(coefficient), GF256.zero());
+		let result = GF256.zero();
```
benr-ml, line 244: unless there is a bug, i'd prefer to revert it (already audited)

packages/seal/src/encrypt.ts
```diff
@@ -51,8 +51,7 @@ export async function encrypt({
 		keyServers.length < threshold ||
 		threshold === 0 ||
 		keyServers.length > MAX_U8 ||
-		threshold > MAX_U8 ||
-		!isValidSuiObjectId(packageId)
```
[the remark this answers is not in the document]
joyqvq: the caller of this function checked using getObject already

```
		const packageObj = await this.#suiClient.core.getObject({ objectId: packageId });
		if (String(packageObj.object.version) !== '1') {
			throw new InvalidPackageError(`Package ${packageId} is not the first version`);
		}

		return encrypt({
			keyServers: await this.#getWeightedKeyServers(),
			kemType,
			threshold,
			packageId,
			id,
			encryptionInput: this.#createEncryptionInput(demType, data, aad),
		});
```

packages/seal/src/shamir.ts
joyqvq: ok, not a bug, but an efficiency improvement (2x)

packages/seal/src/client.ts
```diff
@@ -219,7 +218,7 @@ export class SealClient {
 
 	async getKeyServers(): Promise<Map<string, KeyServer>> {
 		if (!this.#keyServers) {
-			this.#keyServers = this.#loadKeyServers().catch((error) => {
+			this.#keyServers = await this.#loadKeyServers().catch((error) => {
```
hayes-mysten, line 221: this change may lead to double fetching, since you only assign after the request completes, if you have multiple calls to getKeyServers before the first call completes, it will fetch again in parallel

packages/seal/src/key-server.ts
```diff
@@ -26,19 +33,21 @@ export enum KeyServerType {
 }
 
 export const SERVER_VERSION_REQUIREMENT = new Version('0.4.1');
+export const MYSTEN_LABS_KEY_SERVER_URL =
```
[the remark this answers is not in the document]
hayes-mysten: I think this doesn't actually scale that well.

Maybe something like:

```ts
const TESTNET_KEY_SERVERS: { name: string, objectId: string }[] = [{
  name: 'Mysten Labs 1',
  objectId: '0x73d05d62c18d9374e3ea529e8e0ed6161da1a141a94d3f76ae3fe4e99356db75'
}]
```

packages/seal/src/types.ts
```diff
@@ -1,9 +1,17 @@
...
+	Experimental_CoreClient,
+	Experimental_SuiClientTypes,
+} from '@mysten/sui/experimental';
 
 export type KeyCacheKey = `${string}:${string}`;
 export type SealCompatibleClient = ClientWithExtensions<{
-	core: Experimental_CoreClient;
+	core: Experimental_CoreClient & {
+		verifyZkLoginSignature: NonNullable<
+			Experimental_SuiClientTypes.TransportMethods['verifyZkLoginSignature']
```
hayes-mysten, line 14: This shouldn't be necessary, `verifyZkLoginSignature` is now a required part of the core api (if you were seeing errors, run `turbo build` to ensure the dependencies all get built to update types locally)

packages/seal/src/key-server.ts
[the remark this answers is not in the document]
joyqvq: the name will not be used. we will not have any new key servers. there will be 2 for testnet and 1 for mainnet only.

packages/seal/src/client.ts
joyqvq: shouldnt the private variable be a map instead of a promise<map>
```
#keyServers: Map<string, KeyServer> | null = null;
```

hayes-mysten (approved):

PR #449, opened 2025-07-20
[enoki-wallet] support playtron auth provider

pchrysochoidis:
## Description

* add pkce auth flow which is the only one supported be playtron

## Test plan

👀 

---


StefPler (approved):
Looks clean ✨

PR #451, opened 2025-07-21
fix(bcs): update type inferring in generic `option`/`vector`/`map` helpers

pyncz:
## Description

Refactor typings for `option`, `vector`, `map` and `fixedArray` bcs constructors, so the types are inferred properly when called without explicit type params.

## Test plan

> How did you test the new or updated feature?

Vitest type tests (see `bcs.type-d.ts`)

--

### Some more context

Actually, everything started from the codegen. Noticed that some structs that use options are not properly typed. E.g. `sui::borrow::Referent` codegen was the following:
```ts
export function Referent<T extends BcsType<any>>(...typeParameters: [
    T
]) {
    return bcs.struct('Referent', {
        id: bcs.Address,
        value: bcs.option(typeParameters[0])
    });
}
```

As you might notice, the `T` here is the underlying Struct. But when we call `bcs.option(underlyingStruct)` here, its Input and Type can't be inferred natively (additional `InferBcsType` / `InferBcsInput` are needed), so the resulting struct type ends up being **any** (`{ id: string, value: any }`).

Alternative way to fix that would be to tweak the codegen itself and go with smth like this:
```ts
export function Referent<T extends BcsType<any>>(...typeParameters: [
    T
]) {
    return bcs.struct('Referent', {
        id: bcs.Address,
        value: bcs.option<InferBcsType<T>, InferBcsInput<T>>(typeParameters[0])
    });
}
```
...but imo, makes way more sense in `bcs` itself. Lmk your thoughts!

### DIFF
--- a/packages/bcs/src/bcs.test-d.ts
+++ b/packages/bcs/src/bcs.test-d.ts
@@ -6,7 +6,7 @@ import { bcs } from './bcs.js';
 import { it, describe, expectTypeOf } from 'vitest';
 
 describe('generic bcs helpers should infer type correctly', () => {
-	it('options is typed properly', () => {
+	it('option is typed properly', () => {
 		function TestStruct<T extends BcsType<any>>(...typeParameters: [T]) {
 			return bcs.struct('TestStruct', {
 				value: bcs.option(typeParameters[0]),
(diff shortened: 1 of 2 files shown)


hayes-mysten:
I'll probably have some time to play with this more later this week.  I think moving in this direction will also simplify the other PR I was working on for tracking type-names through generics: https://github.com/MystenLabs/ts-sdks/pull/429

We can probably combine these into one solution that solves both problems

hayes-mysten:
I've pulled your commits/changes into https://github.com/MystenLabs/ts-sdks/pull/429 and updated it to use overloads as described above

PR #458, opened 2025-07-23
[sui] fix tx.pure return type when called with a pure typename

hayes-mysten:
## Description

## Test plan

---


packages/typescript/src/bcs/pure.ts
```diff
@@ -48,9 +48,10 @@ export type ShapeFromPureTypeName<T extends PureTypeName> = T extends BasePureTy
 			? ShapeFromPureTypeName<U> | null
 			: never;
 
-type PureTypeValidationError<T extends string> = T & {
+type PureTypeValidationError<T extends string> = {
```
hayes-mysten, line 51: This error message wasn't actually rendering properly in tooltips, this should make it a little better

williamrobertson13 (approved):

PR #470, opened 2025-07-28
Add IKA token to Deepbook

tonylee08:
## Description

Add IKA token to Deepbook

## Test plan

No testing needed, constant update only


Bridgerz (approved):
LGTM

PR #490, opened 2025-08-06
feat: WalletConnect Connector

tomiir:
## Description

- Adds `WalletConnectWallet` standard wallet connector to connect via the WalletConnect relay
- Automatically adds the wallet if a `walletConnectProjectId` is provided in the configuration for `createDappKit`

## Test plan
- Added demo app based off enoki-connect demo 
- Manually validated supported methods

### DIFF
--- a/packages/dapp-kit-next/examples/react/simple/src/dApp-kit.ts
+++ b/packages/dapp-kit-next/examples/react/simple/src/dApp-kit.ts
@@ -6,6 +6,7 @@ import { getFullnodeUrl, SuiClient } from '@mysten/sui/client';
 
 export const dAppKit = createDAppKit({
 	enableBurnerWallet: import.meta.env.DEV,
+	walletConnectProjectId: 'your_project_id',
 	networks: ['mainnet', 'testnet'],
 	defaultNetwork: 'testnet',
 	createClient(network) {
--- a/packages/dapp-kit-next/packages/dapp-kit-core/src/wallets/walletconnect.ts
+++ b/packages/dapp-kit-next/packages/dapp-kit-core/src/wallets/walletconnect.ts
@@ -1,0 +1,17 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { registerWalletConnectWallet } from '@mysten/walletconnect-wallet';
+import type { WalletInitializer } from './index.js';
+
+export function walletConnectWalletInitializer(projectId: string): WalletInitializer {
+	return {
+		id: 'walletconnect-wallet-initializer',
+		async initialize() {
+			const result = await registerWalletConnectWallet(projectId);
+
+			if (!result) throw new Error('Registration un-successful.');
+			return { unregister: result.unregister };
+		},
+	};
+}
--- a/packages/dapp-kit/src/components/WalletProvider.tsx
+++ b/packages/dapp-kit/src/components/WalletProvider.tsx
@@ -25,6 +25,7 @@ import { createInMemoryStore } from '../utils/stateStorage.js';
 import { getRegisteredWallets } from '../utils/walletUtils.js';
 import { createWalletStore } from '../walletStore.js';
 import { InjectedThemeStyles } from './styling/InjectedThemeStyles.js';
+import { useWalletConnectWallet } from '../hooks/wallet/useWalletConnectWallet.js';
 
 export type WalletProviderProps = {
 	/** A list of wallets that are sorted to the top of the wallet list, if they are available to connect to. By default, wallets are sorted by the order they are loaded in. */
@@ -50,6 +51,9 @@ export type WalletProviderProps = {
 
 	/** The theme to use for styling UI components. Defaults to using the light theme. */
 	theme?: Theme | null;
+
+	/** The project id to use for the WalletConnect wallet. */
+	walletConnectProjectId?: string;
 
 	children: ReactNode;
 };
@@ -94,7 +98,12 @@ export function WalletProvider({
 
 type WalletConnectionManagerProps = Pick<
 	WalletProviderProps,
-	'preferredWallets' | 'walletFilter' | 'enableUnsafeBurner' | 'slushWallet' | 'children'
+	| 'preferredWallets'
+	| 'walletFilter'
+	| 'enableUnsafeBurner'
+	| 'slushWallet'
+	| 'children'
+	| 'walletConnectProjectId'
 >;
 
 function WalletConnectionManager({
@@ -103,10 +112,12 @@ function WalletConnectionManager({
 	enableUnsafeBurner = false,
 	slushWallet,
 	children,
+	walletConnectProjectId,
 }: WalletConnectionManagerProps) {
 	useWalletsChanged(preferredWallets, walletFilter);
 	useWalletPropertiesChanged();
 	useSlushWallet(slushWallet);
+	useWalletConnectWallet(walletConnectProjectId);
 	useUnsafeBurnerWallet(enableUnsafeBurner);
 	useAutoConnectWallet();
 
--- a/packages/dapp-kit/src/hooks/wallet/useWalletConnectWallet.ts
+++ b/packages/dapp-kit/src/hooks/wallet/useWalletConnectWallet.ts
@@ -1,0 +1,37 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { registerWalletConnectWallet } from '@mysten/walletconnect-wallet';
+import { useLayoutEffect } from 'react';
+
+export interface WalletConnectWalletConfig {
+	projectId?: string;
+}
+
+export function useWalletConnectWallet(projectId?: string) {
+	useLayoutEffect(() => {
+		if (!projectId) {
+			return;
+		}
+
+		let cleanup: (() => void) | undefined;
+		let isMounted = true;
+
+		try {
+			const result = registerWalletConnectWallet(projectId);
+
+			if (isMounted && result) {
+				cleanup = result.unregister;
+			} else if (result) {
+				result.unregister();
+			}
+		} catch (error) {
+			console.error('Failed to register Slush wallet:', error);
+		}
+
+		return () => {
+			isMounted = false;
+			if (cleanup) cleanup();
+		};
+	}, [projectId]);
+}
--- a/packages/walletconnect-wallet/CHANGELOG.md
+++ b/packages/walletconnect-wallet/CHANGELOG.md
@@ -1,0 +1,177 @@
+# @mysten/slush-wallet
+
+## 0.1.16
+
+### Patch Changes
+
+- Updated dependencies [c76ddc5]
+  - @mysten/sui@1.36.1
+  - @mysten/wallet-standard@0.16.6
+
+## 0.1.15
+
+### Patch Changes
+
+- 1c4a82d: update links in package.json
+- Updated dependencies [1c4a82d]
+- Updated dependencies [783bb9e]
+- Updated dependencies [783bb9e]
+- Updated dependencies [5cbbb21]
+  - @mysten/window-wallet-core@0.0.6
+  - @mysten/utils@0.1.1
+  - @mysten/sui@1.36.0
+  - @mysten/wallet-standard@0.16.5
+
+## 0.1.14
+
+### Patch Changes
+
+- Updated dependencies [888afe6]
+  - @mysten/sui@1.35.0
+  - @mysten/wallet-standard@0.16.4
+
+## 0.1.13
+
+### Patch Changes
+
+- Updated dependencies [3fb7a83]
+  - @mysten/sui@1.34.0
+  - @mysten/wallet-standard@0.16.3
+
+## 0.1.12
+
+### Patch Changes
+
+- Updated dependencies [a00522b]
+- Updated dependencies [a00522b]
+  - @mysten/sui@1.33.0
+  - @mysten/utils@0.1.0
+  - @mysten/wallet-standard@0.16.2
+  - @mysten/window-wallet-core@0.0.5
+
+## 0.1.11
+
+### Patch Changes
+
+- Updated dependencies [6b7deb8]
+  - @mysten/sui@1.32.0
+  - @mysten/wallet-standard@0.16.1
+
+## 0.1.10
+
+### Patch Changes
+
+- d0a406a: Authorize previously saved accounts without needing to call `connect`
+- Updated dependencies [1ff4e57]
+- Updated dependencies [550e2e3]
+- Updated dependencies [550e2e3]
+  - @mysten/sui@1.31.0
+  - @mysten/wallet-standard@0.16.0
+
+## 0.1.9
+
+### Patch Changes
+
+- Updated dependencies [5bd6ca3]
+  - @mysten/sui@1.30.5
+  - @mysten/wallet-standard@0.15.6
+
+## 0.1.8
+
+### Patch Changes
+
+- Updated dependencies [5dce590]
+- Updated dependencies [4a5aef6]
+  - @mysten/sui@1.30.4
+  - @mysten/wallet-standard@0.15.5
+
+## 0.1.7
+
+### Patch Changes
+
+- bb7c03a: Update dependencies
+- Updated dependencies [4457f10]
+- Updated dependencies [bb7c03a]
+  - @mysten/sui@1.30.3
+  - @mysten/window-wallet-core@0.0.4
+  - @mysten/wallet-standard@0.15.4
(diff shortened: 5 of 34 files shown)


packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+
+async function getTransaction(digest: string, chain: 'mainnet' | 'testnet' | 'devnet') {
+	const client = new SuiClient({ url: getFullnodeUrl(chain) });
```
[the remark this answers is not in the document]
hayes-mysten: For the enoki wallet we pass in clients like this:  https://github.com/MystenLabs/ts-sdks/blob/e315e040314edc47b81bf71098e33ee941dd849d/packages/enoki/src/wallet/types.ts#L89-L102

In dapp-kit-core this can work something like this: https://github.com/MystenLabs/ts-sdks/blob/e315e040314edc47b81bf71098e33ee941dd849d/packages/enoki/src/wallet/initializer.ts#L20-L24

In the `@mysten/dapp-kit` we currently just register enoki with a single client/network that gets re-registered in network change: https://github.com/MystenLabs/ts-sdks/blob/e315e040314edc47b81bf71098e33ee941dd849d/packages/enoki/demo/RegisterEnokiWallets.tsx#L29

but this could probably also be updated to work with multiple networks

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+	const response = await client.getTransactionBlock({
+		digest: digest,
+		options: {
+			showInput: true,
+			showEffects: true,
+			showEvents: true,
```
hayes-mysten, line 40: Based on the usage below, you probably want showRawEffects and showRawInput which will give you the base64 encoded bcs bytes for the transaction and effects

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+			},
+			chain,
+		)) as { digest: string };
+
+		const [_, chainId] = chain.split(':');
+		const tx = await getTransaction(response.digest, chainId as 'mainnet' | 'testnet' | 'devnet');
+
+		return {
+			digest: response.digest,
+			effects: tx.effects?.toString() ?? '',
```
hayes-mysten, line 232: tx.rawEffects is probably what you want here

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+				params: {
+					transaction: data,
+					address: account.address,
+				},
+			},
+			chain,
+		)) as { digest: string };
+
+		const [_, chainId] = chain.split(':');
+		const tx = await getTransaction(response.digest, chainId as 'mainnet' | 'testnet' | 'devnet');
```
hayes-mysten, line 228: rather than using this, what I would actually recommend is using the new `core` APIs.  Once your wallet has access to a `client` instance (mentioned in another comment) it should have a `core` API that will be more future proof.

The `core` methods will exist on all Sui clients (json rpc, graphql, grpc).

You should be able to use `client.core.getTransaction({ digest })` which will give you back all the needed transaction data.  The returned data has the same shape as signer.signAndExecuteTransaction and can be used similar to this: https://github.com/MystenLabs/ts-sdks/blob/e315e040314edc47b81bf71098e33ee941dd849d/packages/enoki/src/wallet/wallet.ts#L219-L230

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+				pubkey: string;
+			}[];
+		}
+
+		return accounts.map((account) => {
+			return new ReadonlyWalletAccount({
+				address: account.address,
+				chains: SUI_CHAINS,
+				features: walletAccountFeatures,
+				publicKey: fromBase64(account.pubkey),
```
hayes-mysten, line 283: We should double check that this is correctly formatted.  This should be the "suiBytes" meaning the raw publicKey prefixed with the scheme bit for base keypairs.  If this also supports zklogin, passkeys, or multisig its a little more complicated

packages/walletconnect-wallet/CHANGELOG.md
```diff
@@ -0,0 +1,177 @@
+# @mysten/slush-wallet
```
hayes-mysten, line 1: delete this

packages/walletconnect-wallet/package.json
```diff
@@ -0,0 +1,57 @@
...
+		"@types/node": "^22.15.29",
+		"typescript": "^5.8.3",
+		"vitest": "^3.2.4"
+	},
+	"dependencies": {
+		"@mysten/sui": "workspace:*",
+		"@mysten/utils": "workspace:*",
+		"@mysten/wallet-standard": "workspace:*",
+		"@mysten/window-wallet-core": "workspace:*",
+		"@reown/appkit-universal-connector": "1.7.19-uc-config.1.0",
```
hayes-mysten, line 53: I think with prerelease external packages in this adapter, I'd prefer to avoid the direct dependency on wallet-connect in dapp-kit and dapp-kit-core at least until this has stabilized more

packages/walletconnect-wallet/README.md
```diff
@@ -0,0 +1 @@
+# `@mysten/slush-wallet`
```
hayes-mysten, line 1: wrong package name

packages/walletconnect-wallet/package.json
```diff
@@ -0,0 +1,57 @@
...
+		".": {
+			"import": "./dist/esm/index.js",
+			"require": "./dist/cjs/index.js"
+		}
+	},
+	"sideEffects": false,
+	"files": [
+		"CHANGELOG.md",
+		"dist",
+		"experimental"
```
hayes-mysten, line 21: whats this for?

packages/walletconnect-wallet/typedoc.json
```diff
@@ -0,0 +1,6 @@
+{
```
hayes-mysten, line 1: for this to be picked up in the main typedocs config, you also need an entry in the main typedoc config in the docs package

williamrobertson13 (commented):
Mostly smaller comments, I think @hayes-mysten covered all the major stuff. Still need to give it a spin locally but am very excited about this 🎉

packages/walletconnect-wallet/src/wallet/index.ts
hayes-mysten: I am also realizing there is a race condition here.  You should use `waitForTransaction` instead of `getTransaction` (assuming this only returns on success) because there is a chance the RPC node the app is talking to has not picked up the new transaction after the digest is returned from the wallet

packages/dapp-kit-next/packages/dapp-kit-core/src/wallets/walletconnect.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { registerWalletConnectWallet } from '@mysten/walletconnect-wallet';
+import type { WalletInitializer } from './index.js';
+
+export function walletConnectWalletInitializer(projectId: string): WalletInitializer {
+	return {
+		id: 'walletconnect-wallet-initializer',
```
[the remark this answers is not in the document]
tomiir: Not a usual case but adding it in case there's some specific implementation that requires multiple instances

packages/walletconnect-wallet/src/wallet/index.ts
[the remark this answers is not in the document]
tomiir: @hayes-mysten I added a `getClient` parameter. 
What I found interesting is this mechanism is already considered on `createDappKit`. Did not find the exact equivalent in react context `WalletProvider` but I did find a `SuiContext` with a `client` which I saw is updated on network switches so I made use of it. Let me know if it works 🫡  (pushing in a second)

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,381 @@
...
+		account,
+		chain,
+	}) => {
+		const data = await transaction.toJSON();
+
+		const response = (await this.#connector?.request(
+			{
+				method: 'sui_signAndExecuteTransaction',
+				params: {
+					transaction: data,
```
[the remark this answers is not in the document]
tomiir: will push to team for an update on the docs as this is already correctly parsed in our exampel wallet 🙏

tomiir:
@williamrobertson13 @hayes-mysten Thank you very much for your thorough reviews 🙏 

I believe I've addressed all of the issues, and for those that I haven't I left corresponding questions. 

Let me know if something is missing and I'll promptly amend! 🫡

packages/dapp-kit-next/examples/react/simple/src/dApp-kit.ts
```diff
@@ -6,6 +6,7 @@ import { getFullnodeUrl, SuiClient } from '@mysten/sui/client';
 
 export const dAppKit = createDAppKit({
 	enableBurnerWallet: import.meta.env.DEV,
+	walletConnectProjectId: 'your_project_id',
```
hayes-mysten, line 9: because this requires initialization by apps anyways, I think it would be better to not have this added directly to dapp-kit, and just have apps manually add the initializer from wallet-connect-adapter instead

I would feel a lot more comfortable getting this merged if it is not automatically included in dapp-kit in its initial version.  I think we can revisit including it by default in the fututre, but for now keeping everything in the walletconnect-adapter package with an easy to use initializer should not add much work for devs, and reduces the number of things that get installed for anyone not using walletconnect

packages/dapp-kit/src/components/WalletProvider.tsx
```diff
@@ -51,6 +52,9 @@ export type WalletProviderProps = {
 	/** The theme to use for styling UI components. Defaults to using the light theme. */
 	theme?: Theme | null;
 
+	/** The project id to use for the WalletConnect wallet. */
+	walletConnectProjectId?: string;
```
hayes-mysten, line 56: same comment as dapp-kit-core

packages/dapp-kit/src/hooks/wallet/useWalletConnectWallet.ts
```diff
@@ -0,0 +1,36 @@
...
+import { registerWalletConnectWallet } from '@mysten/walletconnect-wallet';
+
+import { useLayoutEffect } from 'react';
+import { useSuiClientContext } from '../useSuiClient.js';
+
+export interface WalletConnectWalletConfig {
+	projectId?: string;
+}
+
+export function useWalletConnectWallet({ projectId }: WalletConnectWalletConfig) {
```
hayes-mysten, line 13: This is one of the pieces that made the dapp-kit hard to build for, and why its being replaced.  This hook really doesn't belong in dapp-kit.

Unfortunately the alternative forces you to have a peer dependency on react in the adapter package which is also not great.

There are 2 not so good options I can think of to make this work:

We could add a generic useSingleNetworkWalet that takes a register function so that someone could do something like:

```ts
import { useSingleNetworkWallet } from '@mysten/dapp-kit'
import { registerWalletConnectWallet } from '@mysten/walletconnect-wallet';
...
useSingleNetworkWalet(registerWalletConnectWallet)
```

This is probably oversimplifying things a lot, but this would enable us to keep the wallet implementations in their own packages without forcing them to deal with react directly.

The even easier (short term?) solution would be to just put this hook in the docs/setup instructions. 

I don't love it, but its one way to avoid having to solve this until we can get everything migrated to dapp-kit-core.

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,406 @@
...
+		};
+	};
+
+	#signAndExecuteTransaction: SuiSignAndExecuteTransactionMethod = async ({
+		transaction,
+		account,
+		chain,
+	}) => {
+		const [, network] = chain.split(':');
+		const client = this.#getClient(network as Network);
```
hayes-mysten, line 241: This is potentially dangerous with the way the dapp-kit hook was implemented to just give you the current client.  I think this side is right, but if a dapp has its current network as mainnet, and then tries to sign a testnet transaction without changing the active network, the tx will be built built for the wrong network.

I think this is potentially okay, since it should still get executed against the correct network, which will cause it to error, but maybe something worth thinking about.

Once solution might be adding a check in the hook that registers the wallet that does something like:

`getClient: (network) => { if (network !== currentClient.network) { throw } return client }`

hayes-mysten (commented):
Is there a way that this can be tested?  I am not familiar with how the other side of the wallet-connect flow works

packages/dapp-kit-next/examples/react/simple/src/dApp-kit.ts
tomiir: I think there is no problem with this, will remove from here and leave it up for devs to initialize. 
If there are any docs to update to list WC as an available connector, let me know!

packages/dapp-kit/src/hooks/wallet/useWalletConnectWallet.ts
tomiir: I think it's easier to just remove this and keep `registerWalletConnectWallet` as the main/only entrypoint and let devs instantiate as they want and not couple the dependencies.

tomiir:
@hayes-mysten thanks again for the review 🙏 
By testing you mean as in testing coverage (unit/e2e) or as in manual testing against a wallet? 
I can deploy my changes on our example wallet if that's what you require !

hayes-mysten:
> By testing you mean as in testing coverage (unit/e2e) or as in manual testing against a wallet?
I can deploy my changes on our example wallet if that's what you require !

Yes, if there was an easy way to manually run through a couple flows with a demo wallet app that would be cool, not necessarily a blocker, but can be helpful to make sure everything is actually working end to end

tomiir:
@hayes-mysten I'm trying to get it merged but here's a preview you can use to test this 🙏 
https://react-wallet-v2-git-feat-sui-get-accounts-reown-com.vercel.app/walletconnect

packages/walletconnect-wallet/src/wallet/index.ts
hayes-mysten: You should be able to use `effects: toBase64(result.effects.bcs!), ` with that api

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,408 @@
...
+import mitt from 'mitt';
+import type { InferOutput } from 'valibot';
+import { boolean, object, string } from 'valibot';
+import type { CustomCaipNetwork } from '@reown/appkit-universal-connector';
+import { UniversalConnector } from '@reown/appkit-universal-connector';
+import type {
+	Experimental_BaseClient,
+	Experimental_SuiClientTypes,
+} from '@mysten/sui/experimental';
+import { Transaction } from '@mysten/sui/dist/cjs/transactions/Transaction';
```
hayes-mysten, line 44:
```suggestion
import { Transaction } from '@mysten/transactions';
```

hayes-mysten (commented):
I think this PR looks like its just about ready to ship.  I think there are a couple of unaddressed comments, and some build failures that need to get cleaned.  Are there upstream changes we are waiting on before this can be merged and used?

packages/walletconnect-wallet/src/wallet/index.ts
```diff
@@ -0,0 +1,408 @@
...
+						'sui_signAndExecuteTransaction',
+						'sui_getAccounts',
+					],
+					events: ['chainChanged', 'accountsChanged'],
+					chains: SUICaipNetworks,
+				},
+			],
+			modalConfig: {
+				themeVariables: {
+					'--w3m-z-index': 2147483647,
```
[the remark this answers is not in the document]
tomiir: I thought I'd included this in the last version, but didn't. It will be included in next version. Using a canary image with latest published + this flag but can wait until stable is released on monday to merge this pr 🫡

tomiir:
Hey @hayes-mysten ! Thanks again (now for the QA 🙏 ) 

I fixed the issue with the signatures, was wrongly base64 encoding. 
Added the new version with z-index manipulation as well. Stable version including this change is planned to go live next monday, using a canary version atm so you can take a look 🫡

Wondering what other comments are still unresolved? I believe I've removed the dependency on WC from the ui level libs and left the register method alone which I think was the main point of discussion. Let me know and I can address ASAP!

hayes-mysten (approved):
Code generally looks like its in a good place once the CI checks are all passing

williamrobertson13 (approved):

PR #497, opened 2025-08-11
fix: use SharedObjectRef for system objects instead of UnresolvedObject

### LINKED ISSUE #424: tx.object.clock() should add a fully resolved object to tx instead of an unresolved object
## Steps to Reproduce Issue

Fill this in with the concrete steps needed to reproduce the bug. When providing code in the reproduction steps, use the smallest snippet of code that demonstrates the issue, removing any extraneous details.

e.g.

1. Call function <function call>.
2. Use return value and call <other function call>.

## Expected Result

Specify what outcome you expected should have resulted, but didn't.

e.g.
Expected <other function call> to return 42.

## Actual Result

Specify what the actual unexpected outcome was.

e.g.
<other function call> returned 41.

## Environment Information

- Runtime: <specify runtime (eg. node@v22.13.0 or chrome@v131.0.6778.205)>
- Package manager: <specify package manager eg. pnpm@9.9.0>

Vorashil:
## Description
This PR fixes #424 by ensuring system objects (Clock, System, Random, DenyList) are added to transactions as fully resolved SharedObject references instead of UnresolvedObjects.

## Problem
Previously, `tx.object.clock()` and other system object methods would add unresolved objects that required network calls to resolve, even though these are well-known shared objects with fixed properties.

## Solution
Modified the object helper methods to return fully resolved `SharedObjectRef` for all system objects:
- **Clock (0x6)**: SharedObject with `initialSharedVersion: 1`, `mutable: false`
- **System (0x5)**: SharedObject with `initialSharedVersion: 1`, `mutable: true`
- **Random (0x8)**: SharedObject with `initialSharedVersion: 1`, `mutable: false`
- **DenyList (0x403)**: SharedObject with `initialSharedVersion: 1`, `mutable: false`

## Benefits
- No network calls needed for system objects
- Better transaction building performance
- Cleaner transaction data structure

## Testing
- ✅ All unit tests pass
- ✅ Lint checks pass
- ✅ Package builds successfully
- ✅ Test snapshots updated to reflect the new behavior

Fixes #424


Vorashil:
I had to remove `initialSharedVersion: 1` from `UnresolvedObject` as it causes Integration tests to fail. I see that you mentioned in your earlier comment that we should have this set

>Internally, if mutable is set, we can fully construct the shared object reference. If not, we can construct an unresolvedObject with initialSharedVersion set. This will avoid the object lookup, but still look up the moveCall to determine if it uses the shared input mutable or not.

Am I missing something? @hayes-mysten

hayes-mysten:
The seal intergation test doesn't seem related, I think it might just be a little flaky.  the `normalizeMoveArguments` tests that were failing I think just need to be updated. 

Was there a specific faulure that made you think setting `initialSharedVersion` was causing an issue?

Vorashil:
I cannot recreate this locally. All tests seem to be passing for me

hayes-mysten:
@Vorashil the failing tests are in the codegen package.  You probably need to run `pnpm turbo build` in the codegen package to build the new version of the sui SDK before you see the changes in the codegen package

.changeset/fix-clock-system-objects.md
```diff
@@ -0,0 +1,15 @@
+---
+"@mysten/sui": patch
+---
+
+Fix system objects to avoid unnecessary network calls
+
+- Clock (0x6) and Random (0x8) now return fully resolved SharedObject references with mutable: false
+- System (0x5) and DenyList (0x403) now accept optional `mutable` parameter:
+  - System defaults to mutable: true (returns SharedObject)
+  - DenyList defaults to mutable: false (returns UnresolvedObject with initialSharedVersion)
```
hayes-mysten, line 10: these don't have defaults anymore right?

PR #498, opened 2025-08-14
[dapp-kit-next] Unit & Integration tests groundwork

gwro:
## Description

## Test plan

---


packages/dapp-kit-next/packages/dapp-kit-core/test/integration/setup.ts
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import { afterEach, beforeEach } from 'vitest';
+
+export function suppressFetchErrorsDuringTeardown() {
```
hayes-mysten, line 6: what fetch requests are we actually making, is this for slush web?  It seems like we should be able to test/use most of dapp-kit without it making a lot of requests (I might be completely wrong here, its been a while since I looked at dapp-kit stuff)

packages/dapp-kit-next/packages/dapp-kit-core/test/unit/stores/compatible-wallets.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import { createMockAccount } from '../../mocks/mock-account';
+import type { MockWalletOptions } from '../../mocks/mock-wallet';
+import {
+	StandardConnect,
+	SuiSignTransaction,
+	SuiSignTransactionBlock,
+} from '@mysten/wallet-standard';
+import { excludeUiWalletsByName } from '../../test-utils';
+
+describe('[Unit] $compatibleWallets', () => {
```
williamrobertson13, line 15: I think these tests are useful 😄, I do wonder if it would be easier / better value to just test these as part of creating the whole dApp Kit instance but some of these cases are more specific edge cases so I'm fine with the setup here!

packages/dapp-kit-next/packages/dapp-kit-core/test/mocks/mock-ui-wallet.ts
```diff
@@ -0,0 +1,65 @@
...
+
+	return {
+		address,
+		publicKey,
+		chains,
+		features,
+	} as UiWallet['accounts'][number];
+}
+
+export function createMockUiWallet(options: MockUiWalletOptions = {}): UiWallet {
```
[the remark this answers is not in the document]
gwro: Sorry, it's just some leftovers from when I was learning the code in the beginning 😁 Removing 💯

PR #512, opened 2025-08-20
Fix remaining balance check in ParallelExecutor

### LINKED ISSUE #511: parallel-executor potentiall issue on the edge case
Hi, not sure if I understand that well the paraller-executor code but:

Here: https://github.com/MystenLabs/ts-sdks/blob/c0c8d8e4ccf94e51af00b9a14b2c021b65f3f1c6/packages/typescript/src/transactions/executor/parallel.ts#L255

We have code: 
```
if (!usesGasCoin && gasCoin.balance >= this.#minimumCoinBalance) {  // <- 1.
  this.#coinPool.push({
    id: gasResult.ref.objectId,
    version: gasResult.ref.version,
    digest: gasResult.ref.digest,
    balance: gasCoin.balance - totalUsed, // <--------------------------- 2.
  });
} else {
 ...
}
```

In 1. we check if `gasCoin.balance >= this.#minimumCoinBalance` in 2 we set balance to gasCoin.balance - totalUsed so it could be actually lower than this.#minimumCoinBalance and in turn potentially lower than this.#defaultGasBudget .
```
	/** The gasBudget to use if the transaction has not defined it's own gasBudget, defaults to `minimumCoinBalance` */
	defaultGasBudget?: bigint;
```

So the next tx that will use this coin will face:
```
["ExecutionError: ExecutionError { inner: ExecutionErrorInner { kind: InvariantViolation, source: Some(\"Transaction input checker should check that there is enough gas\"), command: None } }"]
```

I think it also could only happen in some extreme situtin (like >n txs in the same block where n is number of coins in pool).

I also could be mistaken and the issue is somewhere else - it is kinda hard to reproduce the error so unfortunately I do not have a code snippet for that,

hayes-mysten:
## Description

Fixes #511 

## Test plan

---

### DIFF
--- a/packages/typescript/test/e2e/parallel-executor.test.ts
+++ b/packages/typescript/test/e2e/parallel-executor.test.ts
@@ -1,8 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { afterEach } from 'node:test';
-import { afterAll, beforeAll, beforeEach, describe, expect, it, Mock, vi } from 'vitest';
+import { afterEach, afterAll, beforeAll, beforeEach, describe, expect, it, Mock, vi } from 'vitest';
 
 import { bcs } from '../../src/bcs';
 import { SuiClient } from '../../src/client';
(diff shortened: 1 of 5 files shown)


packages/typescript/test/e2e/parallel-executor.test.ts
```diff
@@ -1,8 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-import { afterEach } from 'node:test';
```
hayes-mysten, line 4: This was killing me.  There were race conditions in this test, and I couldn't figure out what was happening.  Turns out, if you import the wrong afterEach it just silently does not run

packages/zksend/src/index.test.ts
```diff
@@ -1,14 +1,13 @@
...
 
-import { describe } from 'node:test';
 import { getFullnodeUrl, SuiClient, SuiObjectChange } from '@mysten/sui/client';
 import { decodeSuiPrivateKey, Keypair } from '@mysten/sui/cryptography';
 import { getFaucetHost, requestSuiFromFaucetV2 } from '@mysten/sui/faucet';
 import { Ed25519Keypair } from '@mysten/sui/keypairs/ed25519';
 import { Transaction } from '@mysten/sui/transactions';
 import { MIST_PER_SUI, toBase64 } from '@mysten/sui/utils';
-import { beforeAll, expect, test } from 'vitest';
+import { beforeAll, describe, expect, test } from 'vitest';
```
hayes-mysten, line 10: This test doesn't run in CI, but was kinda broken. Noticed it when I looked for other 'node:test' imports

PR #514, opened 2025-08-21
Export BCS types for Deepbook structs

sb-:
## Description

Export BCS types for Deepbook structs for use by other clients (e.g. gRPC), and remove redundant definitions

## Test plan

`pnpm build && pnpm lint`

---


0xaslan (approved):

hayes-mysten:
You can mark these files as generated so they are suppressed in github (see https://github.com/MystenLabs/ts-sdks/blob/main/.gitattributes#L6)

PR #531, opened 2025-09-02
fix(codegen): skip system args in `parameterNames` used for normalization

pyncz:
## Description

Skip system args in codegen's arg names list.

## Test plan

Added the missing test for args as `object` WITH trailing arg after the system one (we had 'object & no trailing', 'array & trailing' and 'array & no trailing' before ._. )

---

### AI Assistance Notice

### DIFF
--- a/.changeset/gold-cooks-pay.md
+++ b/.changeset/gold-cooks-pay.md
@@ -1,0 +1,7 @@
+---
+'@mysten/deepbook-v3': patch
+'@mysten/codegen': patch
+'@mysten/walrus': patch
+---
+
+Update codegen arg normalization for object args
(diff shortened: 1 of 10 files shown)


hayes-mysten:
I think this may actually be wrong.  The parameter names passed in are supposed to only include the non system parameters.  

I think your test case where you pass in `clock` is invalid, the generated code should not be generating `clock` as a parameter name, are you seeing that happening in your generated code?

hayes-mysten:
Ah, I think the part thats actually wrong is here: https://github.com/MystenLabs/ts-sdks/blob/1b52c499dd85dccfdb0f3c07f216ad6f7475b5f4/packages/codegen/src/move-module-builder.ts#L540-L541

This should be using requiredParameters

hayes-mysten (approved):

PR #554, opened 2025-09-19
[dapp-kit-next] add backward compatible auto-connect functionality

williamrobertson13:
## Description
I had a little extra time, so I figured I'd try to get the remaining dApp Kit work out of the way. This PR adds the backward compatible auto-connect functionality for wallets that don't pre-populate the accounts array.

## Test plan
- Auto-connect works normally with all wallets (e.g. Phantom which doesn't populate accounts on load and Backpack which does)

---

### AI Assistance Notice

- [X] I did not use AI for this PR.

### DIFF
--- a/packages/dapp-kit-next/packages/dapp-kit-core/src/core/initializers/autoconnect-wallet.ts
+++ b/packages/dapp-kit-next/packages/dapp-kit-core/src/core/initializers/autoconnect-wallet.ts
@@ -8,15 +8,18 @@ import { type UiWallet } from '@wallet-standard/ui';
 import { getWalletUniqueIdentifier } from '../../utils/wallets.js';
 
 import { internalConnectWallet } from '../actions/connect-wallet.js';
+import { Networks } from '../../utils/networks.js';
 
 /**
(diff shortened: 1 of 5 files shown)


packages/dapp-kit-next/packages/dapp-kit-core/src/core/initializers/autoconnect-wallet.ts
```diff
@@ -65,15 +73,29 @@ async function getSavedWalletAccount({
...
+		(account) => account.address === savedAccountAddress,
+	);
+
+	if (existingAccount) {
+		return existingAccount;
+	}
+
+	// For wallets that don't pre-populate the accounts array on page load,
+	// we need to silently request authorization and get the account directly.
+	const alreadyAuthorizedAccounts = await internalConnectWallet(targetWallet, networks, {
```
williamrobertson13, line 94: I think it makes sense to avoid updating the connection state and then having to deal with the auto-connection state here. I wouldn't be surprised if I was missing a race condition check somewhere, but I think this is okay for now

PR #559, opened 2025-09-23
transaction analyzer

hayes-mysten:
## Description

## Test plan

---

### AI Assistance Notice

- [x] I used AI for docs / tests, but manually wrote the source code.


packages/wallet-sdk/src/transaction-analyzer/base.ts
```diff
@@ -0,0 +1,64 @@
...
+	objectsById: Map<string, AnalyzedObject>;
+	coins: Record<string, AnalyzedCoin>;
+	gasCoins: AnalyzedCoin[];
+	moveFunctions: Experimental_SuiClientTypes.FunctionResponse[];
+	dryRun: Experimental_SuiClientTypes.DryRunTransactionResponse;
+	balanceChanges: Experimental_SuiClientTypes.BalanceChange[];
+	coinFlows: CoinFlow[];
+}
+
+export const baseAnalyzers: {
```
Jordan-Mysten, line 36: Asking a question because I'm too lazy to go look myself: Are these lazy evaluated? I feel like if yes, then cool, and if no, then we _might_ want to make the base analyzers _really_ minimal?

packages/wallet-sdk/src/transaction-analyzer/base.ts
```diff
@@ -0,0 +1,64 @@
...
+	[k in keyof BaseAnalysis]: Analyzer<BaseAnalysis[k]>;
+} = {
+	bytes: (tx, client) => async () => tx.build({ client }),
+	data:
+		(tx) =>
+		async ({ get }) => {
+			// ensure the transaction is built first
+			await get('bytes');
```
Jordan-Mysten, line 44: I get that this is building effectively a DAG imperatively, but I am _slightly_ concerned that these depend on a string key that is configured outside of the function itself. I always feel like making these work by reference instead is typically nicer

packages/wallet-sdk/src/transaction-analyzer/rules/coin-flows.ts
```diff
@@ -0,0 +1,215 @@
...
+	coinType: string;
+	amount: bigint;
+}
+
+export const coinFlows = createAnalyzer({
+	dependencies: { data, commands, inputs, coins, gasCoins },
+	analyze:
+		() =>
+		async ({ data, commands, inputs, coins, gasCoins }) => {
+			if (data.issues || commands.issues || inputs.issues || coins.issues || gasCoins.issues) {
```
hayes-mysten, line 24: this should get wrapped into a helper

packages/wallet-sdk/src/transaction-analyzer/rules/accessLevel.ts
```diff
@@ -0,0 +1,98 @@
...
+						break;
+					case 'Object':
+						updateAccessLevel(arg.object.id, arg.accessLevel);
+						break;
+				}
+			}
+
+			function updateAccessLevel(id: string, level: 'read' | 'mutate' | 'transfer') {
+				const current = accessLevels[id];
+				if (current === 'transfer') {
```
Jordan-Mysten, line 87: You could also make a map of these to numbers and then just check newLevel > oldLevel but this is also fine

packages/wallet-sdk/src/transaction-analyzer/rules/coin-value.ts
```diff
@@ -0,0 +1,75 @@
...
+			coinType: string;
+			decimals: number;
+			price: number | null;
+		}[]
+	>;
+}
+
+export interface CoinValueAnalysis {
+	coinTypesWithoutPrice: string[];
+	total: number;
```
Jordan-Mysten, line 19: Is this total USD value?

packages/wallet-sdk/src/transaction-analyzer/rules/coin-value.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import { createAnalyzer } from '../analyzer.js';
+import { coinFlows } from './coin-flows.js';
+
+export interface CoinValueAnalyzerOptions {
+	getCoinPrices: (coinTypes: string[]) => Promise<
+		{
+			coinType: string;
+			decimals: number;
+			price: number | null;
```
Jordan-Mysten, line 12: We should probably call these `priceUsd` or something just to make this obvious

packages/wallet-sdk/src/transaction-analyzer/rules/coin-value.ts
```diff
@@ -0,0 +1,75 @@
...
+				amount: bigint;
+				convertedAmount: number;
+			}[] = [];
+
+			for (const flow of coinFlows.outflows) {
+				if (flow.amount > 0n) {
+					const result = prices.find((p) => p.coinType === flow.coinType);
+
+					if (result?.price != null) {
+						const amount = (Number(flow.amount) / 10 ** result.decimals) * result.price;
```
Jordan-Mysten, line 52: I think we may need to pull a library to do this math because u64 value + decimals don't map cleanly to double-precision floats

packages/wallet-sdk/src/transaction-analyzer/rules/coins.ts
```diff
@@ -0,0 +1,72 @@
...
+import { normalizeStructTag, parseStructTag } from '@mysten/sui/utils';
+import { data } from './core.js';
+
+export type AnalyzedCoin = AnalyzedObject & { balance: bigint; coinType: string };
+
+export const Coin = bcs.struct('Coin', {
+	id: bcs.Address,
+	balance: bcs.U64,
+});
+const parsedCoinStruct = parseStructTag('0x2::coin::Coin<0x2::sui::SUI>');
```
Jordan-Mysten, line 17: Very funny way to avoid do checks against static strings, but I'm fine with it

packages/wallet-sdk/src/transaction-analyzer/rules/core.ts
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import type { Transaction } from '@mysten/sui/transactions';
+import { TransactionDataBuilder } from '@mysten/sui/transactions';
+import type { ClientWithCoreApi, Experimental_SuiClientTypes } from '@mysten/sui/experimental';
+import type { AnalyzerResult } from '../analyzer.js';
+import { createAnalyzer } from '../analyzer.js';
+
+export const bytes = createAnalyzer({
+	cacheKey: 'bytes@1.0.0',
```
Jordan-Mysten, line 11: Curious why you opted for manual cache key here

Jordan-Mysten (approved):

packages/wallet-sdk/src/transaction-analyzer/rules/coin-value.ts
hayes-mysten: I was specifically trying to stay away from usd so you could use this with multiple currencies (and wallets could choose their own converted currency or base it on where their users are)

packages/wallet-sdk/src/transaction-analyzer/rules/coin-value.ts
hayes-mysten: Not perfectly, but currency converted values are going to be rounded to 2 decimal places, and I'd be surprised if any of these conversions would be lossy enough to matter at that level of precision

packages/wallet-sdk/src/transaction-analyzer/rules/core.ts
hayes-mysten: this one should be removed, I had it because I went through and blindly added it to anything that made requests, but its really not needed here

hayes-mysten: actually, never mind: I added this because this is actually the most important one to not duplicate.

There should only ever be one version of bytes, because this is what you are signing/analyzing/basing all other analysis on.  If the bytes are different, lots of other stuff could be different then everything else might be wrong

PR #610, opened 2025-10-16
[Payment Kit] Added an initial version of Payment Kit SDK documentation

will-mysten:
## Description

- Initial iteration of the Payment Kit SDK documentation. This will continue to undergo changes/improvements as the SDK is updated.

---

### AI Assistance Notice

- [X] I used AI for docs / tests, but manually wrote the source code.

### DIFF
--- a/packages/docs/content/payment-kit/getting-started.mdx
+++ b/packages/docs/content/payment-kit/getting-started.mdx
@@ -110,30 +110,6 @@ if (paymentRecord) {
 }
 ```
 
-## Extracting the Payment Receipt
-
-Payment receipts are emitted as events. You can extract them from the transaction result:
-
-```ts
-// Get events from the transaction result
-const events = result.events || [];
-
-// Find the PaymentReceipt event
-const receiptEvent = events.find((event) => event.type.includes('PaymentReceipt'));
-
-if (receiptEvent) {
-	const receipt = receiptEvent.parsedJson;
-	console.log('Payment Receipt:', {
-		nonce: receipt.nonce,
-		amount: receipt.amount,
-		receiver: receipt.receiver,
-		coinType: receipt.coin_type,
-		timestampMs: receipt.timestamp_ms,
-		paymentType: receipt.payment_type,
-	});
-}
-```
-
 ## Processing an Ephemeral Payment
 
 If you don't need duplicate prevention or persistent storage, use ephemeral payments:
@@ -215,5 +191,4 @@ Now that you understand the basics, explore more advanced features:
 
 - [Payment Processing](./payment-processing) - Deep dive into payment models
 - [Registry Management](./registry-management) - Create and configure custom registries
-- [Examples](./examples) - Common payment scenarios and patterns
-- [API Reference](./api-reference) - Complete API documentation
+- [SDK API Reference](./payment-kit-sdk) - Complete SDK API documentation
--- a/packages/docs/content/payment-kit/index.mdx
+++ b/packages/docs/content/payment-kit/index.mdx
@@ -128,5 +128,4 @@ type PaymentReceipt = {
 - [Getting Started Guide](./getting-started) - Detailed setup and usage instructions
 - [Payment Processing](./payment-processing) - Learn about different payment models
 - [Registry Management](./registry-management) - Create and configure payment registries
-- [API Reference](./api-reference) - Complete API documentation
-- [Examples](./examples) - Common use cases and code samples
+- [SDK API Reference](./payment-kit-sdk) - Complete SDK API documentation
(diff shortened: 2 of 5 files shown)


packages/docs/content/payment-kit/getting-started.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+// Create or load your keypair
+const keypair = Ed25519Keypair.generate();
+const sender = keypair.getPublicKey().toSuiAddress();
+
+// Define payment parameters
+const paymentParams = {
+	nonce: '0f6a09cf-7b46-4e31-adce-f39e463ccae5', // Unique identifier for this payment
+	coinType: '0x2::sui::SUI', // Coin type (SUI in this case)
+	amount: 1000000000, // 1 SUI (in MIST)
```
hayes-mysten: 1n * MIST_PER_SUI

packages/docs/content/payment-kit/getting-started.mdx
```diff
@@ -0,0 +1,227 @@
...
+## Extracting the Payment Receipt
+
+Payment receipts are emitted as events. You can extract them from the transaction result:
+
+```ts
+// Get events from the transaction result
+const events = result.events || [];
+
+// Find the PaymentReceipt event
+const receiptEvent = events.find((event) => event.type.includes('PaymentReceipt'));
```
hayes-mysten: this is exploitable, we probably shouldn't recommend this

packages/docs/content/payment-kit/getting-started.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+```ts
+try {
+	const tx = client.paymentKit.tx.processRegistryPayment(paymentParams);
+	const result = await client.signAndExecuteTransaction({
+		transaction: tx,
+		signer: keypair,
+	});
+
+	console.log('Success:', result.digest);
```
hayes-mysten: we always need to check status, transactions can fail and still have a digest, so you don't know here that the payment actually happened (we should probably confirm this in every example)

packages/docs/content/payment-kit/index.mdx
```diff
@@ -0,0 +1,133 @@
...
+}).$extend(PaymentKitClient.asClientExtension());
+
+// Process a registry-based payment
+const tx = client.paymentKit.tx.processRegistryPayment({
+	nonce: '0f6a09cf-7b46-4e31-adce-f39e463ccae5',
+	coinType: '0x2::sui::SUI',
+	amount: 1000000000, // 1 SUI in MIST
+	receiver: '0x123...abc',
+	sender: '0x456...def',
+	registryName: 'default-payment-registry',
```
hayes-mysten: this seems like it should probably use the constant if its intended to be the default

packages/docs/content/payment-kit/payment-kit-sdk.mdx
```diff
@@ -0,0 +1,614 @@
...
+### PaymentKitClient
+
+The main client class that provides access to Payment Kit functionality.
+
+#### `asClientExtension()`
+
+Static method to create a Payment Kit client extension for `SuiClient`.
+
+```ts
+static asClientExtension(): SuiClientRegistration<
```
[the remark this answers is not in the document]
will-mysten: Yeah let me make this change before we publish any documentation. I'd prefer to get ahead of this.

packages/docs/content/payment-kit/payment-kit-sdk.mdx
```diff
@@ -0,0 +1,614 @@
...
+
+await client.signAndExecuteTransaction({
+	transaction: tx,
+	signer: keypair,
+});
+```
+
+### Other Call Methods
+
+All transaction methods have corresponding call methods:
```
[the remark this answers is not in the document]
will-mysten: No strong opinions on my end. I can invert this!

packages/docs/content/payment-kit/payment-processing.mdx
```diff
@@ -0,0 +1,409 @@
...
+		registryName: DEFAULT_REGISTRY_NAME,
+	});
+
+	await client.signAndExecuteTransaction({
+		transaction: tx,
+		signer: keypair,
+	});
+} catch (error) {
+	// Check if this is a duplicate payment error
+	if (error.message.includes('already exists') || error.message.includes('duplicate')) {
```
hayes-mysten: I don't have a better suggestion, but this sucks.  This may stop working at any time, and is probably likely to break when switching to gql or grpc.

I'd rather just not document it than tell people to do this

packages/docs/content/payment-kit/payment-processing.mdx
```diff
@@ -0,0 +1,409 @@
...
+
+### Ephemeral Payments
+
+Ephemeral payments have lower gas costs because they:
+
+- Only transfer coins
+- Don't create persistent storage
+- Skip duplicate checking
+
+Approximate gas cost: ~0.5-1M MIST (varies by network conditions)
```
hayes-mysten: We could make this relative to registry payments (eg roughly half) rather than having an explicit amount

PR #613, opened 2025-10-16
Exposed `getRegistryIdFromName` and properly defaulting `registryName`

will-mysten:
## Description

- `getRegistryIdFromName` is now exposed via `PaymentKitClient`
- `registryName` is now properly defaulted to the default payment registry
---

### AI Assistance Notice

- [X] I did not use AI for this PR.

### DIFF
--- a/packages/payment-kit/src/utils.ts
+++ b/packages/payment-kit/src/utils.ts
@@ -3,7 +3,7 @@
 
 import { bcs } from '@mysten/sui/bcs';
 import { deriveObjectID } from '@mysten/sui/utils';
-import { DEFAULT_REGISTRY_NAME } from './constants';
+import { DEFAULT_REGISTRY_NAME } from './constants.js';
 
 export const getRegistryIdFromName = (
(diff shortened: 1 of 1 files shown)


packages/payment-kit/src/client.ts
```diff
@@ -107,4 +107,8 @@ export class PaymentKitClient {
 			epochAtTimeOfRecord: decoded.epoch_at_time_of_record,
 		};
 	}
+
+	getRegistryIdFromName(registryName: string): string {
```
will-mysten: @hayes preference if we expose this here? Or should we add a `@mysten/payment-kit/utils` package?

hayes-mysten: this is fine.  I wouldn't do a subpath export, but you could just export it from the main index file

hayes-mysten (approved):

packages/payment-kit/src/client.ts
will-mysten: Exposed it via the client, since it needs the `packageConfig` for `namespaceId`

PR #615, opened 2025-10-17
[bcs] improve perf of bcs byteVector

hayes-mysten:
## Description

## Test plan

---

### AI Assistance Notice


packages/bcs/src/writer.ts
```diff
@@ -53,7 +53,10 @@ export class BcsWriter {
 	private ensureSizeOrGrow(bytes: number) {
 		const requiredSize = this.bytePosition + bytes;
 		if (requiredSize > this.size) {
-			const nextSize = Math.min(this.maxSize, this.size + this.allocateSize);
+			const nextSize = Math.min(
+				this.maxSize,
+				Math.max(this.size + requiredSize, this.size + this.allocateSize),
```
Jordan-Mysten: I wonder if we could use arraybuffer and feature detect the new native resizing: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/resize

Also, we might want to change this to be an exponential (capped) resizer so we do fewer incremental allocations.

hayes-mysten: gonna punt both of these for now, not a bad idea, but I think require some more work and I have other things to work on that are more urgent

PR #668, opened 2025-11-11
fix(codegen): Generate modules with only private entry functions

fuyutarow:
## Description

Fix module generation for entry-only functions when using privateMethods: 'entry'

This PR fixes a bug where modules containing only private entry functions (with no types or public functions) were not being generated, even when privateMethods was set to 'entry'.

The issue was in the hasFunctions() method, which only checked for public functions (visibility === 'Public'), ignoring the #includedFunctions set that was already populated by includeAllFunctions() based on the privateMethods configuration.

**Changes:**
- Updated hasFunctions() to check this.#includedFunctions.size > 0 instead of only checking for public functions
- This ensures consistency between includeAllFunctions() and hasFunctions(), making the privateMethods option work as documented

**Impact:**
- Modules with only private entry functions (like access_policy.move) will now be generated correctly when privateMethods: 'entry' (the default)
- No breaking changes - only fixes previously broken behavior

## Test plan
- Existing unit tests pass (pnpm --filter @mysten/codegen test)
- Build succeeds (pnpm turbo build --filter=@mysten/codegen)
- Code generation works correctly (pnpm --filter @mysten/codegen codegen)
- Lint and format checks pass

---

### AI Assistance Notice

### DIFF
--- a/.changeset/fix-entry-functions-generation.md
+++ b/.changeset/fix-entry-functions-generation.md
@@ -1,0 +1,7 @@
+---
+"@mysten/codegen": patch
+---
+
+Fix hasFunctions() to respect privateMethods configuration
+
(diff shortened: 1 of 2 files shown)


hayes-mysten (approved):
thanks!

PR #708, opened 2025-12-06
fix(enoki): open window url should be `about:blank` by default

### LINKED ISSUE #707: [@mysten/enoki] Popup is not opened correctly in some browsers
Hello,

With my team, we are using `@mysten/enoki` as part of our solution in production, for the user to be able to sign in with his Google account.

The entire flow works on desktop and on most of browsers.

However, we found out Brave on iOS (https://apps.apple.com/us/app/brave-browser-search-engine/id1052879175) is blocking the opening of the popup to finalize the login process. We did not test all browsers but we think it can occurs on some others as well.

After checking out the source code and with some tests, we found out that we need to specify `about:blank` instead of `undefined` to open the popup, otherwise the window we try to open is indeed blocked in certain cases.
The phenomenon is quite sporadic and can not occurred even in Brave on iOS.

Also, please note that even when trying to open a window with a specific URL, it will first load an `about:blank` page before loading the targeted URL.

LucasRakotomalala:
## Description

In `@mysten/enoki`, the popup window was open with `undefined` as first parameter for URL, which leads to not opening the window in certain browsers with some stricter policies.

Opening with `about:blank` fixes the issue.

Close #707 

## Test plan

I tested the flow with Google Chrome (Desktop and iOS), Safari (Desktop and iOS), Mozilla (Desktop and iOS) as well as Brave (Desktop and iOS) and the flow works for every of those web browsers now.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/dry-games-yell.md
+++ b/.changeset/dry-games-yell.md
@@ -1,0 +1,5 @@
+---
+'@mysten/enoki': patch
+---
+
+open window with `about:blank` instead of `undefined`


LucasRakotomalala:
Sure, let me create one :)

PR #710, opened 2025-12-06
feat(enoki): support `additionalEpochs` as part of configuration

### LINKED ISSUE #709: [@mysten/enoki] `additionalEpochs` for zkLogin nonce is never specified
Hello,

With my team, we are using `@mysten/enoki` as part of our solution in production, for the user to be able to sign in with his Google account.

The entire flow works on desktop and on most of browsers.

However, we found out that session are only working for two days.
We checked the documentation and found that two days is the default value but it can be specified as part of the body of the creation of zkLogin nonce.

After checking out the source code and with some tests, we found out that `input.additionalEpochs` is never assigned when calling `EnokiClient.createZkLoginNonce` (https://github.com/MystenLabs/ts-sdks/blob/main/packages/enoki/src/wallet/wallet.ts#L424-L427 and https://github.com/MystenLabs/ts-sdks/blob/main/packages/enoki/src/EnokiFlow.ts#L149-L152)

_Note: On our side we are not using the deprecated `EnokiFlow`._

LucasRakotomalala:
## Description

In `@mysten/enoki`, a call to Enoki API is made to create a zkLogin nonce.
`additionalEpochs` is property that can be changed but so far it is not possible to do so.

This PR adds `additionalEpochs` parameter as part of `registerEnokiWallets` function as well as `EnokiFlow` deprecated class.

Close #709 

## Test plan

I ran the demo application and made sure the call made to create a zkLogin nonce is done with the `additionalEpochs` passed as parameter of `registerEnokiWallets`.
I also made sure that without specifying anything, the call is done without `additionalEpochs` for compatibility reasons.

---

### AI Assistance Notice

### DIFF
--- a/packages/enoki/src/EnokiClient/index.ts
+++ b/packages/enoki/src/EnokiClient/index.ts
@@ -104,13 +104,13 @@ export class EnokiClient {
 		});
 	}
 
-	createZkLoginNonce(input: Omit<CreateZkLoginNonceApiInput, 'additionalEpochs'>) {
+	createZkLoginNonce(input: CreateZkLoginNonceApiInput) {
 		return this.#fetch<CreateZkLoginNonceApiResponse>('zklogin/nonce', {
 			method: 'POST',
 			body: JSON.stringify({
 				network: input.network,
 				ephemeralPublicKey: input.ephemeralPublicKey.toSuiPublicKey(),
-				additionalEpochs: this.#additionalEpochs,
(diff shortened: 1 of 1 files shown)


LucasRakotomalala:
Hello,
I am not sure about the implementation.

Currently it is global for all providers. However, I can change the implementation if you want it to be per provider.

Kind regards,
Lucas.

PR #763, opened 2026-01-03
fix(dapp-kit): resolve type declaration resolution issues

Danny-Devs:
## Description

Fixes TypeScript type resolution issues in `@mysten/dapp-kit-core` and `@mysten/dapp-kit-react` that prevent consumers from properly importing types when using `moduleResolution: "bundler"`.

### Root Cause

The built output files don't match the paths declared in `package.json`, causing TypeScript to fail type resolution.

---

### 1. `dapp-kit-core/tsdown.config.ts` — Add `hash: false`

```ts
+ hash: false,
```

**Why:** tsdown defaults to hashing declaration filenames (e.g., `index-Dy4LAlgj.d.ts`). This breaks type resolution because `package.json` exports reference `./dist/index.d.ts`, not the hashed filename. Setting `hash: false` ensures predictable output paths that match the declared exports.

---

### 2. `dapp-kit-core/package.json` — Fix file extensions

```diff
- "types": "./dist/index.d.mts"
+ "types": "./dist/index.d.ts"

- "import": "./dist/index.mjs"
+ "import": "./dist/index.js"
```

**Why:** The package.json referenced `.mjs` / `.d.mts` files, but tsdown outputs `.js` / `.d.ts`. This mismatch caused "module not found" errors. Since `"type": "module"` is set, `.js` files are already treated as ESM — the `.mjs` extension is unnecessary.

---

### 3. `dapp-kit-core/package.json` — Add explicit `types` condition to exports

```diff
  ".": {
+   "types": "./dist/index.d.ts",
    "import": "./dist/index.js",
    "require": "./dist/index.cjs"
  }
```

**Why:** TypeScript's `moduleResolution: "bundler"` reads the `exports` field directly. Without an explicit `types` condition, TypeScript doesn't know where to find type declarations for subpath imports. The `types` field must come **first** per the [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/modules/reference.html) and [publint rules](https://publint.dev/rules), as conditions are matched in order.

Same fix applied to the `"./web"` subpath.

---

### 4. `dapp-kit-react/tsdown.config.ts` — Add `hash: false` + `dts.resolve`

```diff
- dts: true,
+ dts: {
+   resolve: true,
+ },
  // ...
+ hash: false,
```

**Why:**
- `hash: false` — same reason as dapp-kit-core (prevents hashed declaration filenames)
- `dts: { resolve: true }` — bundles type declarations from dependencies into the output, ensuring the `.d.ts` files are self-contained without broken import paths

---

### 5. `dapp-kit-react/package.json` — Add `types` condition

```diff
  ".": {
+   "types": "./dist/index.d.ts",
    "import": "./dist/index.mjs",
    "require": "./dist/index.cjs"
  }
```

**Why:** Same as dapp-kit-core — enables TypeScript to resolve types when using `moduleResolution: "bundler"`.

---

### 6. `dapp-kit-react/src/index.ts` — Export hook option types

```diff
  export { useWallets } from './hooks/useWallets.js';
+ export type { UseWalletsOptions } from './hooks/useWallets.js';
```

**Why:** The hook options types (`UseWalletsOptions`, `UseConnectionOptions`, etc.) were defined but not exported. Consumers couldn't type their own abstractions around these hooks. Now all option types are publicly available.

---

## Test Plan

1. Built both packages locally with `pnpm build` — verified clean output
2. Confirmed output declaration files are not hashed (`ls dist/` shows `index.d.ts`, not `index-[hash].d.ts`)
3. Verified output files match package.json export paths
4. Tested type resolution in a consuming project using `moduleResolution: "bundler"`
5. Verified IntelliSense works for exported hook option types

---

### AI Assistance Notice

### DIFF
--- a/.changeset/fix-dapp-kit-type-resolution.md
+++ b/.changeset/fix-dapp-kit-type-resolution.md
@@ -1,0 +1,11 @@
+---
+"@mysten/dapp-kit-core": patch
+"@mysten/dapp-kit-react": patch
+---
+
+Fix type declaration resolution issues in dapp-kit packages
+
+- Add explicit `types` field to package.json exports for proper TypeScript module resolution
+- Set `hash: false` in tsdown config to prevent hashed output filenames that break type resolution
+- Export hook options types from dapp-kit-react for better DX
+- Fix file extensions in dapp-kit-core exports (`.mjs` → `.js`)


Danny-Devs:
Thanks for the heads up\! I'll go ahead and close this out since #626 covers it. Looking forward to v2\!

Issue #757, opened 2025-12-23
useLayoutEffect does nothing on the server during build with SSG


synasapmob:
## What problem
I’m using [vite-ssg](https://www.npmjs.com/package/vite-ssg). When building the packages, I see a console warning like: “useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server…”

I then tried building with static HTML, and the results were good. So I think this warning happens because using useLayoutEffect during a server-side build is not appropriate.

## Proposal
Would it be possible to replace useLayoutEffect with useEffect, or conditionally use useLayoutEffect only on the client side? This could help avoid the warning and improve compatibility with SSR/SSG builds.

> useLayoutEffect:
<img width="3314" height="468" alt="Image" src="https://github.com/user-attachments/assets/b544132c-ae29-4d51-96a7-f42f5593308c" />

> useEffect:
<img width="3250" height="1130" alt="Image" src="https://github.com/user-attachments/assets/99300857-bc78-439c-bc2e-aee716861109" />

<img width="1318" height="472" alt="Image" src="https://github.com/user-attachments/assets/b43ae3d8-3654-49c5-842b-b02f9609136e" />

PR #854, opened 2026-02-09
use isomorphic layout effect in wallet hooks

synasapmob:
## Description

This PR is created to address issue #757. The problem is that `useLayoutEffect` does not work during build time.
I’m following the official error documentation:

- https://reactjs.org/link/uselayouteffect-ssr
- https://github.com/facebook/react/issues/14927#issuecomment-466815232

Resonable:
> As [gaearon](https://gist.github.com/gaearon) mentioned, useLayoutEffect does not run on the server, and it also does not produce a warning.
some applications enable Strict Mode, which may cause errors during build time. To avoid this case, we should check the environment before using it. The idea is to use `useLayoutEffect` in the browser and `useEffect` during runtime. This is a solved problem.

## Test plan

- Reproduced the original build error in an SSG environment where `useLayoutEffect` runs during build time.
- Verified that the build fails before this change.
- Applied the isomorphic effect (`useLayoutEffect` in the browser, `useEffect` on the server).
- Confirmed that the SSG build now completes successfully without warnings or errors.
- Verified that runtime behavior in the browser remains unchanged.

Screenshot 2026-02-10 at 09 56 29

---

### AI Assistance Notice

- [x] I used AI to understand the problem space / repository.

### DIFF
--- a/.changeset/slow-pianos-read.md
+++ b/.changeset/slow-pianos-read.md
@@ -1,0 +1,5 @@
+---
+'@mysten/dapp-kit': patch
+---
+
+Implement an isomorphic effect to fix SSG build errors


hayes-mysten:
Looks like you replaced the PR template, you will need to use the correct PR template for the description for CI to pass

PR #947, opened 2026-03-15
Test: Introduce PAS SDK

hayes-mysten:
Test PR for review evaluation


packages/pas/src/constants.ts
```diff
@@ -0,0 +1,15 @@
...
+export const TESTNET_PAS_PACKAGE_CONFIG: PASPackageConfig = {
+	packageId: '0x7e77592474ea9759b46ca5c6515ed4840952cd80d60ee54a1614382811d46730',
+	namespaceId: '0xf7c77ac8bbbdf47f7b1cf8ac8aa489cfc4dff25847f3e2e1db53bde5c454be2b',
+};
+
+// TODO: also update `client.ts` when adding these variables
+export const MAINNET_PAS_PACKAGE_CONFIG: PASPackageConfig = {
+	packageId: '', // TODO: Replace with actual mainnet package ID
+	namespaceId: '', // TODO: Replace with actual mainnet namespace ID
+};
```
hayes-mysten: **Major**: This config is exported publicly with empty strings. A user can import it directly and pass it as `packageConfig` to `pas()`, bypassing the `PASClient` constructor guard. `deriveObjectID` will then produce valid-looking but wrong addresses from the empty `namespaceId`.

Consider one of:
- Don't export until populated
- Add runtime validation in derivation functions: `if (!namespaceId) throw new PASClientError('...')`
- Export as `undefined` with a type guard

packages/pas/src/intents.ts
```diff
@@ -0,0 +1,839 @@
...
+	const accountRequests = new Map<string, AccountOwner>();
+	const intentDataList: PASIntentData[] = [];
+	let cfg: PASPackageConfig | null = null;
+
+	for (const command of commands) {
+		if (command.$kind !== '$Intent' || command.$Intent.name !== PAS_INTENT_NAME) continue;
+		const data = command.$Intent.data as unknown as PASIntentData;
+
+		if (!cfg) cfg = data.cfg;
+		intentDataList.push(data);
```
hayes-mysten: **Bug: Mixed PAS configs silently resolved against first config.** `collectIntentData()` captures `cfg` from the first PAS intent only. If a PTB has multiple PAS intents with different package configs, subsequent intents silently use the wrong config — wrong derived IDs, wrong Move package. Hard-fail if configs differ.

packages/pas/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Options for creating a PAS client plugin
+ */
+export interface PASOptions<Name extends string = 'pas'> {
+	/** The name of the plugin */
+	name?: Name;
+	/** The package configuration */
+	packageConfig?: PASPackageConfig;
+	/** The network to use */
+	network?: 'mainnet' | 'testnet';
+}
```
hayes-mysten: **Dead API:** `network` is exposed in `PASOptions` but never read by `pas()` — it only uses `config.suiClient.network`. Remove from public type or wire it up.

packages/pas/src/intents.ts
```diff
@@ -0,0 +1,839 @@
...
+	const allIds = [...objectIds];
+	const { objects: fetched } = await client.core.getObjects({
+		objectIds: allIds,
+		include: { content: true },
+	});
+
+	const objects = new Map<string, SuiObject | null>();
+
+	for (const id of allIds) {
+		const obj = fetched.filter((o) => 'content' in o).find((o) => o.objectId === id);
```
hayes-mysten: **Major: Error results from `getObjects` are silently treated as "not found"**

`getObjects` returns `(Object | Error)[]` — an `Error` entry means the fetch failed for that ID, **not** that the object doesn't exist. The `'content' in o` filter skips both `Error` instances and genuinely-missing objects, mapping them both to `null`.

Consequences:
- **Accounts**: An existing account that fails to fetch is treated as non-existent → the resolver emits `account::create`, which fails on-chain (duplicate creation) with an opaque Move error.
- **Policies**: A transient fetch failure for a policy → false `PolicyNotFoundError`, misleading the user into thinking the issuer hasn't configured the asset.

Also: this is O(n²) — for each ID, it rescans the full array with `filter().find()`. Build a `Map` once by index instead.

Suggested fix: iterate `fetched` by index, check `instanceof Error`, throw or propagate accordingly.

packages/pas/src/resolution.ts
```diff
@@ -0,0 +1,285 @@
...
+	/** The sender account argument (already resolved) */
+	senderAccount?: Argument;
+	/** The receiver account argument (already resolved) */
+	receiverAccount?: Argument;
+	/** The policy argument (already resolved) */
+	policy?: Argument;
+	/** The request argument (already resolved) */
+	request?: Argument;
+	/** The system type T (e.g., "0x2::sui::SUI") */
+	systemType?: string;
```
hayes-mysten: **Minor: `systemType` field is declared but never read**

This field is populated by both callers (`intents.ts:449`, `intents.ts:542`) but never accessed in `buildMoveCallCommandFromTemplate` or `resolveRawPasRequest`. Dead code — either remove it or wire it in if intended.

packages/pas/src/resolution.ts
```diff
@@ -0,0 +1,285 @@
...
+ * @param args - The resolved arguments and addInput helper
+ * @returns A Command object ready for `replaceCommand`
+ */
+export function buildMoveCallCommandFromTemplate(
+	command: ParsedTemplateCommand,
+	args: RawCommandBuildArgs,
+): SdkCommand {
+	const resolvedArgs: Argument[] = [];
+
+	for (const arg of command.arguments) {
```
hayes-mysten: **Minor: `FundsWithdrawal` CallArg variant unhandled**

The on-chain `CallArg` enum (`ptb.ts:86-93`) includes a `FundsWithdrawal` variant. The argument resolution loop handles `Ext`, `GasCoin`, `NestedResult`, `Result`, and `Input`, but a `FundsWithdrawal` input falls through silently (no throw, no push). Add an explicit throw for unsupported variants.

packages/pas/pnpm-lock.yaml
```diff
@@ -0,0 +1,1209 @@
+lockfileVersion: '9.0'
```
hayes-mysten: **Minor: Package-level lockfile in a workspace member**

This 1209-line `pnpm-lock.yaml` inside a workspace package is unusual — the monorepo root lockfile should be the single source of truth. This will drift and shouldn't be committed.

.github/workflows/pas-ci.yml
```diff
@@ -0,0 +1,37 @@
+name: PAS tests (unit & e2e)
+on:
+  pull_request:
```
hayes-mysten: **Minor: CI workflow triggers on all PRs without path filtering**

This runs on every PR/push regardless of which files changed. Consider adding path filters (`packages/pas/**`, `pnpm-lock.yaml`) to avoid wasting CI on unrelated changes.

hayes-mysten (commented):
## Review: 1 major finding, 7 minor

**Verdict: Comment** (in addition to the existing 15 findings from previous reviews)

The intent-resolution architecture is well-designed — memoized closures, batch pre-fetching, and deduplication work cleanly together. The major new finding is a silent error-swallowing path in the object fetcher that can cause incorrect behavior under transient RPC failures.

### Minor items not attached inline:
- **Empty `.gitignore`**: `packages/pas/.gitignore` has 0 patterns. Either populate it or remove it.
- **Stray `vite.config.ts`**: Imports `@vitejs/plugin-react` which isn't in devDependencies. A PAS SDK doesn't need a React Vite plugin — leftover scaffolding.

packages/pas/src/intents.ts
```diff
@@ -0,0 +1,839 @@
...
+		objects.set(id, obj ?? null);
+	}
+
+	// 2. Build initial account map (existing vs needs-creation)
+	const accounts = new Map<string, AccountState>();
+	for (const [accountId] of accountRequests) {
```
hayes-mysten: **major: silently drops errored objects → false negatives**

`.filter((o) => 'content' in o)` discards any object that came back as an error (e.g. transient RPC failure). The errored object then maps to `null` in the `objects` Map, which downstream logic interprets as "doesn't exist". This can cause:
- Spurious `account::create` for accounts that do exist
- False `PolicyNotFoundError` for policies that do exist

Consider checking for `instanceof Error` explicitly and throwing, rather than silently dropping.

packages/pas/src/intents.ts
```diff
@@ -0,0 +1,839 @@
...
+			action: 'sendBalance',
+			from,
+			to,
+			amount: String(amount),
+			assetType,
+			cfg: packageConfig,
+		});
+}
+
+export function unlockBalanceIntent(
```
hayes-mysten: **major: `number` amounts silently lose precision**

`String(amount)` on an unsafe JS number (e.g. `2**53 + 1`) rounds before the string conversion, so `BigInt(amount)` downstream gets a wrong value. Token amounts in production can exceed 2^53.

Consider accepting only `bigint | string`, or guarding with `Number.isSafeInteger()` and rejecting unsafe values.

packages/pas/src/derivation.ts
```diff
@@ -0,0 +1,107 @@
...
+	const policyKeyBcs = new Uint8Array([0]);
+
+	const wrap = options?.wrapType ?? DEFAULT_WRAP_TYPE;
+	const typeTag = `${packageId}::keys::PolicyKey<${wrap(assetType)}>`;
+	return deriveObjectID(namespaceId, typeTag, policyKeyBcs);
+}
+
+/**
+ * Derives the templates object address for a given package configuration.
+ *
```
hayes-mysten: **major: `assetType` not normalized before hashing**

`derivePolicyAddress` hashes the raw `assetType` string (e.g. `0x2::sui::SUI`), but `buildSendBalance` passes `normalizeStructTag(assetType)` to the Move call (e.g. `0x0000...0002::sui::SUI`). If a user passes the short form, the derived policy ID won't match the on-chain object.

Call `normalizeStructTag()` on `assetType` inside this function before building the type tag.

packages/pas/src/intents.ts
```diff
@@ -0,0 +1,839 @@
...
+		if (command.$kind !== '$Intent' || command.$Intent.name !== PAS_INTENT_NAME) continue;
+		const data = command.$Intent.data as unknown as PASIntentData;
+
+		if (!cfg) cfg = data.cfg;
+		intentDataList.push(data);
+
+		switch (data.action) {
+			case 'sendBalance': {
+				const fromId = deriveAccountAddress(data.from, cfg);
+				const toId = deriveAccountAddress(data.to, cfg);
```
hayes-mysten: **major: config captured from first intent only**

If a PTB contains multiple PAS intents with different `cfg` values (e.g. different package deployments), subsequent intents silently use the first intent's config — wrong derived IDs, wrong Move package.

Consider validating that all intents share the same `cfg`, or hard-failing if they differ.

packages/pas/src/types.ts
hayes-mysten: **major: dead API**

`network` is accepted here but `pas()` never reads it — it only uses `config.suiClient.network`. This is a misleading public API surface. Either wire it into config resolution or remove it.

packages/pas/src/resolution.ts
```diff
@@ -0,0 +1,285 @@
...
+	policy?: Argument;
+	/** The request argument (already resolved) */
+	request?: Argument;
+	/** The system type T (e.g., "0x2::sui::SUI") */
+	systemType?: string;
+}
+
+/**
+ * Builds a `Command` (TransactionCommands.MoveCall) from a parsed template command,
+ * suitable for use with `transactionData.replaceCommand()`.
```
hayes-mysten: **minor: dead field**

`systemType` is defined on `RawCommandBuildArgs` and passed by callers (`intents.ts:455,549`) but `buildMoveCallCommandFromTemplate` never reads it. Remove or wire it up.

PR #984, opened 2026-03-31
Fix MVR type resolution to handle vector and primitive type parameters

Danny-Devs:
## Description

`extractMvrTypes` and `replaceMvrNames` call `parseStructTag` on all string type parameters, but type parameters from `parseTypeTag` can be vectors (e.g., `"vector<@mvr/demo::baz::Qux>"`) or primitives (e.g., `"u8"`), neither of which are valid struct tags.

This is the same root cause as #983 — functions calling `parseStructTag` on strings that aren't struct tags.

**Bug 1 — vectors produce corrupted MVR type names:**
```ts
extractMvrTypes('0x2::foo::Bar<vector<@mvr/demo::baz::Qux>>')
// Expected: Set(['@mvr/demo::baz::Qux'])
// Actual:   Set(['0x00000000000000000000000000000000000000000000000vector<@mvr/demo::baz::Qux>'])
```

**Bug 2 — primitives crash `replaceMvrNames`:**
When a struct tag has a primitive type parameter (e.g., `u8`), the recursive call `replaceMvrNames("u8", cache)` calls `parseStructTag("u8")` which throws `"Invalid struct tag: u8"`.

**Fix:** Vector types are now unwrapped and recursed into. Primitives are passed through unchanged in `replaceMvrNames`. Both guards use the same pattern: check before calling `parseStructTag`.

## Test plan

- Created new test file `test/unit/client/mvr.test.ts` with 9 test cases for `extractMvrTypes`:
  - MVR type in struct tag (baseline)
  - MVR type in struct tag type parameter
  - MVR type inside vector type parameter (Bug 1 fix)
  - MVR type inside nested vectors
  - MVR type inside vector with parameterized inner type
  - MVR type in top-level vector
  - Mixed MVR + primitive type parameters
  - Non-MVR vector types (no false positives)
  - Primitive type parameters (no false positives)
- All 9 tests pass
- `pnpm turbo build --filter=@mysten/sui` clean
- `pnpm lint` clean

---

### AI Assistance Notice

Written with [Claude Code](https://claude.ai/code) (claude-opus-4-6). Bug originally identified during PR #827 validation work; fix implementation and tests were AI-assisted with human review and validation.

### DIFF
--- a/.changeset/fix-mvr-vector-type-params.md
+++ b/.changeset/fix-mvr-vector-type-params.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': patch
+---
+
+Fix `extractMvrTypes` and `replaceMvrNames` to handle vector and primitive type parameters. Previously, these functions passed all string type parameters directly to `parseStructTag`, which produced corrupted results for vector types (e.g., `vector<@mvr/demo::baz::Qux>`) and threw on primitives (e.g., `u8`). Vector types are now unwrapped and recursed into, and primitive types are passed through unchanged.
--- a/packages/sui/src/client/mvr.ts
+++ b/packages/sui/src/client/mvr.ts
@@ -332,6 +332,10 @@ function validateOverrides(overrides?: {
 export function extractMvrTypes(type: string | StructTag, types = new Set<string>()) {
 	if (typeof type === 'string' && !hasMvrName(type)) return types;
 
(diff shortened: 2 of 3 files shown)


hayes-mysten (approved):
the looks good to, thanks for the fix!

Danny-Devs:
Circling back on this — lint fix was pushed, just rebased onto main.

PR #988, opened 2026-04-01
refactor(client): unify ObjectError across all transports

Danny-Devs:
> **Rearchitected from the previous version.** Original approach added a transport-specific `SuiGrpcRequestError`. This version takes the opposite direction — makes the existing `ObjectError` transport-agnostic so `instanceof ObjectError` works consistently across JSON-RPC, GraphQL, and gRPC.

## Description

`ObjectError` was coupled to JSON-RPC — it imported `ObjectResponseError` from the jsonRpc layer and its `fromResponse()` factory switched on JSON-RPC-specific codes. gRPC didn't use it at all, throwing plain `Error` instead. GraphQL used it but passed a baked-in message string.

This PR makes `ObjectError` transport-agnostic and satisfies each of the five principles you laid out in review:

1. **Consistent errors across RPCs.** `ObjectErrorCode` is the transport-agnostic union `'notFound' | 'deleted' | 'unknown'`. JSON-RPC distinguishes `deleted` at the wire level, so it maps `deleted` → `'deleted'` and `notExists` / `dynamicFieldNotFound` → `'notFound'`; `displayError` / `unknown` surface as `'unknown'`. gRPC can't distinguish deleted from never-existed (its `NOT_FOUND` is not specific), so it maps `google.rpc.Code.NOT_FOUND` (5) → `'notFound'` and everything else → `'unknown'`. GraphQL omits absent objects without saying why, so it also collapses to `'notFound'`. The `'deleted'` code is only surfaced where the wire genuinely distinguishes it; the other two transports stay honest.
2. **Consumers don't need to understand which RPC is in use.** `error.code` and `error.objectId` mean the same thing on all three transports. No transport-specific codes leak through the unified surface.
3. **`instanceof` works consistently across all three clients.** Single `ObjectError` class in `client/errors.ts`; gRPC and GraphQL now emit `ObjectError` instead of plain `Error`. `GraphQLResponseError` now extends `SuiClientError` (previously extended `Error`), and the GraphQL multi-error path wraps the aggregate in `SuiClientError` with `transportDetails: { $kind: 'graphql' }` on the cause — so `instanceof SuiClientError` is genuinely universal across all three transports, not just for object errors. `GetObjectsResponse.objects` is narrowed from `(Object | Error)[]` to `(Object | ObjectError)[]`; existing `instanceof Error` checks continue to work unchanged because `ObjectError extends Error`.
4. **Decisions based on error details are not tied to an RPC implementation.** `ObjectError` lives in `client/errors.ts` with zero dependencies on any transport layer. Consumers act on `error.code` (three values, universal).
5. **RPC-specific detail remains accessible.** Adds `TransportDetails`, a tagged union `{ $kind: 'jsonRpc' | 'grpc' | 'graphql' }` on the `SuiClientError` base class. Consumers that need the raw JSON-RPC response or the `google.rpc.Status` narrow on `error.transportDetails?.$kind` and get typed access to the wire payload.

### Implementation notes

- **`ObjectError.objectId` is non-empty by contract.** In the rare JSON-RPC case where `listOwnedObjects` returns an error that identifies no specific object (e.g. a `displayError` with no `object_id`), we escalate to the base `SuiClientError` rather than fabricating a sentinel id. Consumers who catch `SuiClientError` still catch everything; `transportDetails` carries the raw wire payload. (Chose this over `objectId: string | null` or `''` sentinels because an empty-string sentinel silently breaks `if (error.objectId)` checks, and a nullable field forces every consumer to re-handle the case.)
- **`ObjectError`'s `options` arg is optional**, and the base-class `transportDetails` field is honestly optional (no `declare readonly transportDetails: TransportDetails` narrowing on the `ObjectError` subclass). Consumers can construct `ObjectError` directly without ceremony; the SDK still attaches `transportDetails` at every real construction site.
- **`AggregateObjectError extends SuiClientError`**, carrying all errors on `.errors`. `coreClientResolveTransactionPlugin` now throws the bare `ObjectError` when exactly one input is invalid (narrowing preserved) and `AggregateObjectError` when more than one is invalid — restoring multi-error reporting that was previously flattened to a concatenated-message plain `Error`. Consumers that catch `SuiClientError` catch both cases uniformly.
- **Two exhaustive mapper helpers in `jsonRpc/core.ts`** (`mapJsonRpcObjectErrorCode` + `extractObjectIdFromResponseError`) use the `satisfies never` house pattern from `client/utils.ts` — if the upstream wire protocol adds a new `ObjectResponseError` code, the helpers stop compiling.
- **gRPC unexpected-oneof case now throws `SuiClientError`.** Previously `grpc/core.ts` pushed `new Error(...)` into the response array when the oneof was neither `'object'` nor `'error'`. That case represents a programmer error / proto drift, not per-object data, so it now throws `SuiClientError` — still catchable via `catch (e) { if (e instanceof SuiClientError) ... }`, but no longer surfaced as data.
- **`instanceof` checks in the composed methods now use `ObjectError`.** The narrowed `(Object<Include> | ObjectError)[]` slot type lets `CoreClient.getObject`, `CoreClient.getDynamicField`, and `core-resolver.ts` use `instanceof ObjectError` instead of `instanceof Error` — the tighter check the narrowed type earns.
- **GraphQL preserves the raw input id at the `ObjectError` construction site.** Previously `graphql/core.ts` pre-normalized the input ids before constructing `ObjectError`, so a consumer passing `'0x9999'` saw `error.objectId === '0x000...9999'` on GraphQL but `error.objectId === '0x9999'` on JSON-RPC and gRPC. Now GraphQL normalizes only for the server-side lookup and preserves `rawId` for the error, matching the other two transports.
- **`SuiClientError`, `ObjectError`, `AggregateObjectError`, `ObjectErrorCode`, and `TransportDetails` are exported** from `@mysten/sui/client`.

## Test plan

- `test/unit/client/object-error.test.ts` — unit tests covering:
  - **`ObjectError` shape**: `instanceof` chain, field access, canonical message generation for all three codes (`'notFound'` / `'deleted'` / `'unknown'`), `cause` preservation, optional-options construction, `TransportDetails` for all three `$kind` variants, `SimulationError` distinguishability.
  - **`AggregateObjectError`**: `instanceof SuiClientError`, exposes `.errors`, summarizes child ids in the aggregate message.
  - **`SuiClientError` base**: `transportDetails` on the base class, `cause` preserved independently of `transportDetails`.
  - **`JSONRpcCoreClient.listOwnedObjects` escalation**: `displayError` / `unknown` wire codes (no wire-level object id) escalate to bare `SuiClientError` so `ObjectError.objectId` stays non-empty; `notExists` / `deleted` / `dynamicFieldNotFound` all produce `ObjectError` with the correct mapped code and extracted id; plus a universal `catch (SuiClientError)` contract test pinning compatibility across all five wire codes.
  - **`JSONRpcCoreClient.getObjects` error mapping**: 5-row `it.each` exercising every `ObjectResponseError` wire code (including `deleted` → `'deleted'`); reference-identity test proving `transportDetails.response` round-trips the raw wire payload without copy or transformation.
  - **`GrpcCoreClient.getObjects` error mapping**: 4-row `it.each` exercising `google.rpc.Code.NOT_FOUND` (5) and three non-NOT_FOUND codes (INTERNAL=13, PERMISSION_DENIED=7, OK=0) to pin both branches of the gRPC status-code mapping.
  - **`GraphQLCoreClient.getObjects` error mapping**: regression tests for cross-transport `objectId` parity — `ObjectError.objectId` preserves the raw input id for both an unnormalized (`'0x9999'`) and a pre-normalized input.
- `test/unit/client/core-resolver.test.ts` —
  - **`coreClientResolveTransactionPlugin` single-error path**: when exactly one `tx.object(id)` input is unresolved, `tx.build()` rethrows the original `ObjectError` instance (reference identity) so `instanceof ObjectError` / `error.code` / `error.objectId` / `error.transportDetails` work consistently through the transaction resolver path.
  - **`coreClientResolveTransactionPlugin` multi-error path**: when more than one input is invalid, `tx.build()` throws `AggregateObjectError` with every child `ObjectError` preserved by reference on `.errors`.
  - **GraphQL universal catch**: a schema error from `handleGraphQLErrors` is catchable via `instanceof SuiClientError` and carries `transportDetails: { $kind: 'graphql' }`.
- `test/e2e/clients/core/objects.test.ts` — two new `testWithAllClients` cases enforce cross-transport parity: `getObjects` returns `ObjectError` with the correct `code`, `objectId`, and `transportDetails.$kind` on all three clients, and `getObject` throws `ObjectError` on all three clients.
- `pnpm --filter @mysten/sui test` — **380 passing**, 0 failing (typecheck + unit).
- `pnpm --filter @mysten/sui lint` — 0 warnings, 0 errors; prettier clean.

---

### AI Assistance Notice

- [x] I used AI to understand the problem space / repository.


hayes-mysten (commented):
This looks pretty good in theory.

The plan here was for all 3 core client to emit ObjectError instances, but that somehow got missed as the other core clients got built out.

Generally speaking all of the core clients should have identical behavior.  I think the correct thing to do here would be to have the grpc and graphql clients use (or extend) the existing `ObjectError`.

It unfortunate that this currently has a different concept of code that does not match up with what we do in grpc.

I am a little hesitant to merge this as is, because it only complicates the existing inconsistency.  I think this is better than what we have on the json rpc side, but I want to try to move towards something consistent rather than fragmenting the clients more

hayes-mysten:
I don't think this solution solves the original problem well.  If I understand the problem correctly, when using the core API you can't tell why an object errored.  This solution means you need to know the which RPC implementation is being used, and the specific error code format and meanings for that rpc.  The proposed hierarchy is pretty confusing and its not clear and doubles down on the bad error patterns we already have. 

I think we probably should take a step back and think through how errors should be surfaced in general.

* core methods (and to some extent top level methods) on all clients should return consistent errors
* consumers of clients should not need to think about or understand which underlying RPC is being used
* we should be able to consistently use instanceof checks across errors from all 3 clients
* if you need to make decisions based on error details, that should not be tied to an rpc implementation
* if an rpc implementation provides additional information, we should expose it


I think given this, what we probably want is:

* A base SuiClientError (exists)
* Specific sub-classes for known error cases (ObjectError exists, but is not used consistently)
* a way to tell what type of object error occured (does not really exist, existing "code" property is inconsistent across implementation)
* a way to get the rpc specific error details (I think this could be attached to the base client via a sub class of a new class like RpcErrorDetails or something (error.rpcDetails.$kind === 'grpc', error.rpcDetails.code = grpc error code?)

I don't have an exact implementation in mind, but I think something that doesn't specialize the error class per implementation would be better here

hayes-mysten:
I skimmed the most recent changes, and not all the things flagged by the review bot are valid, but the current iteration has a ton of real issues.  I am not sure this is a good candidate for something to just throw at claude.  It's obviously not understanding the problem space well and making a lot of incorrect assumptions.  (I wouldn't just go make the changes that the review bot flagged either).  I think there is a lot of subtly here that is not trivial to understand. 

Is this critical for a project you are working on, or just something you noticed and wanted to improve?

Danny-Devs:
**Edit:** Just realized I forgot to answer your question directly — this is just something I noticed and wanted to challenge myself on, not critical for any project of mine.

Hey Hayes — sorry the first pass was so directionally off. I appreciate you taking the time to spell out the principles instead of just closing it.

I took the five points to heart and rebuilt the PR around them. Quick tour of where things landed:

1. **Consistent errors across RPCs** — `ObjectErrorCode` is now a transport-agnostic union (`'notFound' | 'deleted' | 'unknown'`). JSON-RPC distinguishes `deleted` at the wire level so it maps `deleted` → `'deleted'` and `notExists`/`dynamicFieldNotFound` → `'notFound'`; gRPC and GraphQL can't distinguish (gRPC's `NOT_FOUND` is not specific; GraphQL omits absent objects without saying why), so both collapse to `'notFound'`. `displayError`/`unknown` surface as `'unknown'`.

2. **Consumers shouldn't need to know which RPC** — `error.code` and `error.objectId` mean the same thing on all three transports. The `fromResponse` factory is gone; each transport owns its own mapping.

3. **`instanceof` works across clients** — gRPC used to emit plain `new Error(...)`; it now emits `ObjectError`. `GraphQLResponseError` now extends `SuiClientError` (previously extended `Error`), and the GraphQL multi-error path wraps the aggregate in `SuiClientError` with `transportDetails: { $kind: 'graphql' }` on the cause — so `instanceof SuiClientError` is genuinely universal across all three transports, not just for object errors. `core-resolver.ts` rethrows the original `ObjectError` instance when one input is invalid, and throws the new `AggregateObjectError extends SuiClientError` when more than one is — restoring multi-error reporting that was previously flattened to a concatenated plain `Error`.

4. **Decisions not tied to RPC shape** — `client/errors.ts` no longer imports from `../jsonRpc/`. Each transport mapper uses `satisfies never` so upstream wire additions fail the type-check instead of silently falling through.

5. **RPC-specific detail still accessible** — `TransportDetails` is a `$kind`-tagged union (`'jsonRpc' | 'grpc' | 'graphql'`) on the `SuiClientError` base. `ObjectError`'s `options` arg is optional and the field is honestly optional on the base (no `declare` narrowing on the subclass) — so consumers can construct `ObjectError` directly without ceremony, while every real SDK construction site still attaches a payload.

Also tightened `GetObjectsResponse.objects` from `(Object | Error)[]` → `(Object | ObjectError)[]` and fixed cross-transport `objectId` parity (GraphQL now preserves the raw input id).

Let me know what you think about the refactor. Happy to split into smaller PRs if the surface is too big, or iterate further on anything that doesn't match what you had in mind — and of course happy to close it out if it's still not the direction you want.

Danny-Devs:
Merged latest main in case the PR refactor is in the right direction. Can close it out if it's not!

Danny-Devs:
Closing this out — thanks for the guidance and the reviews.

PR #999, opened 2026-04-03
Update wallet-sdk coin-flows for address balances and framework functions

hayes-mysten:
## Description

Teach the `@mysten/wallet-sdk` transaction analyzer about address-balance withdrawals (`FundsWithdrawal` inputs), the `0x2::coin` / `0x2::balance` framework functions, coin reservation refs (including the backwards-compat `Object`-input shape old SDKs submit via `listCoins`), and sponsored transactions. The old sender-only `coinFlows` rule is superseded by a per-address `balanceFlows` rule that tracks signed deltas across every party that touched tracked value.

### New: `balanceFlows`

Per-address signed balance deltas. Negative = value left the address on net; positive = value arrived.

```ts
interface BalanceFlowsResult {
  byAddress: Record<string, CoinFlow[]>; // every address that touched tracked value
  sender: CoinFlow[];                    // the tx sender's signed flows
  sponsor: CoinFlow[] | null;            // gas payer's flows when sponsored, else null
}
```

For each coin type the sum across all addresses is zero, modulo coins consumed to an unknown destination (e.g. a non-framework `MoveCall` that takes ownership). The analyzer flags cases that would quietly break conservation (value merged into `zero()` and never transferred, sponsor withdrawal with no `gasData.owner`, `into_balance`/`from_balance` on `tx.gas`, malformed `redeem_funds` argument) as analysis issues rather than silently laundering value.

### Framework support

The analyzer now recognizes these `0x2::coin` / `0x2::balance` functions and updates tracked coin balances / per-address deltas accordingly:

- Address balance: `redeem_funds`, `send_funds` (coin and balance modules)
- Conversions: `coin::into_balance`, `coin::from_balance`
- Splits: `coin::split`, `balance::split`, `coin::take`, `balance::withdraw_all`
- Joins: `coin::join`, `balance::join`, `coin::put`
- Zero: `coin::zero`, `balance::zero`

### Coin reservations

Coin reservation refs are surfaced as synthetic entries in the `objects` / `coins` / `gasCoins` analyzers with `isCoinReservation: true` on the resulting `AnalyzedObject` and balance decoded from the digest:

- **Gas-payment reservations** are always `Coin<SUI>` and owned by the gas payer, so they're synthesized locally with no lookup.
- **`Object`-input reservations** (the backwards-compat shape old SDKs submit after receiving synthetic coins from `listCoins`) don't encode the coin type in the ref. The analyzer unmasks each ref's object id (XOR with the chain identifier) and batches the accumulator-field lookup into the same `getObjects` request as the regular inputs; the coin type is parsed off the `Field<Key<Balance<T>>, U128>` struct tag that comes back.

Net cost per analysis: 1 `getObjects` call + 1 `getChainIdentifier` (client-cached). Transport-agnostic — works on JSON-RPC, gRPC, and GraphQL because the accumulator field is a regular on-chain object readable via the abstract `client.core.getObjects` API.

### Gas handling

Gas-payment coins are smashed into a single `tx.gas` coin owned by the gas payer. The full combined balance is charged to the gas payer up front, and the budget is deducted from the combined coin. When `gasData.payment` is empty, gas is paid from the gas owner's address balance; the `tx.gas` tracked coin is flagged `payFromAddressBalance` so `splitCoins(tx.gas, …)` charges the overflow to the gas payer rather than emitting a spurious over-draw issue.

### Deprecation

`coinFlows` is now a thin wrapper over `balanceFlows.sender` and is deprecated. Its entries use the pre-PR convention (positive = spent). Migrate to `balanceFlows.sender` for signed flows, or iterate `balanceFlows.byAddress[senderAddress]`.

### Other changes

- `inputs.ts`: new `Withdrawal` variant on `AnalyzedCommandInput` (amount, coinType, withdrawFrom).
- `coin-value.ts`: migrated to `balanceFlows.sender` (negative amounts) instead of the deprecated `coinFlows.outflows`.
- `@mysten/sui/utils`: exports `isCoinReservationDigest`, `parseCoinReservationBalance`, `xorCoinReservationObjectId`, `parseAccumulatorFieldCoinType`.
- `MockSuiClient`: configurable address balances via `setAddressBalance()`, configurable chain identifier via `setChainIdentifier()`.

## Test plan

- `pnpm --filter @mysten/wallet-sdk test` — 94 tests pass
- `pnpm --filter @mysten/sui test` — 353 tests pass (includes 13 for the new reservation utilities)
- `pnpm --filter @mysten/wallet-sdk run oxlint:check && pnpm --filter @mysten/wallet-sdk run prettier:check` — clean
- `pnpm --filter @mysten/sui run oxlint:check && pnpm --filter @mysten/sui run prettier:check` — clean
- `pnpm turbo build --filter=@mysten/wallet-sdk --filter=@mysten/sui` — builds clean
- End-to-end devnet verification: faucet → `coin::send_funds` into a fresh address's SUI AB → construct a reservation ref locally → run analyzer → asserts the `Coin<SUI>` type is resolved from the on-chain accumulator field and `balanceFlows.sender` correctly reports the outflow.

---

### AI Assistance Notice


packages/wallet-sdk/src/transaction-analyzer/rules/coins.ts
```diff
@@ -55,18 +60,39 @@ export const gasCoins = createAnalyzer({
...
-						};
-					}),
-				),
-			};
+			const SUI_TYPE = normalizeStructTag('0x2::sui::SUI');
+			const result: AnalyzedCoin[] = [];
+
+			for (const ref of data.gasData.payment ?? []) {
+				// Synthetic coin reservation refs encode their balance in the digest
+				if (isCoinReservationDigest(ref.digest)) {
```
hayes-mysten: I think we just want to omit these from the coins rule

hayes-mysten: We should probably have a reservations or withdrawals rule instead

packages/wallet-sdk/src/transaction-analyzer/rules/coins.ts
```diff
@@ -54,19 +59,68 @@ export const gasCoins = createAnalyzer({
...
+	ref: { objectId: string; version: string; digest: string };
+}
+
+/**
+ * Synthetic coin reservation refs that may appear in `gasData.payment`. Each
+ * represents a pre-authorized withdrawal from the owner's address balance
+ * accumulator. The balance is encoded in the ref's digest, and the object
+ * does not exist on-chain — hence why these are kept separate from `gasCoins`.
+ */
+export const coinReservations = createAnalyzer({
```
hayes-mysten: we actually want a withdrawal analyzer that tracks both withdrawals and coin reservations

PR #1015, opened 2026-04-13
Improvement sort

synasapmob:
## Description

Fix blob sorting in encodeQuilt by avoiding in-place mutation of the input blobs array while preserving the existing sorting behavior. This ensures compatibility with other SDK implementations and maintains deterministic output.

## Test plan

- Verified that sorting output remains unchanged compared to previous behavior
- Confirmed that the original blobs array is not mutated after sorting

---

### AI Assistance Notice

- [x] I used AI for docs / tests, but manually wrote the source code.

### DIFF
--- a/.changeset/quiet-pans-promise.md
+++ b/.changeset/quiet-pans-promise.md
@@ -1,0 +1,5 @@
+---
+"@mysten/walrus": patch
+---
+
+Fix quilt blob sorting in `encodeQuilt` by avoiding in-place mutation of the input `blobs` array while preserving existing sorting behavior.
--- a/packages/walrus/src/utils/quilts.ts
+++ b/packages/walrus/src/utils/quilts.ts
@@ -140,7 +140,7 @@ export function encodeQuilt({ blobs, numShards, encodingType }: EncodeQuiltOptio
 		encodingType,
 	);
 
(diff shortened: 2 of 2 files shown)


hayes-mysten:
We probably should not be mutating when sorting, but the rest of the changes would make the TS SDK inconsistent with the rust SDK.  If we want to change how quilts are sorted, we'd want to ensure this is consistent between implementations.

I'd be hesitant to change just the SDK without thoroughly exploring the implications

hayes-mysten:
The implications of changing this are subtle, and probably not critical, but the main effect is that all your quilt/blob IDs produced for the same inputs would be different in rust vs TS, this means that you can't produce the same blobs or easily verify data generated by the other SDK.

For most users this won't matter, but it would cause issues for anyone trying to re-construct existing blobs.  We also occasionally test inputs across both SDKs to ensure they produce consistent bytes when debugging/testing as we make changes

synasapmob:
Hey @hayes-mysten.

I agree that changing the sorting behavior introduces inconsistencies with the Rust SDK and affects determinism across implementations.

This PR now only avoids in-place mutation of the `blobs` array while preserving the existing sorting behavior. No changes to sorting semantics are introduced, ensuring compatibility with other SDKs and existing data.

hayes-mysten (approved):

PR #1021, opened 2026-04-16
feat(zksend): composable claim via createClaimCommands

tamashi095:
## Summary
Adds `ZkSendLink.createClaimCommands(tx)` that returns claimed assets for composition within a caller-owned transaction. The caller decides what to do with each asset and calls `finalize()` to close the bag.

## Motivation
Currently, `createClaimTransaction` auto-transfers all claimed objects to the recipient. There's no way to intercept the claimed objects to compose additional operations (e.g., adding a record to a player, staking a token, depositing into a vault) in the same transaction.

## Changes
- New `createClaimCommands(tx)` method → `{ assets: ClaimedAsset[], finalize: () => void }`
- New `ClaimedAsset` type: `{ argument: TransactionObjectArgument, type: string, objectId: string }` — pairs the tx argument with on-chain type metadata so callers can filter by Move type or object ID
- `createClaimTransaction(address)` refactored as a thin wrapper (claim → transfer all → finalize). Backwards compatible.
- No auto-transfer, no callbacks — caller explicitly handles every asset

## Usage
```ts
const tx = new Transaction();
const { assets, finalize } = link.createClaimCommands(tx);

const record = assets.find(a => a.type.includes('::record::Record'));
tx.moveCall({ target: 'player::add_record', arguments: [player, record.argument] });
tx.transferObjects(assets.filter(a => a !== record).map(a => a.argument), address);

finalize(); // closes the bag — must call
```

## Test plan
- [x] All 5 existing contract link tests pass unchanged (backwards compatible)
- [x] New test: `createClaimCommands` — claims 2 DemoBear NFTs + SUI coin, filters assets by type, composes separate `transferObjects` calls, verifies link marked as claimed on-chain

### DIFF
--- a/packages/zksend/src/index.test.ts
+++ b/packages/zksend/src/index.test.ts
@@ -281,6 +281,75 @@ describe('Contract links', () => {
 		}
 	}, 60_000);
 
+	test('createClaimCommands — compose with claimed assets', async () => {
+		const link = client.zksend.linkBuilder({
+			sender: keypair.toSuiAddress(),
+		});
+
+		const bears = await createBears(2);
+
+		for (const bear of bears) {
+			link.addClaimableObject(bear.objectId);
+		}
+
+		link.addClaimableMist(100n);
+
+		const linkUrl = link.getLink();
+
+		await link.create({
+			signer: keypair,
+			waitForTransaction: true,
+		});
+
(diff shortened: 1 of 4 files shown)


hayes-mysten:
merged your commits as part of #1025

PR #1051, opened 2026-05-06
Enable isolatedDeclarations across the monorepo

hayes-mysten:
## Summary

Enables TypeScript's `isolatedDeclarations` flag repo-wide and resolves the ~2,000 violations it surfaces. This is required for the hardened per-package release flow (forthcoming follow-up PR), where `tsdown` uses oxc-transform's iso-decls dts emitter to ship `.d.mts` files without resolving any sibling/runtime types — meaning no `node_modules` need be installed during the release build.

All 40 packages now build clean with `pnpm turbo build --filter='./packages/*'`.

### What this PR does

- **`tsconfig.shared.json`**: turns on `isolatedDeclarations: true`. CI on every PR will catch incompliance going forward.
- **`@mysten/codegen` template** (`packages/codegen/src/move-module-builder.ts`): emits explicit `MoveStruct<{...}>` / `MoveEnum<{...}>` / `MoveTuple<[...]>` annotations on generated struct/enum/tuple declarations and `: (tx: Transaction) => TransactionResult` on move-call functions. Adds a new `tsType` rendering mode in `render-types.ts` that emits TS type expressions parallel to the existing `bcs` value-format.
- **Regenerated Move bindings** for `@mysten/walrus`, `@mysten/suins`, `@mysten/payment-kit`, `@mysten/deepbook-v3`, `@mysten/pas` against the latest sibling Move contract repos. There is some unrelated contract drift (notably in deepbook-v3 since the last regen).
- **`@mysten/codegen` package**: per-package `isolatedDeclarations: false` override (its zod schemas and `@stricli/core` generic factories make iso-decls compliance impractical without substantial duplication or escapes the rule doesn't accept). The codegen package isn't on the hardened release flow; it stays on the existing path.
- **`@mysten/bcs` foundation**: explicit return types on factory functions (`tuple`, `struct`, `enum`, `bytes`, `byteVector`, etc.) and on `BcsType`/`SerializedBcs` methods. The `SerializedBcs` brand symbol moved into the constructor (computed property names aren't allowed under iso-decls).
- **`@mysten/utils`**: explicit return types on `toBase58`/`fromBase58`.
- **Hand-fixes across 19 packages** for ~500 hand-written violations (Promise return types, BCS schema annotations, valibot schemas, React/Lit components, vanilla-extract CSS files, transaction builders, etc).
- **Regex codemod** for the 216 violations in `packages/sui/src/grpc/proto/*` (protoc-generated): `export const X = new X$Type();` → `export const X: X$Type = new X$Type();`.

### Trade-offs

- **`@mysten/codegen` is the one exception** to repo-wide iso-decls: its zod-driven config and stricli-driven CLI surface use generic types those libraries don't export, so explicit annotation is impractical. The override is documented inline.
- **Some valibot schemas** had their type aliases hand-written (instead of `v.InferOutput<typeof X>`) so iso-decls doesn't have to chase the schema's full type. Listed packages: `wallet-sdk`, `slush-wallet`, `walletconnect-wallet`, `enoki`, `enoki-connect`, `window-wallet-core`. The schema-output type and the manual type are kept in sync via `as v.GenericSchema<T, T>` casts.
- **Codegen drift**: regenerating Move bindings against current sibling contracts pulled in unrelated contract changes. Each regenerated package has both iso-decls type fixes AND any contract changes that occurred since the last regen — flagged here so reviewers can sanity-check the bindings.

## Test plan

- [x] `pnpm turbo build --filter='./packages/*' --force` — 40/40 packages build cleanly
- [x] `pnpm exec tsc --noEmit -p tsconfig.json` per package — zero TS9 errors
- [x] `tsdown` (which the hardened release flow uses for dts emission) — builds clean for every package with a `tsdown.config.ts`
- [ ] Run the existing test suites: `pnpm test`
- [ ] Spot-check the regenerated Move bindings vs. sibling contracts to confirm no functional regressions in the contract surface

---

### AI Assistance Notice


packages/dapp-kit-next/packages/dapp-kit-core/src/utils/name.ts
```diff
@@ -6,7 +6,10 @@ import type { DAppKitCompatibleClient } from '../core/types.js';
 
 const cache = new Map<string, string | null>();
 
-export async function resolveNameServiceName(client: DAppKitCompatibleClient, address: string) {
+export async function resolveNameServiceName(
+	client: DAppKitCompatibleClient,
+	address: string,
+): Promise<string | null | undefined> {
```
hayes-mysten, line 12: we should return null OR undefined unless there is a meaningful difference

packages/dapp-kit-next/packages/dapp-kit-core/src/utils/wallets.ts
```diff
@@ -14,11 +14,17 @@ import { getWalletAccountFeature } from '@wallet-standard/ui';
 import { getWalletForHandle_DO_NOT_USE_OR_YOU_WILL_BE_FIRED as getWalletForHandle } from '@wallet-standard/ui-registry';
 import { ChainNotSupportedError, FeatureNotSupportedError } from './errors.js';
 
-export const requiredWalletFeatures = [StandardConnect, StandardEvents] as const;
+export const requiredWalletFeatures: readonly ['standard:connect', 'standard:events'] = [
+	StandardConnect,
+	StandardEvents,
+] as const;
```
hayes-mysten, line 20: is this really the only way for this to work with isolated modules?

packages/dapp-kit-next/packages/dapp-kit-core/src/utils/wallets.ts
```diff
@@ -62,7 +68,7 @@ export function getAccountFeature<TAccount extends UiWalletAccount>({
 
 export function tryGetAccountFeature<TAccount extends UiWalletAccount>(
 	...args: Parameters<typeof getAccountFeature<TAccount>>
-) {
+): object | null {
```
hayes-mysten, line 71: this seems wrong

packages/deepbook-v3/src/transactions/deepbook.ts
```diff
@@ -41,90 +41,94 @@ export class DeepBookContract {
...
-				tx.pure.bool(payWithDeep),
-				tx.pure.u64(expiration),
-				tx.object.clock(),
-			],
-			typeArguments: [baseCoin.type, quoteCoin.type],
-		});
-	};
+	placeLimitOrder =
+		(params: PlaceLimitOrderParams) =>
+		(tx: Transaction): void => {
```
hayes-mysten, line 46: Maybe we should have a TransactionThunk<T = void> type we can use instead

packages/deepbook-v3/src/transactions/deepbookAdmin.ts
```diff
@@ -38,179 +38,197 @@ export class DeepBookAdminContract {
...
 	 */
-	createPoolAdmin = (params: CreatePoolAdminParams) => (tx: Transaction) => {
-		tx.setSenderIfNotSet(this.#config.address);
-		const { baseCoinKey, quoteCoinKey, tickSize, lotSize, minSize, whitelisted, stablePool } =
-			params;
-		const baseCoin = this.#config.getCoin(baseCoinKey);
-		const quoteCoin = this.#config.getCoin(quoteCoinKey);
+	createPoolAdmin =
+		(params: CreatePoolAdminParams) =>
+		(tx: Transaction): void => {
```
hayes-mysten, line 43: Or maybe we have a MoveCall<Params, Return = void> type instead?

packages/payment-kit/src/calls.ts
```diff
@@ -73,7 +76,9 @@ export class PaymentKitCalls {
 	 * tx.add(client.paymentKit.call.processEphemeralPayment({ nonce, coinType, sender, amount, receiver }));
 	 * ```
 	 */
-	processEphemeralPayment = (options: ProcessEphemeralPaymentOptions) => {
+	processEphemeralPayment = (
```
hayes-mysten, line 79: Lets add a helper for these kinds of methods to the sui sdk that we can reuse here and in deepbook (and other sdks)

hayes-mysten:
this is probably too complicated, and bcs gets very ugly with this

PR #1056, opened 2026-05-08
fix(release): make publish-time build work under prod-only install

hayes-mysten:
## Description

The new release flow's `pnpm install --filter <pkg>... --prod` install broke for every workspace package that uses the standard `tsc --noEmit && tsdown` build script. When the orchestrator dispatched `release-<pkg>.yml` workflows after #1049 merged, **all 9 dispatches failed at the Build step**. Three independent root causes, all surfacing before `pnpm publish` was reached, so nothing actually published.

### 1. Workspace peer deps weren't being auto-installed under `--prod`

21 packages declare `@mysten/sui` as both `peerDependencies` AND `devDependencies`. The duplication satisfied dev installs while `autoInstallPeers: false` was set in `pnpm-workspace.yaml`. Under `--prod`, the devDep entry took precedence and was excluded, leaving sui unresolvable at build time:

```
src/objects.ts(4,34): error TS2307: Cannot find module '@mysten/sui/transactions'
```

**Fix:** Flipped `autoInstallPeers: false → true` and dropped the redundant peer+devDep duplications across 22 package.jsons. Workspace peers now auto-install under `--prod`, dev installs still get sui via the same mechanism, and **devDeps stay out of the release runner** (the original supply-chain hygiene goal stays intact).

### 2. `wallet-sdk` / `wallet-standard` peer specifier was `*`

Wildcard ranges aren't satisfied by workspace resolution, so pnpm pulled npm-published `@mysten/sui@2.16.0` alongside the workspace copy — two distinct `Transaction` types caused TS2322 errors throughout the dapp-kit / walrus chain.

**Fix:** Switched both to `workspace:^`. Resolves to `^<version>` at publish time — also a real improvement vs. `*`, which let consumers ship arbitrarily-old sui.

### 3. Source trees included build-time devDep imports

- `packages/sui/src/transactions/__tests__/{Transaction,bcs}.test.ts` → moved to `test/unit/transactions/`. The build's `tsc --noEmit` was trying to typecheck these against `vitest` (a workspace devDep, not in `--prod`).
- `packages/dapp-kit-core/src/web/*.stories.ts` → moved to `stories/` at the package root. The build was pulling in `@storybook/web-components` (workspace devDep). Updated `.storybook/main.ts` glob, added `stories/tsconfig.json` mirroring the existing `test/` layout, and chained the new project into `test:typecheck` via `tsc -b ./test ./stories` so stories keep getting typechecked outside the publish build. Verified `storybook build` still works.
- `packages/docs/scripts/build-docs.ts` dropped its `npx prettier --write` post-processing — formatting of generated LLM markdown that the published artifact didn't need, and prettier wasn't in the prod install.

### Template fixes

- Added `@types/semver` to the `tools:` list for `release-{ledger-signer,signers,ledgerjs-hw-app-sui}.yml`. `semver` itself is a runtime dep already; only the types are devDep-only and `tsc --noEmit` needs them.
- Moved the artifact-slug step from after `Build` to right after `Checkout`. Previously, when Build failed, the `always()` artifact upload used an empty slug and produced the invalid `name: tarball-`.

## Verification

Reproduced the CI flow locally for each failure mode (`pnpm install --filter <pkg>... --prod --frozen-lockfile --ignore-scripts` + sandboxed tools install + build). All pass:

| Package | Was failing on | Now |
|---|---|---|
| `@mysten/sui` | tests in src/ | ✓ |
| `@mysten/ledger-signer` | sui peer + @types/semver | ✓ |
| `@mysten/walrus` | dapp-kit chain + dual sui | ✓ |
| `@mysten/docs` | prettier missing | ✓ |
| `@mysten/dapp-kit-core` | storybook stories | ✓ |
| `@mysten/dapp-kit-react` | sui peer | ✓ |
| `@mysten/enoki`, `@mysten/seal`, `@mysten/signers`, `@mysten/{aws,gcp,webcrypto}-kms-signer` | sui peer | ✓ |

Also verified `pnpm --filter @mysten/dapp-kit-core run test:typecheck` (now `tsc -b ./test ./stories`) and `storybook build` both pass.

## Test plan

- [x] Local: reproduce CI flow with prod-only install + tools sandbox for 12 packages
- [x] `storybook build` succeeds with stories in new location
- [x] `test:typecheck` covers both test/ and stories/
- [ ] Reviewer to validate: orchestrator dispatch on the merged main runs all per-package release workflows in dry-run mode and they pass through to `pnpm publish --dry-run`

---

### AI Assistance Notice


hayes-mysten:
@clud-bot please review

hayes-mysten:
@clud-bot check again

PR #1057, opened 2026-05-08
chore: release all packages via the new flow

hayes-mysten:
## Summary

Three coordinated changes that, together, cause the next Version Packages PR merge to publish every public `@mysten` package via the new orchestrator + per-package workflows for the first time:

### 1. Fix the `pnpm publish` workspace-protocol resolution (`_release-package.yml`)

The first orchestrator run on `main` after #1056 merged dispatched 9 release workflows. Six passed dry-run; **three failed at the `Publish to npm` step** with:

```
ERR_PNPM_CANNOT_RESOLVE_WORKSPACE_PROTOCOL  Cannot resolve workspace
protocol of dependency "@mysten/codegen" because this dependency is not
installed.
```

Same root cause for all 3:
- `pnpm publish` rewrites every `workspace:` ref in the published manifest, including ones in `devDependencies`
- Under the publish-time `--prod` install, workspace devDeps aren't symlinked into the target's `node_modules`
- pnpm can't read the dep's version → bail

Audit shows **11 packages** with this pattern (`deepbook-v3`, `enoki`, `kiosk`, `move-bytecode-template`, `pas`, `payment-kit`, `suins`, `wallet-sdk`, `walrus`, `walrus-wasm`, plus the `create-dapp` template). Only 3 surfaced in the last run because only those bumped versions; the other 8 would hit the same wall once they bump.

**Fix**: a 15-line template step that drops just the `workspace:`-prefixed entries from `devDependencies` before `pnpm publish`. Third-party devDeps (`vitest`, `@types/*`, `@iarna/toml`, etc.) keep their real ranges and are left intact for any tooling that reads them. npm doesn't install `devDependencies` for consumers, so the absence in the published manifest is consumer-invisible.

Verified locally for the 3 known-failing packages — all reach `Publishing to https://registry.npmjs.org/ with tag latest (dry-run)` cleanly.

### 2. Flip `dry_run` defaults to `false`

- 33 `release-<pkg>.yml` files: input default + forwarded value
- `release.yml` orchestrator: `DRY_RUN` env default
- `_release-package.yml` template: input default

`workflow_dispatch` still accepts `dry_run: true` for manual validation, but the orchestrator now defaults to real publishes.

### 3. All-packages patch changeset

Empty patch bump for every public `@mysten` package (33 entries). Forces every package to bump versions when the resulting Version Packages PR merges so the orchestrator dispatches all 33 `release-<pkg>.yml` runs in real (non-dry-run) mode, validating the new flow end-to-end.

## Test plan

- [x] Local: minimal-install + strip-devDeps + `pnpm publish --dry-run` succeeds for `@mysten/{deepbook-v3,enoki,suins}` (the three that failed in the post-#1056 orchestrator run).
- [ ] Reviewer to verify: trusted publishing is configured on npm.com for every `@mysten/*` package the orchestrator might dispatch (33 packages). Without this, the real `pnpm publish` step will fail with auth errors.
- [ ] After this PR's resulting Version Packages PR merges: confirm the orchestrator dispatches all 33 release workflows and they reach `Publishing to https://registry.npmjs.org/ with tag latest`.

## What happens after merge

1. Changesets bot opens (or updates) the Version Packages PR using all 33 patch bumps from the changeset in this PR.
2. Reviewer merges the Version Packages PR.
3. Turborepo CI runs on bumped main; on success the orchestrator fires.
4. Orchestrator finds 33 packages with versions not yet on npm; dispatches all 33 `release-<pkg>.yml` workflows in real (non-dry-run) mode.
5. Each runs the now-fixed publish chain → publishes via OIDC trusted publishing.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/release-all-with-new-flow.md
+++ b/.changeset/release-all-with-new-flow.md
@@ -1,0 +1,40 @@
+---
+'@mysten/aws-kms-signer': patch
+'@mysten/bcs': patch
+'@mysten/codegen': patch
+'@mysten/create-dapp': patch
+'@mysten/dapp-kit': patch
+'@mysten/dapp-kit-core': patch
+'@mysten/dapp-kit-react': patch
+'@mysten/deepbook-v3': patch
+'@mysten/docs': patch
+'@mysten/enoki': patch
+'@mysten/enoki-connect': patch
+'@mysten/gcp-kms-signer': patch
+'@mysten/kiosk': patch
+'@mysten/ledger-signer': patch
+'@mysten/ledgerjs-hw-app-sui': patch
+'@mysten/move-bytecode-template': patch
+'@mysten/mvr-static': patch
+'@mysten/pas': patch
+'@mysten/payment-kit': patch
+'@mysten/seal': patch
+'@mysten/signers': patch
+'@mysten/slush-wallet': patch
+'@mysten/sui': patch
+'@mysten/suins': patch
+'@mysten/utils': patch
+'@mysten/wallet-sdk': patch
+'@mysten/wallet-standard': patch
+'@mysten/walletconnect-wallet': patch
+'@mysten/walrus': patch
+'@mysten/walrus-wasm': patch
+'@mysten/webcrypto-signer': patch
+'@mysten/window-wallet-core': patch
+'@mysten/zksend': patch
+---
+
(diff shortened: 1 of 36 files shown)


hayes-mysten:
@clud-bot review this quickly

PR #1059, opened 2026-05-08
fix(release): restore type defs + per-package docs in published tarballs

hayes-mysten:
## Summary

Two regressions surfaced from the audit of the first round of releases via the new flow (#1057's all-packages bump). One is broken on npm right now; the other dropped LLM-consumed markdown from most packages' tarballs.

### 1. `@mysten/dapp-kit@1.0.5` (legacy) shipped with zero type definitions

The legacy dapp-kit build runs esbuild → `dist/*.mjs`, then `tsc --emitDeclarationOnly` → `dist/*.d.ts`. The tsc step was wrapped in `await exec('pnpm tsc ...')`, which silently masked tsc's exit status in CI — the script printed `Build complete!` even when no `.d.ts` files were emitted. Result: 71 `.mjs` files in the published tarball, **0 `.d.ts`**. TypeScript consumers can't use `1.0.5`.

**Fix**: replace `exec` with `spawnSync(stdio: 'inherit')` so tsc output and exit code are visible, plus a hard assertion after the call that `dist/**/*.d.ts` is non-empty. Verified locally — emits 71 `.d.ts` files, all packed into the tarball.

### 2. `docs/*.md` missing from most published tarballs

`turbo.json` registers `dependsOn: ["build:docs"]` on the `build` task, so the legacy `pnpm turbo run build` flow regenerated each package's `docs/*.md` (LLM-consumed markdown listed in each package's `files` field) as a transitive turbo dependency. The new flow uses `pnpm --filter <pkg>... run build` directly, which bypasses turbo's task graph — so `build:docs` never fired and `docs/` went stale or absent across 13 packages.

Affected (per-tarball diff vs previous published version):
```
bcs, codegen, dapp-kit-core, dapp-kit-react, dapp-kit (legacy), kiosk,
payment-kit, seal, slush-wallet, sui, walrus, zksend, docs
```

**Fix**: add a `Build docs` step after the main `Build` step that runs `pnpm --filter <pkg> --if-present run build:docs`. `--if-present` skips the 20 packages without that script. Skipped entirely when a package overrides `build_command` (only `docs` and `walrus-wasm` do that). Verified locally for both top-level (sui → 41 docs files) and nested (dapp-kit-core → 29 docs files) packages under the prod-only install.

### 3. All-packages patch changeset

Forces every public `@mysten` package to bump on the next Version Packages PR merge so we re-publish all 33 with the corrected content (types where they were missing, docs where they were missing). Same shape as #1057's all-packages changeset.

## Audit context

Other diffs spotted during the audit are accounted for and don't need fixes:

- **Sui (and similar in others) — type-union member reordering**: `"A" \| "B" \| "C"` vs `"B" \| "C" \| "A"`. Semantically identical; cosmetic emit-order non-determinism between tsdown runs.
- **`@mysten/signers` removed `dist/aws/`, `dist/gcp/`, etc.**: intentional refactor in #1052 splitting backends into separate packages.
- **`@mysten/move-bytecode-template`, `@mysten/walrus-wasm` added `index.d.mts`**: tsdown emit shape change; types are correct.
- **4 new signer packages added `CHANGELOG.md`**: first time published with changesets-managed changelogs.

## Test plan

- [x] Local: legacy dapp-kit fresh build emits 71 `.d.ts` files (matching pre-`1.0.5` baseline).
- [x] Local: prod-only install + build:docs for sui generates `docs/index.md`, `docs/llms-index.md`, etc. (41 files).
- [x] Local: same for nested package (dapp-kit-core, 29 files).
- [ ] Reviewer to verify on the resulting Version Packages PR merge: each release runner's CI log shows `Build complete! (<n> .d.ts files emitted)` for legacy dapp-kit and `Generated <n> docs files` lines for the 13 packages with `build:docs`.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/restore-dapp-kit-types.md
+++ b/.changeset/restore-dapp-kit-types.md
@@ -1,0 +1,5 @@
+---
+'@mysten/dapp-kit': patch
+---
+
+Restore type definitions in published package.
--- a/.changeset/restore-docs.md
+++ b/.changeset/restore-docs.md
@@ -1,0 +1,37 @@
+---
+'@mysten/aws-kms-signer': patch
+'@mysten/bcs': patch
+'@mysten/codegen': patch
+'@mysten/create-dapp': patch
+'@mysten/dapp-kit': patch
+'@mysten/dapp-kit-core': patch
+'@mysten/dapp-kit-react': patch
+'@mysten/deepbook-v3': patch
+'@mysten/docs': patch
+'@mysten/enoki': patch
+'@mysten/enoki-connect': patch
+'@mysten/gcp-kms-signer': patch
+'@mysten/kiosk': patch
+'@mysten/ledger-signer': patch
+'@mysten/ledgerjs-hw-app-sui': patch
+'@mysten/move-bytecode-template': patch
+'@mysten/mvr-static': patch
+'@mysten/pas': patch
+'@mysten/payment-kit': patch
+'@mysten/seal': patch
+'@mysten/signers': patch
+'@mysten/slush-wallet': patch
+'@mysten/sui': patch
+'@mysten/suins': patch
+'@mysten/utils': patch
+'@mysten/wallet-sdk': patch
+'@mysten/wallet-standard': patch
+'@mysten/walletconnect-wallet': patch
+'@mysten/walrus': patch
+'@mysten/walrus-wasm': patch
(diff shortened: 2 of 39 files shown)


hayes-mysten:
@clud-bot review this

PR #1070, opened 2026-05-15
fix(sui): set ValidDuring expiration for gasless txs over JSON-RPC

hayes-mysten:
## Description

When a user calls `tx.setGasPrice(0)` to opt into the free-tier / address-balance gasless flow on a PTB whose only inputs are a `FundsWithdrawal` and pure args (e.g. `0x2::balance::send_funds<USDC>` for a recipient), `coreClientResolveTransactionPlugin`'s `setGasBudget` runs a simulate with `gasData.payment: []`. JSON-RPC's `dryRunTransactionBlock` then rejects the simulate with:

> Error checking transaction input objects: Invalid transaction expiration: Transactions must either have address-owned inputs, or a ValidDuring expiration with at most two epochs of validity

The gRPC server-side simulate fills in this expiration itself ([sui#26576](https://github.com/MystenLabs/sui/pull/26576)), but JSON-RPC's `dryRunTransactionBlock` does not. The existing resolver does set a `ValidDuring` later (after `setGasPayment` zeroes out the payment), but by then the simulate inside `setGasBudget` has already failed.

This change pre-fills the `ValidDuring` expiration before calling `setGasBudget` whenever `gasData.price === 0n`. The expiration would have been set anyway by the existing post-resolve check, so the wire bytes are unchanged — we just move the assignment earlier so the simulate can pass on JSON-RPC. `getCurrentSystemState` / `getChainIdentifier` are also added to the existing parallel fetch when gasless to avoid an extra round trip.

## Test plan

- Reproduced the bug end-to-end on testnet against `https://fullnode.testnet.sui.io:443`:
  - `tx.balance({ type: USDC, balance: 10_000n })` → `0x2::balance::send_funds<USDC>` → `tx.setGasPrice(0)`
  - Without the fix: `SuiJsonRpcClient.signAndExecuteTransaction` fails inside `setGasBudget`'s dryRun with the error quoted above.
  - With the fix: same tx succeeds. Confirmed digest `13QKGrXeHegj8ztHaFMsZdMDfVYE4RxE8qHucbAx4id7`.
- `pnpm --filter @mysten/sui exec tsc --noEmit` passes.
- `pnpm --filter @mysten/sui lint` passes.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/free-tier-jsonrpc-expiration.md
+++ b/.changeset/free-tier-jsonrpc-expiration.md
@@ -2,10 +2,12 @@
 '@mysten/sui': patch
 ---
 
-Set a `ValidDuring` expiration before simulating gasless transactions so the
-resolver works over JSON-RPC. Previously, calling `tx.setGasPrice(0)` on a PTB
-with no address-owned inputs would fail at the `setGasBudget` dryRun with
-"Transactions must either have address-owned inputs, or a ValidDuring
-expiration with at most two epochs of validity". The gRPC server fills in the
-expiration on its own (sui#26576) but JSON-RPC's dryRun does not, so the SDK
-now sets it client-side whenever `gasData.price === 0`.
+Pass a `ValidDuring` expiration as a simulate-only override when the resolver's
+budget computation is about to dryRun a gasless transaction. Previously,
+calling `tx.setGasPrice(0)` on a PTB with no address-owned inputs would fail
+at the `setGasBudget` dryRun on JSON-RPC with "Transactions must either have
+address-owned inputs, or a ValidDuring expiration with at most two epochs of
+validity". The gRPC server fills the expiration in on its own (sui#26576) but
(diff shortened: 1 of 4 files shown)


hayes-mysten:
@clud-bot review

hayes-mysten:
@clud-bot the referenceGasPrice fallback gap is fixed in 46b2a4c0 — the check now runs after the fallback so it uses the final `gasData.price`. Mind taking another look?

hayes-mysten:
@clud-bot widened the fix in cdf9a771 — the previous `gasPrice === 0` check was a proxy for "PTB has no address-owned inputs," but we can't tell `AddressOwner` from `Immutable` for an `ImmOrOwnedObject` input without owner info we don't keep. So I dropped the gasPrice check and now provide a simulate-only `ValidDuring` whenever `setGasBudget` actually simulates (no preset budget) and the caller hasn't set an expiration. Wire bytes for the final tx are unchanged. Added unit tests for the new path.

PR #1074, opened 2026-05-18
update schemas

hayes-mysten:
## Description

Regenerates gRPC, GraphQL, and JSON-RPC types from upstream sources, and adds a new ForkingService gRPC client pulled from the `sui` repo.

Notable upstream changes picked up:

- **gRPC** (from `sui-apis`): `AccumulatorWrite` splits the old `value` field into `integer_value`, `integer_tuple`, and `event_digest_value` to support authenticated events; adds new `EventDigestEntry` message and `AccumulatorValue` enum. Other proto regen produces minor noise (`RepeatType.PACKED` → `RepeatType.UNPACKED` for message-type repeated fields, which is a correctness fix per the protobuf spec).
- **GraphQL** (from `sui-indexer-alt-graphql`): new `verifySignature` query (deprecates `verifyZkLoginSignature`), new `IntentScope` enum and `SignatureVerifyResult` type, new `digest` argument on `Query.checkpoint`, new `version` field on `TransactionEffects`.
- **JSON-RPC** (from `sui-open-rpc`): typo / wording fixes in generated parameter docs. The `DisplayFieldsResponse.data => Record<string, unknown>` override from #993 is preserved manually (upstream OpenRPC spec still reports it as `string`).

### ForkingService

The `sui-fork` tool defines its admin control plane in `sui/crates/sui-fork/proto/sui/forking/v1alpha/forking_service.proto`, which is **not** mirrored into `sui-apis`. To pick it up, `codegen:grpc-protoc` now also includes `../../../sui/crates/sui-fork/proto` as an `-I` path and compiles its `*.proto` alongside the rpc/v2 set. `generate-grpc-types.ts` walks both proto subtrees so `GrpcTypes` re-exports forking messages too. `ForkingServiceClient` is wired into `SuiGrpcClient` as `forkingService`, matching the pattern used for `ledgerService`, `stateService`, etc. The forking proto is self-contained (no imports), so no types are shared with or duplicated against rpc/v2.

ForkingService is admin-only and intended for use against `sui-fork` instances; calling it against a regular Sui node will fail.

## Test plan

- [x] `pnpm --filter @mysten/sui test:typecheck`
- [x] `pnpm --filter @mysten/sui build`

---

### AI Assistance Notice


hayes-mysten:
@clud-bot review

PR #1082, opened 2026-06-01
zklogin: reject claim inputs with JSON escapes in genAddressSeed

mskd12:
## Summary

Hardens zkLogin address derivation against the claim-value escaping divergence (audit finding `02`, LOW).

The zkLogin circuit derives the address seed from the **raw JWT bytes** of the key claim name, value, and aud. The SDK derives it from **JSON-decoded** claim values (`decodeJwt` → `jwtDecode`). When a claim contains a JSON escape, the two diverge — e.g. an honest username `abc\` is encoded `"abc\\"`, decodes to `abc\` (address A) here, but the circuit hashes the raw `abc\\` (address B). The user funds A while the proof authorizes B (a completeness failure, and a possible cross-user collision).

## What

`genAddressSeed` now rejects `name`, `value`, and `aud` if they contain a character that reveals an escape was used — `"`, `\`, or a control char (`< 0x20`). These cannot appear literally in a JSON string, so their presence in the *decoded* value proves the raw JWT used an escape.

```ts
assertNoJsonEscape(name, 'key claim name');
assertNoJsonEscape(value, 'key claim value');
assertNoJsonEscape(aud, 'aud');
```

Real `sub`/`aud` values never contain these, so no legitimate flow is affected (existing address tests unchanged).

## This is a stopgap, not the full fix

It catches the **mandatory** JSON escapes only. **Optional** escapes still slip through, because they decode to innocuous characters:
- `\/` for `/`
- `\uXXXX` for *any* character

The root cause is that `decodeJwt` uses a standard JSON parser that resolves escapes, whereas the circuit works on raw bytes. The complete fix is **circuit-aligned claim parsing** (extract claim values over the raw JWT bytes the way the circuit does). A `TODO` documenting this is added on `decodeJwt`.

## Test

- `address.test.ts`: backslash / double-quote / control-char rejection across name, value, and aud, plus a plain-input pass — 18 passed (was 8).
- `oxlint` 0/0, `prettier` clean.

### AI Assistance Notice

### DIFF
--- a/.changeset/zklogin-reject-escaped-claim-values.md
+++ b/.changeset/zklogin-reject-escaped-claim-values.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': patch
+---
+
+zkLogin: `genAddressSeed` now rejects key claim name, value, or aud that contain a JSON escape (`"`, `\`, or a control character). The circuit derives the address seed from the raw JWT bytes, so an escaped claim value decodes differently than the circuit hashes and would yield a mismatched address. This is a stopgap; the complete fix is circuit-aligned claim parsing (see the TODO on `decodeJwt`).
--- a/.changeset/zklogin-reject-escaped-claim-values.md
+++ b/.changeset/zklogin-reject-escaped-claim-values.md
@@ -2,4 +2,4 @@
 '@mysten/sui': patch
 ---
 
-zkLogin: `genAddressSeed` now rejects key claim name, value, or aud that contain a JSON escape (`"`, `\`, or a control character). The circuit derives the address seed from the raw JWT bytes, so an escaped claim value decodes differently than the circuit hashes and would yield a mismatched address. This is a stopgap; the complete fix is circuit-aligned claim parsing (see the TODO on `decodeJwt`).
+zkLogin: `genAddressSeed` now rejects key claim name, value, or aud that contain a JSON escape (`"`, `\`, or a control character).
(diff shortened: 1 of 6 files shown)


.changeset/zklogin-reject-escaped-claim-values.md
```diff
@@ -0,0 +1,5 @@
+---
+'@mysten/sui': patch
+---
+
+zkLogin: `genAddressSeed` now rejects key claim name, value, or aud that contain a JSON escape (`"`, `\`, or a control character). The circuit derives the address seed from the raw JWT bytes, so an escaped claim value decodes differently than the circuit hashes and would yield a mismatched address. This is a stopgap; the complete fix is circuit-aligned claim parsing (see the TODO on `decodeJwt`).
```
benr-ml:
```suggestion
zkLogin: `genAddressSeed` now rejects key claim name, value, or aud that contain a JSON escape (`"`, `\`, or a control character).
```

packages/sui/src/zklogin/utils.ts
```diff
@@ -87,6 +87,21 @@ export function hashASCIIStrToField(str: string, maxSize: number) {
 	return poseidonHash(packed);
 }
 
+// Reject claim inputs whose decoded form reveals a JSON escape ('"', '\', control char):
+// the circuit hashes raw JWT bytes, so an escaped value would derive a different address.
+// TODO: stopgap — misses optional escapes (`\/`, `\uXXXX`); real fix lives in decodeJwt.
```
benr-ml: remove this todo (already has one in decodeJwt)

benr-ml (approved):

hayes-mysten (approved):

PR #1089, opened 2026-06-12
codegen: expose type tag getters on generated Move types

### LINKED ISSUE #426: feat(codegen): expose type tag getter for generated structs
Currently, codegen creates bcs models for module's structs, as well as wrappers for its methods. A helpful part that is missing, imo, is a helper function for each struct, that returns its TypeTag string.

These strings are constructed inside the methods to use in `argumentsTypes`, e.g.
```ts
`${packageAddress}::coin::Coin<${options.typeArguments[0]}>`
```
...but they are never exposed from the codegen.

As the result, if you want to use that TypeTag as a type param for some other call, you need to hardcode them in your code and basically repeat what codegen is already doing under the hood, which may become pretty sinister if you're dealing not with just `0x2::coin::Coin`, but with some custom work-in-progress APIs.

Would be nice to have something like this generated as well, e.g. for Coin struct:
```ts
export function CoinTypeTag(options: CoinTypeTagOptions) {
  const packageAddress = options.package ?? '0x2';
  return `${packageAddress}::coin::Coin<${options.typeArguments[0]}>`;
}
```
(idk about naming convention tho, should not overlap with actual methods of the module)

hayes-mysten:
## Description

This branch contains two pieces of work:

1. **Type tag getters for generated Move types** (closes #426).
2. **A workspace-wide dependency upgrade** and the fixes for the breakage it surfaced.

---

## 1. Type tag getters

Adds `typeTag` and `resolveTypeTag` methods to the generated `MoveStruct`, `MoveEnum`, and `MoveTuple` classes in the generated `utils/index.ts`, exposing the type tag strings codegen already constructs internally. Generated module output is unchanged — no new props or generics on instantiation sites — so every generated package picks this up by regenerating only its utils file (done in this PR for all in-repo consumers).

### `typeTag(options?)`

```ts
StakedWal.typeTag();                                    // '@local-pkg/walrus::staked_wal::StakedWal'
Balance.typeTag({ typeArguments: ['0x2::sui::SUI'] });  // '0x2::balance::Balance<0x2::sui::SUI>'
StakedWal.typeTag({ package: '0x9f9...' });             // package override, as on function wrappers

// nested calls compose:
MapOfBalances.typeTag({
  typeArguments: ['u8', Balance.typeTag({ typeArguments: ['0x2::sui::SUI'] })],
}); // '0x2::vec_map::VecMap<u8, 0x2::balance::Balance<0x2::sui::SUI>>'
```

- `typeArguments` is the **full positional list in Move declaration order** — for `Wrapper<phantom A, B, phantom C>` you pass all three. Supplied arguments are used to build the tag verbatim (the runtime checks argument arity, but does not re-verify each position against the instance's baked types).
- Each argument is a type tag string or a **`BcsType`** (its name is used). The `{ name: string }` shape was rejected because every JS function structurally matches it.
- **The compile-time surface is deliberately minimal** — a couple of small type declarations, no template-literal parsing. It enforces exactly two things: `typeArguments` is required whenever a type's name contains unfilled phantom parameters (including holes nested inside factory-instantiated generics), and arguments must be strings or `BcsType`s. Everything else — argument arity, smuggled unfilled holes, non-Move names like enum-variant structs — is validated at **runtime** with descriptive errors, which applies equally to literal-typed callers, dynamic callers, and previously generated code.
- Address-only output is fully validated with `TypeTagSerializer`; the result may contain MVR names, which are valid in transaction `typeArguments` (resolved by the named-packages plugin at build time).

### `resolveTypeTag({ client, ... })`

Builds the tag, resolves MVR names through `client.core.mvr.resolveType` (using the client's configured overrides and the MVR API, which return defining-package-id forms), and returns the normalized address-only tag — the correct form for query filters and comparisons against on-chain data. Configuration lives where it already lives for transactions: on the client.

### Follow-ups (not in this PR)

- Type resolution in `MvrClient` only consults per-type overrides; a package-level override does not help resolve types, and defining ids for types introduced in upgrades can differ per type within one package. A follow-up could fall back to `getMovePackage` (which returns `definingId` per datatype) so a single configured package id resolves all of a package's types exactly — benefiting transactions with `@local-pkg` type arguments as well.
- `includePhantomTypeParameters` could later be extended to accept tag strings (not just BcsTypes) in factory signatures, giving instantiated instances fully concrete names and zero-option `typeTag()` calls.

---

## 2. Dependency upgrade

Bumps dependencies across the workspace (TypeScript 6, Next.js 16, vitest 4, fumadocs 16, esbuild 0.28, react 19.2, etc.) with a minor changeset across the published SDK packages. **`engines.node` is unchanged** — `@mysten/sui` and `@mysten/deepbook-v3` already require `>=22` — so consumer runtime requirements are not affected.

Fixes for breakage the upgrade surfaced:

- **create-dapp**: move `tsconfig` off the deprecated `node10` module resolution to `NodeNext` (still CJS output) and add `types: ["node"]` for node globals (TS 6 removed the legacy resolver).
- **enoki-connect & walletconnect demo-dapps**: add `/// <reference types="vite/client" />` so CSS side-effect imports type-check (TS now checks side-effect imports).
- **docs**: rename `middleware.ts` → `proxy.ts` (Next 16 convention) and pin `turbopack.root` so the workspace root isn't inferred from a stray lockfile.
- **deepbook-v3**: `test` script `vitest` → `vitest run` (bare `vitest` hung in watch mode under turbo).
- **prettier**: ignore generated `docs/` in each package's `.prettierignore` (per-package prettier runs don't read the root ignore).
- **changesets CI**: bump the workflow from node 18 → 24 (the upgraded changesets CLI needs node 20+; `human-id` v4 is ESM-only).
- **audit / pnpm overrides**: trimmed to the minimal set addressing current prod findings (`axios`, `postcss`, scoped `ws` and `brace-expansion`); dropped stale/harmful ones (`@grpc/grpc-js` absent; `lodash`/`hono` resolve clean; the `protobufjs` and `vm2` overrides were *forcing vulnerable versions*). `pnpm audit --prod` is clean and the `--prod --audit-level high` gate passes.

---

## Test plan

- New `packages/codegen/tests/type-tag.test.ts` and `type-tag-codegen.test.ts`:
  - runtime suite covering fill/arity errors, mixed phantom/non-phantom positional semantics, nested holes, smuggled-hole rejection, verbatim type-argument passthrough, package override, BCS-type arguments, rejection of functions/non-BCS objects and non-Move names (enum variant structs, `bcs.string()`), nested `typeTag` composition, and `resolveTypeTag` normalization with a mocked client
  - an end-to-end suite that renders real codegen output for phantom/generic fixtures and calls `typeTag`/`resolveTypeTag` on the actual generated exports
  - compile-time fixture typechecked with `tsc --strict` asserting required `typeArguments` for phantom types (including nested holes), `@ts-expect-error` on function/named-object arguments and missing client, and graceful widening for dynamic callers
- Existing `generated-utils-typecheck.test.ts` passes (template compiles under `strict` + `noUncheckedIndexedAccess`)
- All 125 codegen tests pass; all six regenerated consumer packages (`walrus`, `suins`, `kiosk`, `deepbook-v3`, `payment-kit`, `pas`) build
- `pnpm audit --prod` clean; docs production build (`typedoc && next build && build:docs`) passes

---

### AI Assistance Notice


hayes-mysten:
@clud-bot review

PR #1091, opened 2026-06-15
transactions: support custom intents in Transaction.from

### LINKED ISSUE #1090: Request: Transaction.from supporting custom intents
## Problem

Currently, to copy a non-built transaction, you must call `Transaction.from`. It performs a deep copy of the transaction but handles only the `CoinWithBalance` intent. If any other intent is used inside the transaction, an error will be thrown. To copy the transaction with custom intents, the async `prepareForSerialization` must be called before creating the copy.
As far as I can see from the code, the only thing that should be performed regarding intents to make them copied in the new transaction is to add intent resolvers. At least, that's what is done for the `CoinWithBalance` intent.

## Proposal

`Transaction.from` would be able to receive an `options` argument, which may contain an intent-name-to-resolver map that should unblock fast transaction copying, even if it contains custom intents.

Please let me know if I missed anything. This feature will help us a lot, thanks!

hayes-mysten:
## Description

Resolves #1090.

Today, `Transaction.from` can only copy a non-built transaction synchronously if the only intent it contains is the built-in `CoinWithBalance` intent. Any other (custom) intent causes `Transaction.from` to throw, forcing callers to `await prepareForSerialization()` before copying — which is async and, for some intents, requires a client.

This PR adds an optional `options` argument to `Transaction.from` with an `intentResolvers` map (intent name → resolver), as proposed in the issue:

```ts
const copy = Transaction.from(tx, {
  intentResolvers: {
    MyCustomIntent: resolveMyCustomIntent,
  },
});
```

The copied transaction carries the unresolved intents over as-is and re-attaches the supplied resolvers, so the intents resolve later when the copy is built or serialized — keeping copying fully synchronous.

Details:

- New exported `TransactionCopyOptions` type (from `@mysten/sui/transactions`).
- Built-in intents (e.g. `CoinWithBalance`) are still handled automatically; caller-supplied resolvers take precedence, so a built-in resolver can be overridden if ever needed.
- All provided resolvers (plus the built-in `CoinWithBalance` resolver) are registered on the copy. Registering a resolver for an intent that isn't present is harmless — `prepareForSerialization` only runs a resolver when its intent actually appears in the transaction.
- Unresolved async thunks still throw, since they can't be carried across a synchronous copy. The error message now also points at the new `intentResolvers` option.
- Updated the error message, `plugins.mdx` docs (new "Copying transactions that use custom intents" section), and added a `minor` changeset.

## Test plan

- Added unit tests in `packages/sui/test/unit/transactions/Transaction.test.ts` covering: throwing when a custom intent has no resolver, successfully copying + resolving a custom intent when `intentResolvers` is provided (no client needed), accepting resolvers for intents not present without affecting serialization, and a regression check that the built-in `CoinWithBalance` path still copies without extra options.
- `pnpm --filter @mysten/sui vitest run test/unit/transactions` — 85 passing.
- `pnpm turbo build --filter=@mysten/sui` (runs `tsc --noEmit`), `pnpm --filter @mysten/sui test:typecheck`, prettier, and oxlint all pass.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/transaction-from-custom-intents.md
+++ b/.changeset/transaction-from-custom-intents.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': minor
+---
+
+`Transaction.from` now accepts an optional `intentResolvers` option, a map of intent names to resolvers. This lets you synchronously copy a transaction that still contains unresolved custom intents without first awaiting `prepareForSerialization`. Built-in intents (such as `CoinWithBalance`) continue to be handled automatically.
--- a/packages/docs/content/sui/plugins.mdx
+++ b/packages/docs/content/sui/plugins.mdx
@@ -274,3 +274,28 @@ function transferToSender(objects: TransactionObjectInput[]) {
 const transaction = new Transaction();
 transaction.add(transferToSender(['0x1234']));
 ```
+
+## Copying transactions that use custom intents
+
+`Transaction.from` creates a deep copy of a transaction. Copying is synchronous, so any unresolved
+intents are carried over as-is and resolved later when the copy is built or serialized. To do that,
+the copy needs the intent resolvers, but resolvers are functions and cannot be represented in the
+serialized transaction data.
+
+Built-in intents (such as the `CoinWithBalance` intent added by `tx.coin()`) are re-attached
+automatically. For custom intents, pass the resolvers via the `intentResolvers` option so the copy
+knows how to resolve them:
+
+```typescript
+import { Transaction } from '@mysten/sui/transactions';
+
+const copy = Transaction.from(transaction, {
+	intentResolvers: {
+		TransferToSender: resolveTransferToSender,
+	},
+});
+```
+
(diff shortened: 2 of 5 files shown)


hayes-mysten:
@clud-bot review

jessiemongeon1:
## Style Guide Audit

Audited **1** file(s) against the [Sui Documentation Style Guide](https://docs.sui.io/references/contribute/style-guide).

**2** violation(s) found. All must be fixed before merge.

### `packages/docs/content/sui/plugins.mdx` (2 violation(s))

_2 violation(s) (1 regex, 1 claude)_

- **Line 286** — Use "through" not "via"
  - Current: `via`
  - Fix: `through`
- **Line 286** — word-preference
  - Current: `pass the resolvers via the `intentResolvers``
  - Fix: `pass the resolvers through the `intentResolvers``

---
_Automated audit using the Sui Documentation Style Guide._

PR #1100, opened 2026-06-17
fix(sponsor): unblock CI release + add sender/gas/signature validators

hayes-mysten:
## Description

Two things, both for `@mysten-incubation/sponsor`.

### 1. Unblock the CI release pipeline

The published `0.0.0` on npm is non-installable — its manifest carries literal `workspace:^` specifiers (`@mysten/wallet-sdk`, `@mysten/sui`) because it was hand-published with `npm publish`, which doesn't rewrite the workspace protocol (CI's `pnpm publish` does — confirmed `pnpm pack` rewrites them to `^0.5.0` / `^2.18.0`).

The reason it was hand-published is the real bug: the package was added with **stale devDependency ranges** that fail `pnpm manypkg check`:

- `@types/node@^25.0.8` → `^25.9.3`
- `typescript@^5.9.3` → `^6.0.3`
- `vitest@^4.0.17` → `^4.1.8`

`manypkg check` exits 1 → the "Lint, Build, and Test" job fails → **Turborepo CI on `main` has failed on every commit since the package landed**. The Release Orchestrator only dispatches per-package releases when its triggering CI run concludes `success`, so it has been **skipping — publishing nothing repo-wide**, not just sponsor.

This PR bumps the three ranges to the repo's common versions and prunes the now-orphaned `pnpm-lock.yaml` entries. With CI green again, the orchestrator will detect `0.0.1` as unpublished and dispatch `release-sponsor.yml`, which publishes a correct manifest.

> **Follow-up outside this repo** (for the maintainer): the publish step uses pure OIDC trusted publishing (no `NODE_AUTH_TOKEN`), so `@mysten-incubation/sponsor` must be registered as a trusted publisher on npm (pointing at `release-sponsor.yml`) or the CI publish will fail with an auth error. The broken `0.0.0` should also be deprecated.

### 2. Harden the sender check + add validators

- **Rename `senderIsNotSponsor()` → `validSender()`**, folding in an explicit sender-set check. The old rule had a latent hole: an unset sender (`data.sender ?? ''`) normalizes to the zero address, which doesn't equal the (set) sponsor gas owner, so a missing sender **silently passed**. `validSender()` now rejects `SENDER_UNSET` first, then `SENDER_IS_SPONSOR`. (This replaces a separate `senderIsSet()` — one explicit "valid sender" check rather than two rules.)
- **`onlyAddressBalanceGas()`** — the gas payment is empty (`[]`) so gas is paid from the sponsor's address balance, not nominated gas coins.
- **`userSignatureMatchesSender()`** — a supplied user signature is cryptographically valid over the bytes **and** resolves to the sender, checked before the sponsor co-signs rather than only at execution.

`defaults()` now bundles `validSender()` + `onlyAddressBalanceGas()` + `gasCoinNotUsed()` + `onlySenderWithdrawals()` + `simulationSucceeds()` + `boundedExpiration()`. The sponsor injects the request's user signature(s) into validation as `userSignatures` (a reserved option) so signature-aware validators can read them without the caller routing the signature through `validationOptions`.

## Test plan

- `pnpm manypkg check` → passes ("workspaces valid!")
- `pnpm --filter @mysten-incubation/sponsor test` → typecheck + 54 unit tests pass
- `pnpm --filter @mysten-incubation/sponsor run build` → builds under TS 6
- `pnpm --filter @mysten-incubation/sponsor lint` → clean
- New/updated tests cover `validSender` (unset sender, sender==sponsor, valid), `onlyAddressBalanceGas`, and `userSignatureMatchesSender` (valid, mismatch, no-signature sponsor-builds flow).

## Versioning

No changeset: `0.0.1` is already staged (via Version Packages #1098) and was never published — only the broken hand-published `0.0.0` is on npm. These changes ship as part of `0.0.1`, so no version bump is needed.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/is-valid-signature.md
+++ b/.changeset/is-valid-signature.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': minor
+---
+
+Add `isValidSignature`, `isValidPersonalMessageSignature`, and `isValidTransactionSignature` to `@mysten/sui/verify` — boolean-returning siblings of the existing `verify*` functions, taking the same arguments. They return `false` for a malformed or invalid signature (or one that doesn't match a supplied `address`) instead of throwing, while still letting a genuine environmental failure during verification (e.g. a zkLogin JWK/epoch lookup) propagate. The `verify*` functions now delegate to these.
--- a/packages/sponsor/README.md
+++ b/packages/sponsor/README.md
@@ -61,22 +61,24 @@ Because gas is the sponsor's, **a sponsored transaction must never use the gas c
 
 ## Defaults
 
-If you pass no `validate`, the sponsor runs `defaults()` — `senderIsNotSponsor()`,
-`gasCoinNotUsed()`, `onlySenderWithdrawals()`, `simulationSucceeds()`, and `boundedExpiration()`.
-Once you add your own validators they no longer run automatically; drop them back in as one entry
-(the `validate` array flattens nested arrays):
+If you pass no `validate`, the sponsor runs `defaults()` — `validSender()`,
+`onlyAddressBalanceGas()`, `gasCoinNotUsed()`, `onlySenderWithdrawals()`, `simulationSucceeds()`,
+and `boundedExpiration()`. Once you add your own validators they no longer run automatically; drop
+them back in as one entry (the `validate` array flattens nested arrays):
 
 ```ts
 createSponsor({ signer, client }); // runs defaults()
 createSponsor({ signer, client, validate: [defaults(), allowedPackages(['0xabc'])] });
 ```
 
-Each default guards something real: `senderIsNotSponsor()` and `gasCoinNotUsed()` stop a caller from
-spending the sponsor's gas coin; `onlySenderWithdrawals()` rejects any `FundsWithdrawal` input that
-isn't the sender's — including one from the sponsor's **address balance** (the same balance that
-pays gas — a direct drain the gas-coin check can't see, since the withdrawal is an _input_, not a
-command argument); `simulationSucceeds()` avoids paying for a transaction that aborts; and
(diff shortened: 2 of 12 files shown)


hayes-mysten:
@clud-bot review

hayes-mysten:
@clud-bot review again, new changes pushed

manolisliolios (approved):
lgtm

hayes-mysten:
Addressed the two correctness findings (pushed):

**1. Major — `catch {}` misclassifying analysis errors.** `userSignatureMatchesSender` no longer passes `address` into `verifyTransactionSignature` and no longer swallows errors. It now verifies the signature cryptographically (a throw — malformed/unsupported scheme/transient zkLogin lookup — propagates as `ANALYSIS_FAILED`), then compares the recovered public key's address to the sender itself. Only a *valid* signature resolving to a different address is the policy verdict `USER_SIGNATURE_MISMATCH`. New test asserts a malformed signature → `ANALYSIS_FAILED`.

**2. Minor — multi-signature consistency.** Switched from "any one matches → pass" to "**every** supplied signature must be the sender's", since `signTransaction` attaches all of them to execution. New tests cover all-sender (passes) and sender+impostor (rejected).

**3. Changeset.** Intentionally omitted — `0.0.1` is already staged (Version Packages #1098) and was never published; only the broken hand-published `0.0.0` is on npm. The Release Orchestrator publishes by "version not on npm", so `0.0.1` (with all of this) ships without a changeset; adding one would just renumber it to `0.1.0`. Happy to add a `minor` changeset if you'd prefer the bump + CHANGELOG entry.

Tests: 57 pass; build (TS 6), lint, and `manypkg check` all clean.

PR #1105, opened 2026-06-22
feat(zklogin): add transport-agnostic ZkLoginSigner; rebase EnokiKeypair onto it

hayes-mysten:
## Description

Adds a transport- and provider-agnostic **`ZkLoginSigner`** to `@mysten/sui/zklogin`, and rebases `@mysten/enoki`'s `EnokiKeypair` onto it.

### Motivation

zkLogin signatures wrap an ephemeral ed25519 signature in a proof envelope (`getZkLoginSignature`). Today the only ready-made zkLogin signer lives in `@mysten/enoki` and is coupled to the Enoki flow — anyone obtaining a proof from another prover has to hand-roll the wrapping. The [EVE Frontier wallet-core](https://github.com/evefrontier/wallet-core) team did exactly that, using a mixin to graft zkLogin wrapping onto arbitrary signer classes. This PR fills that gap in core with a composition-based signer that works for any ephemeral `Signer` (so no mixin is needed), and folds `EnokiKeypair` into it.

### What's added

- **`ZkLoginSigner`** wraps any ephemeral `Signer`. It overrides `signWithIntent` **once** — since `signTransaction`, `signPersonalMessage`, and `signAndExecuteTransaction` all funnel through it, that single override covers every path (and removes the duplicated wrapping `EnokiKeypair` had).
- Construction: `new ZkLoginSigner({ ephemeralSigner, maxEpoch, inputs, legacyAddress })`. The address is **derived from the proof** (`addressSeed` + issuer); `legacyAddress` is a **required boolean**, consistent with the rest of the zkLogin address API (`jwtToAddress`, `computeZkLoginAddress`, `toZkLoginPublicIdentifier` — all required booleans that refuse to guess). No externally-obtained address is needed.
- Following the **`MultiSigSigner`** precedent for composite signers: `sign()` throws (a raw ephemeral signature is not valid for the zkLogin address), and `getKeyScheme()` returns `'ZkLogin'`.
- Adds a read-only `legacyAddress` getter to `ZkLoginPublicIdentifier`.
- **`EnokiKeypair`** now `extends ZkLoginSigner`. It keeps resolving its public key via `EnokiPublicKey.fromProof(address, proof)` (unchanged — same legacy auto-detection and proof/address validation) and bridges that to the base `legacyAddress` flag.

### Enoki API: verified byte-identical

The Enoki rebase was checked against a reconstruction of the previous implementation, for **both a legacy and a current address** (using a leading-zero seed so the two derivations genuinely diverge). Identical in every case:

- `getPublicKey().toSuiAddress()`, `.toRawBytes()`, `.flag()`, `.legacyAddress`
- `signTransaction()` and `signPersonalMessage()` output

The constructor signature, `getPublicKey()` (still an `EnokiPublicKey`), and `EnokiPublicKey` export are unchanged. Two previously-broken inherited behaviors are corrected (no caller in the repo depends on them): `getKeyScheme()` → `'ZkLogin'` (was the ephemeral scheme), and `sign()` now throws (was returning a bare ephemeral signature that never verified against the zkLogin address).

### Non-goals

- No mutable "attach proof after construction" mode — the signer is immutable; construct it once the proof is available.
- No mixin — composition over `Signer` covers every ephemeral scheme.

## Test plan

- New unit tests in `packages/sui/test/unit/zklogin/signer.test.ts` (real proof fixture): `signTransaction` / `signPersonalMessage` produce the expected zkLogin signature; the address derives from the proof for both `legacyAddress` values; `getKeyScheme()` is `'ZkLogin'`; `sign()` throws. 6 passing.
- Old-vs-new `EnokiKeypair` parity verified byte-for-byte (see above) for legacy and current addresses.
- `turbo build` for `@mysten/sui` and `@mysten/enoki` pass (typecheck clean); prettier + oxlint clean.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/enoki-zklogin-signer.md
+++ b/.changeset/enoki-zklogin-signer.md
@@ -1,5 +1,5 @@
 ---
-'@mysten/enoki': patch
+'@mysten/enoki': minor
 ---
 
 `EnokiKeypair` now extends the shared `ZkLoginSigner` from `@mysten/sui/zklogin`, removing duplicated
--- a/.changeset/enoki-zklogin-signer.md
+++ b/.changeset/enoki-zklogin-signer.md
@@ -1,0 +1,9 @@
+---
+'@mysten/enoki': patch
+---
+
+`EnokiKeypair` now extends the shared `ZkLoginSigner` from `@mysten/sui/zklogin`, removing duplicated
+signature-wrapping logic. `signTransaction` / `signPersonalMessage` / `getPublicKey` are unchanged.
+Two behaviors are corrected to match `ZkLoginSigner`: `getKeyScheme()` now returns `'ZkLogin'`
+(previously the ephemeral key's scheme), and `sign()` now throws instead of returning a bare
+ephemeral signature (which was never valid for the zkLogin address).
--- a/.changeset/zklogin-signer.md
+++ b/.changeset/zklogin-signer.md
@@ -6,7 +6,9 @@ Add `ZkLoginSigner`, a transport- and provider-agnostic zkLogin signer. It wraps
 `Signer` and transforms its signatures into zkLogin signatures using the supplied proof `inputs` and
 `maxEpoch`: `new ZkLoginSigner({ ephemeralSigner, maxEpoch, inputs, legacyAddress })`. The address is
 derived from the proof; `legacyAddress` is a required boolean (consistent with `jwtToAddress`,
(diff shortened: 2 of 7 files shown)


hayes-mysten:
@clud-bot review

hayes-mysten:
@clud-bot review

PR #1118, opened 2026-06-25
feat(aws-kms-signer): Ed25519 support + pluggable credential provider

### LINKED ISSUE #1111: Feature: add Ed25519 support to @mysten/aws-kms-signer
## Summary

`@mysten/aws-kms-signer` currently supports only `Secp256k1` (`ECC_SECG_P256K1`) and
`Secp256r1` (`ECC_NIST_P256`) KMS keys. AWS KMS also supports `ECC_NIST_EDWARDS25519`,
and Sui's **native** signature scheme is Ed25519 — so the one scheme most Sui accounts
use can't be backed by an `AwsKmsSigner` today.

`AwsKmsClient.getPublicKey` throws `Unsupported key spec: ECC_NIST_EDWARDS25519`, and
`AwsKmsSigner.sign` is hardcoded to `ECDSA_SHA_256`.

## Motivation

Ed25519 is already first-class everywhere else in the SDK (keypairs, `@mysten/signers/ledger`,
etc.). The KMS signer is the only place where a Sui-native Ed25519 key can't be used, which
forces teams running KMS-backed automation on Sui to either dr
adopt a non-native curve.

## Proposed change

A small, additive change (no breaking changes — `minor`):

1. `AwsKmsClient.getPublicKey`: handle the `ECC_NIST_EDWARDS25519` key spec and return an
   `Ed25519PublicKey` (raw 32 bytes parsed from the SPKI DER).
2. `AwsKmsSigner.sign`: when the key scheme is `ED25519`, sign with `ED25519_SHA_512` and
   return the raw 64-byte signature directly — Ed25519 needs n
   normalization (those stay on the ECDSA path).

`getKeyScheme()` already resolves correctly from the public ked there.

## Reference implementation

I've implemented this on a branch:
https://github.com/0xkurious/ts-sdks/commit/6971ae7

It includes the two source changes, an SPKI parser (`publicKeyFromEd25519DER`), a changeset,
and unit tests for the parser.

## Verification

Tested end-to-end against a real AWS KMS `ECC_NIST_EDWARDS25519` key:
- `getPublicKey` returns the correct `Ed25519PublicKey` and des
- `signPersonalMessage` produces a signature that verifies against the public key
- A real testnet transaction signed by the `AwsKmsSigner` was
  `BAxdpeQahWoASnWHmLzKQozT9fAwzYL6mgQboRuccb1W`

## Note

I can't open a PR directly (the repo restricts PRs to collaborators), but the branch above is
ready if a maintainer wants to pull it in or grant access.

### LINKED ISSUE #1112: Feature: support a pluggable credential provider in @mysten/aws-kms-signer
## Summary

> While working on Ed25519 support for this package [issue #1111](https://github.com/MystenLabs/ts-sdks/issues/1111) I also ran into this — it's an
> independent change (separate branch off `main`), filing it separately so each can be
> evaluated on its own.

`AwsKmsClient` requires static `accessKeyId` + `secretAccessKey` at construction and offers
no way to plug in a credential resolver. As a result the signer can't use the standard AWS
credential provider chain (SSO, IAM roles, container/instance metadata, `credential_process`),
and temporary credentials can't refresh — once the underlying session expires, every
subsequent `sign()` fails and the only recovery is to rebuild the signer with new credentials.

## Root cause

Credentials are captured as immutable values at construction and never re-resolved per request:

- `AwsKmsClient` throws unless static `accessKeyId`/`secretAccessKey` are supplied.
- The underlying `aws4fetch` `AwsClient` stores them as fixed `)
  and reads those same fields for every signature. There is nore
  in the path (a grep for `refresh|credentialProvider|expirati

So both gaps — no provider chain **and** no refresh — share onn
credentials.

## Why this matters

- **No IAM roles**: the dominant production deployment model (
  profiles) can't be used without manually resolving STS crede
- **No SSO / `credential_process`**: doesn't integrate with th
  practice, an SSO session yields no static keys to pass at allly.
- **No refresh**: a long-lived signer (a server signing over h
  once the temporary session expires and starts failing.

## What I'm *not* proposing

Not "drop `aws4fetch` for `@aws-sdk/client-kms`." That choice `
is tiny and runs on `fetch` + Web Crypto in edge/browser/Workers runtimes where the full AWS
SDK is heavy or unavailable (consistent with `@mysten/signers/
dependency would regress those consumers. (Swapping in the SDK alone wouldn't fix refresh
either, unless credentials are re-resolved per request.)

## Proposed change (additive, non-breaking)

Accept an optional async `credentials` provider on `AwsClientO
credentials are resolved **before each request** instead of statically at construction:

```ts
import { fromNodeProviderChain } from '@aws-sdk/credential-pro

const signer = await AwsKmsSigner.fromKeyId(keyId, {
  region: 'us-east-1',
  credentials: fromNodeProviderChain(), // SSO, IAM roles, ref
});
```

- Edge/browser: keep passing static credentials (or a custom r
bundle cost.
- Server: pass fromNodeProviderChain() and opt into @aws-sdk/c
if you need it.
- The resolver is invoked per request, so provider chains (whi
expiry) handle credential refresh transparently.

Static credentials keep working unchanged, so it's fully backwards compatible.

Reference implementation

https://github.com/0xkurious/ts-sdks/commit/5d36b82d

Adds the optional credentials provider to AwsClientOptions, re
in AwsKmsClient, exports AwsCredentialIdentity / AwsCredential
changeset and unit tests (including one asserting the providerest —
the refresh behavior).

Verification

Tested end-to-end against a real AWS KMS key using fromNodeProviderChain() as the provider
(no static credentials in the code; credentials resolved via t
- signer constructed and Sui address derived
- signPersonalMessage produced a signature that verifies again
- a real testnet transaction signed by the signer was accepted
7UaKCmLNPaMLR9GRUFyQboso8b2HGpkxbzkEUD5syxsW

hayes-mysten:
## Description

Adds two additive, non-breaking features to `@mysten/aws-kms-signer`, both originally
implemented and verified end-to-end by **@0xkurious (Antonio Jiménez)** and filed as issues
(the repo restricts PRs to collaborators, so the work is pulled in here on their behalf).

Closes #1111 — **Ed25519 support**
- `AwsKmsClient.getPublicKey` now recognizes the `ECC_NIST_EDWARDS25519` key spec and returns an
  `Ed25519PublicKey` (raw 32 bytes parsed from the SPKI DER via a new `publicKeyFromEd25519DER`).
- `AwsKmsSigner.sign` signs Ed25519 keys with `ED25519_SHA_512` (`MessageType: RAW`) and returns
  the raw 64-byte signature, skipping the DER parsing / low-S normalization used for the ECDSA
  curves. This unblocks Sui's native Ed25519 scheme, previously unsupported.
- Grounded in a real AWS capability — KMS added EdDSA / `ECC_NIST_EDWARDS25519` support in Nov 2025.

Closes #1112 — **Pluggable credential provider**
- `AwsClientOptions` accepts an optional async `credentials` provider (structurally compatible
  with `@aws-sdk/credential-providers`, e.g. `fromNodeProviderChain()`). When supplied,
  credentials are resolved per request, enabling the standard AWS provider chain (SSO, IAM roles,
  container/instance metadata) and automatic refresh of temporary credentials.
- Static `accessKeyId`/`secretAccessKey` continue to work unchanged (fully backwards compatible).
- New `AwsCredentialIdentity` / `AwsCredentialProvider` types are exported.
- The `@aws-sdk/credential-providers` dependency is intentionally **not** added, to preserve the
  package's edge/browser-light footprint — consumers bring their own provider.

### Review notes / changes from the original branches

I reviewed every byte of both contributed commits for correctness and safety (no exfiltration,
no out-of-scope changes, DER constants and test vectors verified). One fix was added on top:

- **`fix: resolve credentials per-request`** — the original provider implementation stored
  resolved credentials on mutable `AwsClient` instance fields before each request. Concurrent
  `sign()` calls sharing one client could interleave between resolving and signing, so a request
  could be signed with another in-flight request's credentials. Credentials are now passed to
  `aws4fetch` per-request via its `aws` override, keeping each request isolated. A regression test
  asserts concurrent requests are each signed with their own credentials.

## Test plan

- `pnpm --filter @mysten/aws-kms-signer test` — 10 unit tests pass (Ed25519 SPKI parser + round-trip,
  credential provider resolution/back-compat/per-request isolation).
- `pnpm --filter @mysten/aws-kms-signer build` (tsc `--noEmit` + tsdown), oxlint, and prettier all clean.
- **Original author verification** (per the issues): both features were tested end-to-end against
  real AWS KMS keys, with `signPersonalMessage` signatures verifying and real testnet transactions
  accepted on-chain.
- Not covered here: an automated e2e run against a real `ECC_NIST_EDWARDS25519` KMS key (requires
  live AWS credentials). Worth a manual confirmation before/at release.

---

### AI Assistance Notice

> Note: the two feature commits were authored by the external contributor **@0xkurious**; AI
> (Claude Code) reviewed every byte, bundled them into this branch, and added the per-request
> credential fix and its regression test.

### DIFF
--- a/packages/signers/aws/src/aws-kms-signer.ts
+++ b/packages/signers/aws/src/aws-kms-signer.ts
@@ -50,7 +50,7 @@ export class AwsKmsSigner extends Signer {
 
 	/**
 	 * Retrieves the key scheme used by this signer.
-	 * @returns AWS supports only Secp256k1 and Secp256r1 schemes.
+	 * @returns The scheme derived from the public key — `Ed25519`, `Secp256k1`, or `Secp256r1`.
 	 */
 	getKeyScheme() {
 		return SIGNATURE_FLAG_TO_SCHEME[this.#publicKey.flag() as SignatureFlag];
@@ -58,7 +58,7 @@ export class AwsKmsSigner extends Signer {
 
 	/**
 	 * Retrieves the public key associated with this signer.
-	 * @returns The Secp256k1PublicKey instance.
+	 * @returns The public key instance (`Ed25519PublicKey`, `Secp256k1PublicKey`, or `Secp256r1PublicKey`).
 	 * @throws Will throw an error if the public key has not been initialized.
 	 */
 	getPublicKey() {
@@ -72,6 +72,21 @@ export class AwsKmsSigner extends Signer {
 	 * @throws Will throw an error if the public key is not initialized or if signing fails.
 	 */
 	async sign(bytes: Uint8Array): Promise<Uint8Array<ArrayBuffer>> {
+		const keyScheme = this.getKeyScheme();
+
+		// Ed25519 keys use the EdDSA signing algorithm and return a raw 64-byte signature,
+		// so no DER parsing or low-S normalization (required for the ECDSA curves) is needed.
+		if (keyScheme === 'ED25519') {
+			const signResponse = await this.#client.runCommand('Sign', {
+				KeyId: this.#kmsKeyId,
+				Message: toBase64(bytes),
+				MessageType: 'RAW',
+				SigningAlgorithm: 'ED25519_SHA_512',
+			});
+
+			return fromBase64(signResponse.Signature) as Uint8Array<ArrayBuffer>;
+		}
+
 		const signResponse = await this.#client.runCommand('Sign', {
 			KeyId: this.#kmsKeyId,
 			Message: toBase64(bytes),
@@ -80,7 +95,7 @@ export class AwsKmsSigner extends Signer {
 		});
 
 		// Concatenate the signature components into a compact form
-		return getConcatenatedSignature(fromBase64(signResponse.Signature), this.getKeyScheme());
+		return getConcatenatedSignature(fromBase64(signResponse.Signature), keyScheme);
 	}
 
 	/**
--- a/packages/signers/aws/src/aws-kms-signer.ts
+++ b/packages/signers/aws/src/aws-kms-signer.ts
@@ -9,6 +9,36 @@ import { AwsKmsClient } from './aws-client.js';
 import { getConcatenatedSignature } from './utils.js';
 
 /**
+ * Maps each supported key scheme to its AWS KMS signing algorithm and the function that
+ * converts the raw KMS response into the compact signature Sui expects.
+ *
+ * - Ed25519 keys use EdDSA and return a raw 64-byte signature, so no DER parsing or low-S
+ *   normalization is needed.
+ * - The ECDSA curves return a DER-encoded signature that is parsed, low-S normalized, and
+ *   concatenated into compact form by `getConcatenatedSignature`.
+ */
+const SIGNING_ALGORITHMS: Record<
+	string,
+	{
+		algorithm: 'ED25519_SHA_512' | 'ECDSA_SHA_256';
+		format: (signature: Uint8Array) => Uint8Array<ArrayBuffer>;
+	}
+> = {
+	ED25519: {
+		algorithm: 'ED25519_SHA_512',
+		format: (signature) => signature as Uint8Array<ArrayBuffer>,
+	},
+	Secp256k1: {
+		algorithm: 'ECDSA_SHA_256',
+		format: (signature) => getConcatenatedSignature(signature, 'Secp256k1'),
+	},
+	Secp256r1: {
+		algorithm: 'ECDSA_SHA_256',
+		format: (signature) => getConcatenatedSignature(signature, 'Secp256r1'),
+	},
+};
+
+/**
  * Configuration options for initializing the AwsKmsSigner.
  */
 export interface AwsKmsSignerOptions {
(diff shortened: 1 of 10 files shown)


jessiemongeon1:
## Style Guide Audit

All 1 file(s) pass the style guide audit.

packages/signers/aws/src/aws-kms-signer.ts
```diff
@@ -72,6 +72,21 @@ export class AwsKmsSigner extends Signer {
 	 * @throws Will throw an error if the public key is not initialized or if signing fails.
 	 */
 	async sign(bytes: Uint8Array): Promise<Uint8Array<ArrayBuffer>> {
+		const keyScheme = this.getKeyScheme();
+
+		// Ed25519 keys use the EdDSA signing algorithm and return a raw 64-byte signature,
+		// so no DER parsing or low-S normalization (required for the ECDSA curves) is needed.
+		if (keyScheme === 'ED25519') {
```
joyqvq: this if is an error prone / ugly sign path. 

perhaps move this into a table so its mapped to a sig format and leave the sign function untouched? this way you can avoid the nonobvious switch on the algo 				SigningAlgorithm: 'ED25519_SHA_512', vs ECDSA_SHA_256

e.g.
```
const SIGNING: Record<
      string,
      {
              algorithm: 'ECDSA_SHA_256' | 'ED25519_SHA_512';
              format: (sig: Uint8Array) => Uint8Array<ArrayBuffer>;
      }
> = {
      // Ed25519 returns a raw 64-byte signature — no DER parsing or low-S normalization.
      ED25519: { algorithm: 'ED25519_SHA_512', format: (sig) => sig as Uint8Array<ArrayBuffer> },
      Secp256k1: { algorithm: 'ECDSA_SHA_256', format: (sig) => getConcatenatedSignature(sig, 'Secp256k1'
      Secp256r1: { algorithm: 'ECDSA_SHA_256', format: (sig) => getConcatenatedSignature(sig, 'Secp256r1'
};

async sign(bytes: Uint8Array): Promise<Uint8Array<ArrayBuffer>> {
      const { algorithm, format } = SIGNING[this.getKeyScheme()];

      const signResponse = await this.#client.runCommand('Sign', {
              KeyId: this.#kmsKeyId,
              Message: toBase64(bytes),
              MessageType: 'RAW',
              SigningAlgorithm: algorithm,
      });

      return format(fromBase64(signResponse.Signature));
}
```

packages/signers/aws/tests/credential-provider.test.ts
```diff
@@ -0,0 +1,103 @@
...
+		const client = new AwsKmsClient({
+			region: 'us-east-1',
+			accessKeyId: 'AKID',
+			secretAccessKey: 'SECRET',
+		});
+
+		await client.runCommand('GetPublicKey', { KeyId: 'k' });
+		expect(fetchMock).toHaveBeenCalledTimes(1);
+	});
+});
```
joyqvq, line 103: i dont see tests added for actual signing, add?

PR #1140, opened 2026-07-15
Migrate Hashi SDK into the monorepo as @mysten/hashi

Tzal3x:
## Description

Migrates the Hashi TypeScript SDK (native BTC collateralization on Sui) from the standalone
[hashi-ts-sdk](https://github.com/MystenLabs/hashi-ts-sdk) repo into `packages/hashi`, renaming it
from `@mysten-incubation/hashi` (v0.5.0 on npm) to `@mysten/hashi` (SEDEFI-163).

- **Source, tests, and committed codegen bindings** copied from the standalone repo and reformatted
  to monorepo prettier style (tabs, single quotes, import sort) — most of the diff is this
  mechanical reformat plus the generated `src/contracts/**` bindings.
- **package.json** rewritten to monorepo conventions: `@mysten/sui` as a `workspace:^` peer
  dependency, `@mysten/codegen` as a workspace devDependency, standard build/lint/test scripts.
- **Codegen** (`sui-codegen.config.ts`) retargeted at the `../hashi` sibling checkout
  (walrus/payment-kit pattern); bindings stay committed at the revision generated from
  MystenLabs/hashi `cd2b81f`.
- **Monorepo wiring**: `release-hashi.yml` workflow (OIDC trusted publishing via
  `_release-package.yml`), oxlint ignore for `packages/hashi/src/contracts`,
  `linguist-generated` for `src/contracts/**`, `AGENTS.md` sibling-repo/codegen entries, and
  minor changesets — so the first publish under the new name is 0.6.0 with changelog entries
  explaining the rename and the breaking API change below.
- **CI**: unit tests run via the standard `test` task; `hashi-ci.yml` ports the standalone repo's
  localnet integration harness — it clones `MystenLabs/hashi` (pinned to `cd2b81f`), builds the
  `hashi-localnet` Rust CLI, installs a version-pinned, SHA256-verified Bitcoin Core, and runs the
  env-gated integration tests (`HASHI_E2E_*`) against a live localnet.
- **Test type fixes**: the new `test:typecheck` step surfaced latent type errors the standalone
  repo never caught (it never ran `tsc` over tests): unnarrowed `TransactionHistoryItem` union
  access in three integration tests, and a parse-only `$kind` tag in a BCS enum serialize input.
- Docs are README-only for now (the standalone repo's comprehensive guide travels as the package
  README); a docs-site section can follow at graduation.

### Review follow-ups (post-migration commits)

API changes from review:

- **Breaking**: `hashi()` no longer takes a `network` option — the network is derived from the Sui
  client being extended (`.$extend(hashi())`), still throwing for networks without a
  `NETWORK_CONFIG` entry unless custom `hashiObjectId`/`packageId` are passed.
- `HashiClientOptions<Name>` defaults to `string` instead of a wrong `'HashiClient'` literal.

Internal cleanups from review (no API change):

- Withdrawal building uses `tx.balance(...)` instead of hand-composing `coinWithBalance` +
  `0x2::coin::into_balance`.
- Gas estimators reuse the real `tx.deposit`/`tx.requestWithdrawal` builders instead of
  duplicating the PTBs, so estimates can't drift from what users sign.
- Dropped the manual 500-object chunking (`core.getObjects` batches internally on all three
  transports) and replaced the raw string-interpolated GraphQL fetch with a `SuiGraphQLClient`
  query using proper GraphQL variables.
- `rpcCall<T>` is generic instead of casting at each Bitcoin RPC call site.

## Test plan

- `pnpm turbo build --filter @mysten/hashi` — typecheck + tsdown pass against workspace
  `@mysten/sui` 2.20.x (verifies the peer-range move from `^2.14.1`)
- `pnpm --filter @mysten/hashi test` — test-dir typecheck + 185 unit tests pass
- `pnpm turbo lint --filter @mysten/hashi` + root oxlint — clean (0 findings on ported code)
- `pnpm changeset status` — `@mysten/hashi` queued for a minor bump (0.6.0)
- Codegen exercised end-to-end against the `../hashi` sibling checkout (regenerates successfully;
  regenerated output not committed since that checkout is ahead of the pinned contract revision)
- `Localnet Integration` CI job green — full deposit/withdrawal integration flow against a live
  localnet with pinned Bitcoin Core

Post-merge follow-ups (outside this PR): dispatch `release-hashi.yml` (dry-run first), deprecate
`@mysten-incubation/hashi` on npm, archive the hashi-ts-sdk repo.

---

### AI Assistance Notice

### DIFF
--- a/.github/workflows/hashi-ci.yml
+++ b/.github/workflows/hashi-ci.yml
@@ -87,29 +87,21 @@ jobs:
         run: bash hashi/.github/scripts/install-sui.sh
 
       # Mirrors the "Install Bitcoin Core" step in
-      # hashi/.github/workflows/ci.yml. Vendored verbatim because hashi-core
-      # has it inline (no install-bitcoin.sh to reuse). Revisit if hashi
-      # extracts a script.
+      # hashi/.github/workflows/ci.yml, but pinned rather than tracking the
+      # latest upstream release: an unpinned bitcoind would let a new Bitcoin
+      # Core release change what these tests run against with no PR, and a
+      # floating version can't be checksum-verified. Bump both values together
+      # (checksums: https://bitcoincore.org/bin/bitcoin-core-<version>/SHA256SUMS).
       - name: Install Bitcoin Core
         env:
-          FALLBACK_VERSION: '29.1'
+          BITCOIN_VERSION: '29.1'
+          BITCOIN_SHA256: '2dddeaa8c0626ec446b6f21b64c0f3565a1e7e67ff0b586d25043cbd686c9455'
         run: |
-          echo "Detecting latest Bitcoin Core version..."
-          BITCOIN_VERSION=$(curl -s https://api.github.com/repos/bitcoin/bitcoin/releases/latest | \
-            sed -n 's/.*"tag_name": "v\([^"]*\)".*/\1/p' || echo "$FALLBACK_VERSION")
-
-          if [ -z "$BITCOIN_VERSION" ]; then
-            echo "Failed to detect version, using fallback: $FALLBACK_VERSION"
-            BITCOIN_VERSION="$FALLBACK_VERSION"
-          fi
-
           echo "Installing Bitcoin Core v${BITCOIN_VERSION}..."
 
-          wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz || {
-            echo "Download failed for ${BITCOIN_VERSION}; falling back to ${FALLBACK_VERSION}"
-            BITCOIN_VERSION="$FALLBACK_VERSION"
-            wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
-          }
+          wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
+
+          echo "${BITCOIN_SHA256}  bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz" | sha256sum -c -
 
           tar -xzf bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
           sudo mv bitcoin-${BITCOIN_VERSION}/bin/bitcoind /usr/local/bin/
--- a/.github/workflows/hashi-ci.yml
+++ b/.github/workflows/hashi-ci.yml
@@ -1,0 +1,309 @@
+name: Hashi Integration Tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths:
+      - 'packages/hashi/**'
+      - 'pnpm-lock.yaml'
+      - '.github/workflows/hashi-ci.yml'
+  pull_request:
+    paths:
+      - 'packages/hashi/**'
+      - 'pnpm-lock.yaml'
+      - '.github/workflows/hashi-ci.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  localnet-integration:
+    name: Localnet Integration
+    runs-on: ubuntu-latest
+    timeout-minutes: 35
+    env:
+      # Pin the hashi checkout to the commit the committed SDK bindings in
+      # `packages/hashi/src/contracts` were generated from. Unlike the
+      # standalone hashi-ts-sdk repo, ts-sdks does NOT vendor hashi as a
+      # submodule, so we clone it explicitly below. Keep this in sync with
+      # the codegen revision when the bindings are regenerated.
+      HASHI_PIN: cd2b81f9a208f5077032110652f1e8bba297278c
+      LOCALNET_DATA_DIR: ${{ github.workspace }}/.hashi/localnet
+      LOCALNET_LOG: ${{ github.workspace }}/localnet.log
+    steps:
+      - name: Checkout ts-sdks
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # pin@v4
+
+      # ts-sdks has no hashi submodule (../hashi is only a codegen sibling),
+      # so pull the matching hashi revision into ./hashi. `pnpm-workspace.yaml`
+      # only globs `packages/**`, so this checkout is invisible to pnpm.
+      - name: Checkout hashi (localnet harness + contracts)
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # pin@v4
+        with:
+          repository: MystenLabs/hashi
+          ref: ${{ env.HASHI_PIN }}
+          path: hashi
+          submodules: recursive
+
+      - name: Record hashi pin in step summary
+        run: |
+          PIN=$(git -C hashi rev-parse HEAD)
+          echo "Hashi pin: \`$PIN\`" >> "$GITHUB_STEP_SUMMARY"
+
+      - uses: pnpm/action-setup@b906affcce14559ad1aafd4ab0e942779e9f58b1 # pin@v4
+        name: Install pnpm
+        with:
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # pin@v4
+        with:
+          node-version: '24'
+          cache: 'pnpm'
+
+      # `hashi/rust-toolchain.toml` pins 1.91 — match it so cargo doesn't
+      # auto-install a second toolchain inside the build step (slower and
+      # uncached).
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # pin@stable
+        with:
+          toolchain: '1.91'
+
+      - uses: Swatinem/rust-cache@23869a5bd66c73db3c0ac40331f3206eb23791dc # pin@v2.9.1
+        with:
+          workspaces: hashi -> hashi/target
+          # Bumping the pin should bust the cache. `Swatinem/rust-cache`
+          # already keys on Cargo.lock; the explicit hashi SHA in the key
+          # makes the relationship unambiguous in cache hit/miss logs.
+          shared-key: hashi-${{ env.HASHI_PIN }}
+
+      - name: Install Sui binary
+        # install-sui.sh queries the GitHub releases API with GITHUB_TOKEN
+        # (set -u, so it must be set) — mirrors the env in hashi's own ci.yml.
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: bash hashi/.github/scripts/install-sui.sh
+
+      # Mirrors the "Install Bitcoin Core" step in
+      # hashi/.github/workflows/ci.yml. Vendored verbatim because hashi-core
+      # has it inline (no install-bitcoin.sh to reuse). Revisit if hashi
+      # extracts a script.
+      - name: Install Bitcoin Core
+        env:
+          FALLBACK_VERSION: '29.1'
+        run: |
+          echo "Detecting latest Bitcoin Core version..."
+          BITCOIN_VERSION=$(curl -s https://api.github.com/repos/bitcoin/bitcoin/releases/latest | \
+            sed -n 's/.*"tag_name": "v\([^"]*\)".*/\1/p' || echo "$FALLBACK_VERSION")
+
+          if [ -z "$BITCOIN_VERSION" ]; then
+            echo "Failed to detect version, using fallback: $FALLBACK_VERSION"
+            BITCOIN_VERSION="$FALLBACK_VERSION"
+          fi
+
+          echo "Installing Bitcoin Core v${BITCOIN_VERSION}..."
+
+          wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz || {
+            echo "Download failed for ${BITCOIN_VERSION}; falling back to ${FALLBACK_VERSION}"
+            BITCOIN_VERSION="$FALLBACK_VERSION"
+            wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
+          }
+
+          tar -xzf bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz
+          sudo mv bitcoin-${BITCOIN_VERSION}/bin/bitcoind /usr/local/bin/
+          sudo mv bitcoin-${BITCOIN_VERSION}/bin/bitcoin-cli /usr/local/bin/
+          rm -rf bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz bitcoin-${BITCOIN_VERSION}/
+
+          bitcoind --version | head -n1
+
+      - name: Build hashi-localnet binary
+        working-directory: hashi
+        run: cargo build --release --bin hashi-localnet -p e2e-tests
+
+      - name: Start hashi-localnet (background)
+        env:
+          NUM_VALIDATORS: 4
+        run: |
+          mkdir -p "$LOCALNET_DATA_DIR"
+          # 4 validators is the right floor: on-chain `committee::verify_certificate`
+          # threshold is 2 (`MpcManager initialized` shows `threshold=2 max_faulty=1`).
+          # We tried bumping to 6 in 0d3ddf5 expecting 2/6 fault tolerance, but it
+          # made things worse — every extra validator adds a Kyoto BIP-157 client
+          # contending for the single regtest bitcoind, which amplified the peer-
+          # disconnect storm seen in run 25097936525. Recovery here is
+          # "pre-warm Kyoto in the readiness step" (next), not "more validators".
+          nohup hashi/target/release/hashi-localnet start \
+            --num-validators "$NUM_VALIDATORS" \
+            --sui-rpc-port 9000 \
+            --btc-rpc-port 18443 \
+            --data-dir "$LOCALNET_DATA_DIR" \
+            --verbose \
+            > "$LOCALNET_LOG" 2>&1 &
+          echo $! > "$LOCALNET_DATA_DIR/runner.pid"
+          echo "NUM_VALIDATORS=$NUM_VALIDATORS" >> "$GITHUB_ENV"
+          echo "hashi-localnet PID: $(cat $LOCALNET_DATA_DIR/runner.pid)"
+
+      - name: Wait for localnet readiness
+        run: |
+          set -euo pipefail
+          echo "Phase 1/3: waiting for state.json (cap 5 min)…"
+          for i in $(seq 1 60); do
+            if [[ -f "$LOCALNET_DATA_DIR/state.json" ]]; then
+              echo "state.json appeared after ~$((i * 5)) s"
+              break
+            fi
+            sleep 5
+          done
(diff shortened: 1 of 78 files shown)


Tzal3x:
The Turborepo CI failure is unrelated to this PR: the `Lint, Build, and Test` job passes, and only the `pnpm audit` job fails because npm has retired the legacy audit endpoints (they now return HTTP 410), which the pinned pnpm version still calls — a repo-wide infra issue affecting all branches, not this change.

packages/hashi/src/contracts/hashi/deps/sui/group_ops.ts
```diff
@@ -0,0 +1,15 @@
+/**************************************************************
```
hayes-mysten, line 1: these generated files should be marked as generated via linguist configs so they auto-collapse in reviews

packages/hashi/src/types.ts
```diff
@@ -0,0 +1,355 @@
...
+
+export type SuiNetwork = 'devnet' | 'testnet' | 'mainnet' | 'localnet';
+
+export interface NetworkConfig {
+	hashiObjectId: string;
+	packageId: string;
+	bitcoinNetwork: BitcoinNetwork;
+}
+
+export interface HashiClientOptions<Name = 'HashiClient'> {
```
hayes-mysten: the default Name here seems wrong, I assume this is meant to be `hashi`

I think actually `string` is probably a better default here.  If something accepts `HashiClientOptions` only `string` would allow any options object with different names set

packages/hashi/src/types.ts
```diff
@@ -0,0 +1,355 @@
...
+	name?: Name;
+	/** Sui network — determines Hashi object IDs and default Bitcoin network. */
+	network: SuiNetwork;
```
hayes-mysten: we generally don't want to pass network separately, and instead get it from the client being extended

packages/hashi/src/btc-rpc.ts
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import type { UtxoLookupResult } from './types.js';
+
+async function rpcCall(
```
hayes-mysten: I like just making these kinds of methods take a generic so you can pass in the data type rather than always casting after calling it

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+		hashiObjectId,
+		packageId,
+		bitcoinNetwork,
+		btcRpcUrl,
+		graphqlUrl,
+		guardianUrl,
+		guardianInfoProvider,
+	}: {
+		client: ClientWithCoreApi;
+		network: SuiNetwork;
```
hayes-mysten: network should always come from the client, and we should throw if its unsupported

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+			throw new Error(
+				`Hashi is not yet supported on Sui ${network}. Provide a custom hashiObjectId and packageId.`,
+			);
+		}
+		this.#client = client;
+		this.#hashiObjectId = resolvedObjectId;
+		this.#packageId = resolvedPackageId;
+		this.#bitcoinNetwork = bitcoinNetwork ?? config?.bitcoinNetwork ?? 'testnet';
+		this.#btcRpcUrl = btcRpcUrl;
+		this.#graphqlUrl = graphqlUrl ?? defaultGraphqlUrl(network);
```
hayes-mysten: why not just create a graphql client here

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+			/**
+			 * Target Bitcoin address as raw witness program bytes — 20 bytes for
+			 * P2WPKH, 32 bytes for P2TR. Callers decode their own bech32(m)
+			 * strings for now; a string-input overload may land in a follow-up.
+			 */
+			bitcoinAddress: Uint8Array;
+		}): Transaction => {
+			const tx = new Transaction();
+			const btcType = `${this.#packageId}::btc::BTC`;
+			const coin = tx.add(
```
hayes-mysten: we've added tx.coin and tx.balance (tx.coin is just an alias for coinWithBalance).

You also don't need to tx.add explicitly here you can just do `const coin = conWithBalance() || tx.coin(..)`.

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+				coinWithBalance({
+					type: btcType,
+					balance: options.amount,
+					useGasCoin: false,
+				}),
+			);
+			const [balance] = tx.moveCall({
```
hayes-mysten: if you are using a balance directly you should just use `tx.balance(...)`

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+				package: '0x2',
+				module: 'coin',
+				function: 'into_balance',
+				typeArguments: [btcType],
+				arguments: [coin],
```
hayes-mysten: intents like coins and balances can also be used dirctly in arguments (`arguments: [tx.coin()])` (splitting them out is totally fine, but if they are only being used in one place it can make it easier to read in some cases (this is very subjective and I don't have strong opinions on how to do it)

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+			const tx = new Transaction();
+			const balance = tx.add(this.call.cancelWithdrawal({ requestId: options.requestId }));
+			const [coin] = tx.moveCall({
+				package: '0x2',
+				module: 'coin',
+				function: 'from_balance',
+				typeArguments: [`${this.#packageId}::btc::BTC`],
+				arguments: [balance],
+			});
+			tx.transferObjects([coin], options.recipient);
```
hayes-mysten, line 587: I am curious about plans for if/when we start just sending users balances instead of coins

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+		 * @param sender - If provided, estimates gas cost via dry-run.
+		 */
+		withdrawalFees: async (sender?: string): Promise<WithdrawalFees> => {
+			const snap = await this.view.all();
+
+			let gasEstimateMist = 0n;
+			if (sender) {
+				const dummyAmount = snap.bitcoinWithdrawalMinimum + 1n;
+				const btcType = `${this.#packageId}::btc::BTC`;
+				const tx = new Transaction();
```
hayes-mysten: Are these methods just building exactly the same transactions as the `client.tx` transactions?  This seems like we should be reusing those to get accurate gas estimates that stay in sync without manually keeping them in sync with changes

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+	 * Resolve the Bag id holding a user's request IDs from `user_requests`,
+	 * or `null` if the user has never had a request.
+	 */
+	async #fetchUserRequestsBagId(tableId: string, suiAddress: string): Promise<string | null> {
+		try {
+			const { dynamicField } = await this.#client.core.getDynamicField({
+				parentId: tableId,
+				name: {
+					type: 'address',
+					bcs: bcs.Address.serialize(suiAddress).toBytes(),
```
hayes-mysten, line 1399: we added a top level toBytes method to avoid the extra chained serialize call

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+		} while (cursor);
+		return keys;
+	}
+
+	/**
+	 * Batch-fetches objects in chunks of `GET_OBJECTS_BATCH`, concatenating
+	 * the results. Each element is either an object or an `Error` (for
+	 * missing / deleted objects).
+	 */
+	async #batchGetObjects(objectIds: string[], include?: { content?: boolean }) {
```
hayes-mysten: the core getObjects API already handles batching internally

packages/hashi/src/client.ts
```diff
@@ -0,0 +1,1635 @@
...
+                events(
+                    filter: { sender: "${sender}", type: "${eventType}" }
+                    first: 50${afterClause}
+                ) {
+                    nodes { contents { json } }
+                    pageInfo { hasNextPage endCursor }
+                }
+            }`;
+
+			const res = await fetch(this.#graphqlUrl, {
```
hayes-mysten: this should use the graphql client, which will properly type all the graphql requests for you

hayes-mysten: we are also a couple weeks away from having event queries work in grpc and the core APIs

hayes-mysten (commented):
Overall this looks really good and aligns well with this repos best practices.   Code looks clean and well organized.

I've gone over everything from a best practices and sdk usage perspective and it all looks pretty solid to me.  I did not review it for correctness, edge cases, etc, and I did not really look at the tests or docs.

hayes-mysten:
@clud-bot review

packages/hashi/sui-codegen.config.ts
```diff
@@ -0,0 +1,16 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import type { SuiCodegenConfig } from '@mysten/codegen';
+
+const config: SuiCodegenConfig = {
+	output: './src/contracts',
+	packages: [
+		{
+			package: '@local-pkg/hashi', // TODO: update this when hashi is published on MVR.
+			path: '../../../hashi/packages/hashi',
```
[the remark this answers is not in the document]
Tzal3x: Skipping this one — I think the prerequisite is already documented in the place you pointed at.

`AGENTS.md` gained a `../hashi` row in its "Local Sibling Repositories" table as part of this PR:

| `../hashi` | Hashi BTC-bridge Move contracts | `@mysten/hashi` codegen |

And none of the five sibling configs (`walrus`, `payment-kit`, `deepbook-v3`, `pas`, `suins-v2`) carry an inline comment either — they all use a bare `../../../<repo>/...` path. So the established pattern is "path in the config, prose in AGENTS.md", which hashi already follows; adding a comment only here would make it the asymmetric one.

Happy to revisit as a convention, but it should land as its own change across all six configs rather than just this one.

packages/hashi/src/contracts/hashi/deps/sui/group_ops.ts
Tzal3x: Fixed in 0876689a — added `packages/hashi/src/contracts/** linguist-generated=true` to the root `.gitattributes`, alongside the existing entries for walrus, payment-kit, deepbook-v3, pas, etc. Note this only affects diff rendering on future PRs; the files will still show expanded in this one.

packages/hashi/CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Minor Changes
+
+- 75fcdca: Fix btcTxid display values to strip the 0x prefix. Add GraphQL-based discovery of pending deposits to transaction history — confirmed requests still read from the on-chain user_requests index; in-flight deposits are discovered via DepositRequestedEvent queries and deduplicated. Bump GET_OBJECTS_BATCH to 500.
+
+## 0.0.2
+
+### Patch Changes
+
+- 9422708: Add a package-level `README.md` so the npm landing page has a real overview (install, one quickstart snippet, link to the repo README for full docs). Also corrects stale `@mysten/hashi` references in the root README to the actual published name `@mysten/hashi`.
```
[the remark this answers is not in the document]
Tzal3x: Fixed in b68aede9. Checked the standalone repo history — the original entry read "... to the actual published name `@mysten-incubation/hashi`", and the migration ran a blanket `@mysten-incubation/hashi` → `@mysten/hashi` rename that collapsed the second occurrence. So the fix belongs on the trailing name, not the leading one; correcting the first would have inverted the historical fact. Diffed the full CHANGELOG against the original and this was the only line the rename damaged.

packages/hashi/test/integration/_env.ts
```diff
@@ -0,0 +1,323 @@
...
+		throw new Error(
+			'No signer available: set HASHI_E2E_FUNDED_KEY_PEM (localnet, written ' +
+				'by `hashi-localnet start`) or HASHI_E2E_SUI_PRIVATE_KEY (devnet, bech32).',
+		);
+	}
+	return Ed25519Keypair.fromSecretKey(decodeSuiPrivateKey(bech32).secretKey);
+}
+
+/**
+ * Invokes the `hashi-localnet` Rust CLI from the submodule. The CI workflow
```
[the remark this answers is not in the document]
Tzal3x: Fixed in 5922f2b9 — "from the submodule" → "from the hashi repo checkout", matching `hashi-ci.yml` which clones `MystenLabs/hashi` into `hashi/` and builds the binary there. Also grepped for other stale references; the remaining "submodule" hits in the workflow are correct (two explain that ts-sdks has no submodule, one is `submodules: recursive` on the hashi checkout itself).

.github/workflows/hashi-ci.yml
```diff
@@ -0,0 +1,309 @@
...
+            sed -n 's/.*"tag_name": "v\([^"]*\)".*/\1/p' || echo "$FALLBACK_VERSION")
+
+          if [ -z "$BITCOIN_VERSION" ]; then
+            echo "Failed to detect version, using fallback: $FALLBACK_VERSION"
+            BITCOIN_VERSION="$FALLBACK_VERSION"
+          fi
+
+          echo "Installing Bitcoin Core v${BITCOIN_VERSION}..."
+
+          wget -q https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz || {
```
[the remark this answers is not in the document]
Tzal3x: Fixed in 460a075d, and went a bit further than the suggestion.

Reading the whole step, the tarball was not just unverified — the version was auto-detected from the GitHub releases API, so CI silently tracked whatever upstream tagged most recently. That is its own reproducibility problem, and it also blocks checksum pinning, since you cannot pin a hash for a version you do not know in advance.

Also worth noting: fetching SHA256SUMS from bitcoincore.org to verify a tarball from bitcoincore.org is a same-origin check — it catches truncation and network MITM, but not a compromise of the host serving both.

So the step now pins `BITCOIN_VERSION: 29.1` plus its SHA256 and runs `sha256sum -c -` before `tar`, and the auto-detect and fallback-retry are gone (net -8 lines). This also brings the step in line with the rest of the workflow, which SHA-pins actions and pins the hashi clone.

Verified both directions locally: the real tarball passes, a tampered hash exits 1 before extraction. Tradeoff is that picking up new bitcoind releases is now a manual two-line bump; the step comment says where to get the checksum.

packages/hashi/CHANGELOG.md
```diff
@@ -0,0 +1,58 @@
...
+
+## 0.5.0
+
+### Minor Changes
+
+- ff9a398: Wire up Sui testnet: add the testnet Hashi object and package ids to `NETWORK_CONFIG` (BTC signet), so `hashi({ network: "testnet" })` works out of the box
+
+### Patch Changes
+
+- ff9a398: Bump hashi submodule to cd2b81f (no contract or binding changes)
```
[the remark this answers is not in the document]
Tzal3x: Leaving this one as-is — `hashi-ts-sdk/.gitmodules` confirms the standalone repo did have a `hashi` submodule, so this historical release note was accurate when written and the monorepo layout does not apply retroactively to it. (Contrast with the `_env.ts` comment, which describes current behaviour and I did fix.) Happy to change it if you would rather changelog entries read against the current repo layout.

packages/hashi/src/types.ts
Tzal3x: Fixed in d9bd8688 — changed the default to `string`.

packages/hashi/src/btc-rpc.ts
Tzal3x: Fixed in 38b3d35a — rpcCall<T> now takes the type param instead of casting at each call site.

packages/hashi/src/client.ts
Tzal3x: I couldn't find a top-level toBytes in @mysten/bcs or @mysten/sui/bcs — BcsType has no toBytes(value), only the no-arg SerializedBcs.toBytes() after .serialize(). Checked origin/main too, no newer commit adds one. Can you point me at where it lives? Leaving this open until then.

packages/hashi/src/client.ts
Tzal3x: The teams that will be using this sdk will be defi builders. I have the impression that it would be more useful providing them with that kind of granularity. 

On the other hand sending the user balance directly seems very convenient. Maybe we could add this as extra functionality instead of _replacing_ the coins.

packages/hashi/src/types.ts
Tzal3x: Fixed in 3bf25ce8 — `network` is removed from `HashiClientOptions`; the constructor now reads `client.network` from the extended client. Added a major changeset since this changes the public API.

packages/hashi/src/client.ts
Tzal3x: Fixed in 3bf25ce8 — network now always comes from `client.network`, and the existing unsupported-network throw is kept (still bypassable with custom `hashiObjectId`/`packageId` for local deployments).

packages/hashi/src/client.ts
Tzal3x: Fixed in 82cd7467 — went one step further: since the coin was only created to feed `into_balance`, the whole pair collapses into a single `tx.balance(...)` (the intent resolver inserts the `into_balance` conversion at build time). Also applied at the gas-estimate site.

packages/hashi/src/client.ts
Tzal3x: Fixed in 82cd7467 — `tx.balance(...)` replaces the coin + `into_balance` pair, and the `coinWithBalance` import is gone.

packages/hashi/src/client.ts
Tzal3x: After 82cd7467 this is moot — the coin/balance intermediates are gone, and the remaining `balance` result feeds a generated-binding call where a named variable reads fine. Resolving alongside the other two.

packages/hashi/src/client.ts
Tzal3x: Fixed in 0942662b — `#batchGetObjects` is gone; verified all three transports chunk internally at 50 in their `getObjects`, so the local 500-chunking never changed what hit the wire anyway.

packages/hashi/src/client.ts
Tzal3x: Fixed in 0942662b — the raw fetch is replaced with a `SuiGraphQLClient` query using real GraphQL variables ($sender/$type/$after), which also removes the string-interpolation injection surface. Happy to move this to the core event API once the gRPC/core event queries land.

packages/hashi/src/client.ts
Tzal3x: Fixed in 0942662b — the constructor now builds a `SuiGraphQLClient` (from `graphqlUrl` or the network default) instead of storing a bare URL, and the events query goes through it.

packages/hashi/src/client.ts
Tzal3x: Yes, they were duplicates — fixed in a01838b6. Both estimators now call the real `tx.deposit`/`tx.requestWithdrawal` builders with dummy values, so the dry-run transaction can't drift from what users sign (the earlier `tx.balance` cleanup had to touch both copies, which proved the drift risk).

packages/hashi/src/client.ts
hayes-mysten: I really thought we has this, ignore me

hayes-mysten:
@clud-bot rereview

hayes-mysten (approved):

packages/hashi/src/client.ts
```diff
@@ -1459,37 +1443,15 @@ export class HashiClient {
...
-                    pageInfo { hasNextPage endCursor }
-                }
-            }`;
-
-			const res = await fetch(this.#graphqlUrl, {
-				method: 'POST',
-				headers: { 'Content-Type': 'application/json' },
-				body: JSON.stringify({ query }),
+			const result: GraphQLQueryResult<EventRequestIdsResult> = await this.#graphql.query({
+				query: EVENT_REQUEST_IDS_QUERY,
```
hayes-mysten, line 1448: Check out https://sdk.mystenlabs.com/sui/clients/graphql#custom-graphql-queries for the right way to use the graphql client, this avoids needing to manually type your result which is very error prone

PR #1152, opened 2026-07-20
feat(deepbook-predict): add DeepBook Predict TypeScript SDK

tonylee08:
## Description

Adds `@mysten/deepbook-predict`, the TypeScript SDK for DeepBook Predict (European cash-settled range digitals on Sui).

The SDK registers as a **client extension** — `client.$extend(predict({ network }))`, then reached at `client.predict.*` — matching the `@mysten/deepbook-v3` shape. It's **codegen-based**: transaction builders, on-chain reads, and event decoders are built on `sui-ts-codegen`-generated contract bindings (`src/contracts/**`, transaction thunks + BCS structs) rather than hand-rolled `tx.moveCall` calls and hand-written layouts. A curated `PredictClient` facade sits on top for the ergonomic surface — human-unit conversions, market resolution, dry-run quotes, chain-only position enumeration, and typed execution-receipt decoders. The SDK never signs and never touches keys: every `tx.*` returns a `Transaction`.

- **Client extension + thunks** — `predict()` registration factory; the client is typed `ClientWithCoreApi` and reads go through `client.core.*`. Builders are transaction thunks (`(cfg, args) => (tx) => result`) so they compose across SDKs in one PTB via `tx.add(...)`.
- **Codegen bindings** for predict / account / propbook, generated from the deployed Move sources. The hand layer injects the well-known objects (protocol config, pool vault, accumulator root) from `PredictConfig`; the generated layer auto-injects the `Clock`.
- **Clever errors** — Move aborts decode from the fullnode's clever-error `constantName` on the gRPC `MoveAbort`, so `PredictMoveError.abortName` comes from chain rather than a table that goes stale on redeploy.
- **Narrowed public surface** — the facade (`PredictClient` + `predict`), config, domain-specific unit/tick helpers, typed errors, and the decoder receipt types. Low-level builders and raw reads are internal; power users compose PTBs via the generated bindings under `contracts/`.
- **Deployed-truth targeting** — bindings and config are generated / transcribed from the current testnet deployment `predict-testnet-7-29` (published commit `a92ceb01`); re-pinned whenever it is re-published. `getConfig('mainnet')` throws until a mainnet deployment exists.

### Testing lanes

The builders are generated from a specific deployed package set, so the drift that matters is between this SDK and the live deployment — checked by a smoke suite whose arity guards catch deployed-surface signature drift. It runs against **live testnet, not localnet** (localnet publishes `main`, which drifts from the deployed surface — the wrong parity oracle), so it is deliberately off the PR lane:

- `test` — offline suite only, runs on every PR (`vitest.config.ts` excludes `tests/testnet/**` unless `PREDICT_SDK_TESTNET` is set).
- `test:e2e` — the live suite, on a daily schedule + `workflow_dispatch` (`predict-e2e.yml`), and as a `needs:` gate before publishing (`release-deepbook-predict.yml`). It is oracle-tolerant: NAV/price assertions accept a typed stale-oracle `pricing` abort (by code or name) so an idle testnet oracle keeper can't false-fail the publish gate — while still failing on config errors (wrong feed object) that would signal a real bug.

Version is `0.0.0` + a patch changeset, so the first published version is `0.0.1`.

### Follow-ups (not in this PR)

- Teaching the codegen a well-known-object injection pattern (and adding `ACCUMULATOR_ROOT_ID` to its known objects) would let the thin builder wrappers drop away entirely — a codegen-side improvement, not predict-specific.
- A full live *mint* additionally requires the testnet oracle keeper to be feeding fresh Block-Scholes prices; when it's idle, pricing correctly surfaces a typed `EBlockScholesPriceStale` (verified) rather than a successful fill.

## Test plan

- `pnpm turbo build --filter @mysten/deepbook-predict...` — clean (incl. the `tsc --noEmit` build step).
- `pnpm --filter @mysten/deepbook-predict test` — **112/112** offline tests across 12 files (`tests/testnet/` excluded from this lane).
- `pnpm --filter @mysten/deepbook-predict test:e2e` — **10/10** against the live `predict-testnet-7-29` deployment: arity guards on mint/redeem, every configured object id resolves, gRPC `SimulateTransactionResult` shape + event BCS decode against real emitted events.
- **End-to-end on testnet**: SDK-built account creation and deposit executed on-chain (`success: true`); the created account wrapper matched the SDK's deterministic derivation; reads return correct live state.
- `pnpm --filter @mysten/deepbook-predict run lint` — `oxlint` 0/0, `prettier -c` clean.
- `pnpm changeset status` — resolves `@mysten/deepbook-predict` for a patch bump.

Not yet exercised: the release workflow. Recommend triggering `release-deepbook-predict.yml` with `dry_run: true` once before the first publish.

---

### AI Assistance Notice


packages/deepbook-predict/src/config/index.ts
```diff
@@ -0,0 +1,35 @@
+import { TESTNET_CONFIG } from './testnet.js';
+import type { PredictConfig } from './types.js';
+
+export type { PredictConfig, PredictPackages, UnderlyingConfig } from './types.js';
+export { TESTNET_CONFIG } from './testnet.js';
```
hayes-mysten: this is imported twice?

packages/deepbook-predict/src/config/index.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// MVR: when @deepbook/predict is registered, resolve the name here (named-packages
+// plugin); builders never hardcode targets.
+export function predictTarget(
+	cfg: PredictConfig,
+	module: string,
+	fn: string,
+): `${string}::${string}::${string}` {
+	return `${cfg.packages.predict}::${module}::${fn}`;
```
hayes-mysten: you can pass package/module/fn separately the tx.moveCall to avoid having these helpers

packages/deepbook-predict/src/index.ts
```diff
@@ -0,0 +1,112 @@
+export const SDK_NAME = '@mysten/deepbook-predict';
```
hayes-mysten: why is this exported?

packages/deepbook-predict/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+	fromRaw,
+	leverageToRaw,
+	priceToRaw,
+	probabilityToRaw,
+	rawToPrice,
+	rawToProbability,
+	rawToUsdc,
+	toRaw,
+	usdcToRaw,
+} from './units.js';
```
hayes-mysten, line 37: There is a lot of stuff that seems like we probably want to think through more carefully before putting in the public API surface

packages/deepbook-predict/src/errors.ts
```diff
@@ -0,0 +1,121 @@
...
+		this.abortName = abortName;
+	}
+}
+
+// Abort code → name, per module. Generated from the DEPLOYED sources at commit
+// ec99cfae (this repo's packages/predict + packages/account): grep of
+// `const E<Name>: u64 = <n>;` in each module. Regenerate from that same commit if
+// the deployed package changes — a stale table mislabels a real abort.
+export const ABORT_TABLES: Record<string, Record<number, string>> = {
+	expiry_market: {
```
hayes-mysten: shouldn't we be handling this with clever errors instead?

packages/deepbook-predict/src/decode.ts
```diff
@@ -0,0 +1,576 @@
+import { bcs } from '@mysten/sui/bcs';
+import { fromBase64, normalizeSuiAddress } from '@mysten/sui/utils';
+import type { PredictConfig } from './config/index.js';
+import { PredictInputError } from './errors.js';
+import { fromRaw } from './units.js';
+
+// ============================================================================
+// Execution-result decoders.
+//
+// Every tx.* builder returns a Transaction; after the app executes it (with
```
hayes-mysten, line 13: why are we hand writing all of this instead of using codegen?

packages/deepbook-predict/src/tx/plp.ts
```diff
@@ -0,0 +1,104 @@
...
+	const auth = generateAuth(cfg, tx);
+	tx.moveCall({
+		target: predictTarget(cfg, 'plp', 'request_supply'),
+		arguments: [
+			tx.object(cfg.objects.poolVault),
+			tx.object(args.wrapperId),
+			auth,
+			tx.object(cfg.objects.protocolConfig),
+			tx.pure.u64(args.amountRaw),
+			tx.object(ACCUMULATOR_ROOT_ID),
```
hayes-mysten: All of these functions looks like they should just be generated.  We may want to add ACCUMULATOR_ROOT_ID to the codegens list of known objects so it can be omitted in the arguments, but there shoudn't be a need for all these hand written move call wrappers.  

I do see that we are injecting a bunch of object references from the cfg, I could see updating the codegen establishing a pattern for this so that the generated methods can accept a config object with well known object ids

hayes-mysten (commented):
I think a very large portion of this SDK can be replaced by codegen.  I'd also recommend checking out https://sdk.mystenlabs.com/sui/sdk-building

The [Hashi SDK](https://github.com/MystenLabs/ts-sdks/pull/1140) is a great example of these patterns

happy to jump on a call to talk through this if it would be helpful

packages/deepbook-predict/src/config/index.ts
tonylee08: Done — `config/index.ts` now imports `TESTNET_CONFIG` once and re-exports the binding (1489c7ac).

packages/deepbook-predict/src/config/index.ts
tonylee08: Removed — the tx builders now call the generated codegen move-calls, so the hand-written `predictTarget`/`accountTarget` target helpers are gone.

packages/deepbook-predict/src/index.ts
tonylee08: Removed `SDK_NAME` from the public surface.

packages/deepbook-predict/src/index.ts
tonylee08: Narrowed — the public surface is now just the `PredictClient` facade + config + unit/tick helpers + typed errors + typed decoders. The low-level builders and raw reads are internal now; power users compose PTBs via the generated bindings under `contracts/`.

packages/deepbook-predict/src/errors.ts
tonylee08: Done — `ABORT_TABLES` is deleted; `decodeMoveAbort` now reads the fullnode's clever-error `constantName` off the gRPC `MoveAbort` (JSON-RPC doesn't surface it; the SDK is gRPC-only).

packages/deepbook-predict/src/decode.ts
tonylee08: Reworked — the decoders now parse via the generated event `MoveStruct`s from `contracts/**` instead of hand-written BCS layouts.

packages/deepbook-predict/src/tx/plp.ts
tonylee08: Done — the tx builders (plp/trade/account/builderCode) now call the generated `contracts/**` move-calls; the hand layer only injects the cfg objects + accumulator root. Good call on teaching codegen the well-known-object injection pattern (and adding ACCUMULATOR_ROOT_ID to its known objects) to drop even these thin wrappers — noted as a codegen-side follow-up.

0xaslan (approved):

packages/deepbook-predict/src/config/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export type { PredictConfig, PredictPackages, UnderlyingConfig } from './types.js';
+
+/**
+ * The shared AccumulatorRoot object — a fixed well-known id on every network. The
+ * Clock (`0x6`) is not exported: the generated move-call layer auto-injects it.
+ */
+export const ACCUMULATOR_ROOT_ID = '0xacc';
+
+export function getConfig(network: 'testnet' | 'mainnet'): PredictConfig {
+	if (network === 'mainnet') throw new Error('no mainnet deployment');
```
hayes-mysten: nit: I'd invert this so you return when you hit a known network instead

packages/deepbook-predict/src/contracts/account/deps/sui/bag.ts
```diff
@@ -0,0 +1,41 @@
+/**************************************************************
```
hayes-mysten, line 1: can these get added to the github linguist config to mark them as generated

packages/deepbook-predict/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```ts
+import { SuiGrpcClient } from '@mysten/sui/grpc';
+import { PredictClient } from '@mysten/deepbook-predict';
+
+const client = new SuiGrpcClient({
+	network: 'testnet',
+	baseUrl: 'https://fullnode.testnet.sui.io:443',
+});
+const predict = new PredictClient({ network: 'testnet', client });
```
hayes-mysten: we should use the client extensions pattern in all docs

packages/deepbook-predict/src/units.ts
```diff
@@ -0,0 +1,49 @@
...
+	const digits = whole + frac;
+	const point = whole.length + Number(expStr);
+	if (point <= 0) return '0.' + '0'.repeat(-point) + digits;
+	if (point >= digits.length) return digits + '0'.repeat(point - digits.length);
+	return digits.slice(0, point) + '.' + digits.slice(point);
+}
+
+/** Exact decimal→raw conversion via string math. Throws on negatives and excess precision. */
+export function toRaw(value: number | string, decimals: number): bigint {
+	const s = expandExponential(typeof value === 'number' ? value.toString() : value.trim());
```
hayes-mysten, line 20: should we use Intl.NumberFormat instead of custom formatters for all of this? (no strong opinion, but this feels like something that could benefit from the more standardized tools)

packages/deepbook-predict/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+export {
+	decodeAccountsCreated,
+	decodeBuilderCodeSets,
+	decodeClaims,
+	decodeDeposits,
+	decodeMints,
+	decodePlpCancels,
+	decodePlpRequests,
+	decodeRedeems,
+	decodeWithdrawals,
```
hayes-mysten: these are now exposed through the client right, do we still need these here as well?

packages/deepbook-predict/src/index.ts
```diff
@@ -0,0 +1,74 @@
...
+	U64_MAX,
+	fromRaw,
+	leverageToRaw,
+	priceToRaw,
+	probabilityToRaw,
+	rawToPrice,
+	rawToProbability,
+	rawToUsdc,
+	toRaw,
+	usdcToRaw,
```
hayes-mysten, line 36: I don't love having very generic function names like `toRaw` exported directly from an SDK, this kind of thing gets annoying when the collide with other tools or your own custom functions when you accidentally pull in the wrong one.  I'd love to avoid having these if possible, but another option might be to wrap them up in a class/singleton or something like like priceUtils.toRaw() or something

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+		if (!u) throw new PredictInputError(`unknown underlying: ${underlying}`);
+		return {
+			pythFeed: u.pythFeed,
+			blockScholesValueStore: u.blockScholesValueStore,
+			blockScholesSviStore: u.blockScholesSviStore,
+		};
+	}
+
+	// Resolve (and cache) a market's id + state from its human coordinates.
+	private async resolveMarket(
```
hayes-mysten: We usually use `#` instead of `private`

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+/**
+ * The one object an app constructs. Wraps the config, a gRPC client for reads, and
+ * a derived-account model so callers pass owner addresses, decimal amounts, and
+ * human market coordinates — the facade converts to raw units, resolves markets
+ * (cached), and delegates to the internal tx primitives / reads. Callers who need
+ * to compose their own PTBs can use the generated bindings under `contracts/`.
+ */
+export class PredictClient {
+	readonly cfg: PredictConfig;
+	private readonly client: ReadClient;
```
hayes-mysten: You should be using `ClientWithCoreApi` here

packages/deepbook-predict/src/reads/inspect.ts
```diff
@@ -0,0 +1,76 @@
+import type { SuiGrpcClient } from '@mysten/sui/grpc';
+import type { Transaction } from '@mysten/sui/transactions';
+import { normalizeSuiAddress } from '@mysten/sui/utils';
+import { decodeMoveAbort, type MoveAbortError } from '../errors.js';
+
+// The one network seam every read sits on. Structural — any object with a
+// `simulateTransaction` matching the gRPC client's satisfies it, so callers can
+// inject a mock in tests and the SDK never constructs its own transport.
+export type ReadClient = Pick<SuiGrpcClient, 'simulateTransaction'>;
```
hayes-mysten: use `ClientWithCoreApi`

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+	}
+
+	// Position enumeration needs object reads beyond the simulate-only
+	// ReadClient contract; real SuiGrpcClient/SuiJsonRpcClient instances have
+	// them, simulate-only mocks do not — fail with a pointed message.
+	private objectReadClient(): ObjectReadClient {
+		const c = this.client as unknown as ObjectReadClient;
+		if (typeof c.getObject !== 'function' || typeof c.listDynamicFields !== 'function') {
+			throw new PredictInputError(
+				'read.positions needs a client with getObject + listDynamicFields (e.g. SuiGrpcClient)',
```
hayes-mysten: ClientWithCoreApi should remove the need for these kinds of checks across the whole SDK

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+		owner: string,
+		m: MarketDescriptor,
+		opts: CloseOptions,
+	): Promise<Transaction> {
+		const feeds = this.feeds(m.underlying);
+		const { id } = await this.resolveMarket(m);
+		const closeQuantityRaw = usdcToRaw(opts.quantity);
+		this.assertLot(closeQuantityRaw);
+		const tx = new Transaction();
+		redeemLive(this.cfg, tx, {
```
hayes-mysten: All of these methods should be using transaction thunks instead:

`tx.add(redeemLive(this.config, { ...args }))`

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+					useGasCoin: false,
+				}),
+			);
+			depositFunds(this.cfg, tx, { wrapperId: this.wrapperIdFor(owner), coin });
+			return tx;
+		},
+
+		withdraw: (owner: string, amountUsdc: number | string): Transaction => {
+			const tx = new Transaction();
+			const coin = withdrawFunds(this.cfg, tx, {
```
hayes-mysten: `const coin = tx.add(withdrawFunds(this.confg, { ...args `

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+				wrapperId: this.wrapperIdFor(owner),
+				amountRaw: usdcToRaw(amountUsdc),
+			});
+			tx.transferObjects([coin], owner);
```
hayes-mysten: do we actually want to transfer as coin instead of balance?

packages/deepbook-predict/src/client.ts
```diff
@@ -0,0 +1,675 @@
...
+			return tx;
+		},
+
+		mint: (owner: string, m: MarketDescriptor, opts: MintOptions): Promise<Transaction> =>
+			this.buildMint(owner, m, opts),
```
hayes-mysten: why do do have both versions?

packages/deepbook-predict/src/tx/trade.ts
```diff
@@ -0,0 +1,190 @@
...
+
+// Load a fresh `Pricer` from the live oracle feeds. Every live-flow trade call
+// (`mint_*`, `redeem_live`) borrows this `&Pricer` and it must be loaded first in the
+// PTB. Deployed sig `load_live_pricer` (expiry_market.move): market, config,
+// propbook_registry (&OracleRegistry), pyth, bs_values, bs_svi, clock (auto-injected).
+export function loadLivePricer(
+	cfg: PredictConfig,
+	tx: Transaction,
+	args: { expiryMarketId: string } & MarketFeeds,
+): TransactionResult {
```
hayes-mysten: these should all be thunks for better composition:
```ts
export function loadLivePricer(
	cfg: PredictConfig,
	args: { expiryMarketId: string } & MarketFeeds,
) => (tx: Transaction): TransactionResult => { 
```

Then you can do 

`const result = tx.add(loadLivePricer(cfg, args))`

packages/deepbook-predict/src/reads/parse.ts
```diff
@@ -0,0 +1,57 @@
+// BCS decoders for the return values of on-chain reads. Hand-written (not via the
+// `bcs` codec) so the read layer carries no runtime coupling to a schema per read —
+// these shapes (u64 LE, vector<ID>, Option) are stable Move ABI. Ported from the
+// harness's devInspect parsers (`packages/predict/harness/ts/runtime.ts:394-415`).
+// Round-tripped against `@mysten/sui/bcs` in tests/reads.test.ts.
+
+// Read a ULEB128-encoded length prefix, returning [value, nextOffset].
+function readUleb(bytes: Uint8Array, offset: number): [number, number] {
```
hayes-mysten: why custom bcs decoders instead of using the bcs library instead?  this does not decrease dependencies, and should not be a significant perf bottle neck

packages/deepbook-predict/src/config/index.ts
tonylee08: Done (0cf06e7f) — `getConfig` now returns on a known network and throws otherwise.

packages/deepbook-predict/src/contracts/account/deps/sui/bag.ts
tonylee08: Done (0cf06e7f) — added `packages/deepbook-predict/src/contracts/** linguist-generated=true` to the root .gitattributes, matching the deepbook-v3/walrus/kiosk convention.

packages/deepbook-predict/README.md
tonylee08: Done — the README quickstart and every example now use `client.$extend(predict({ network }))` and `client.predict.*`.

packages/deepbook-predict/src/units.ts
tonylee08: Kept the custom converters: they do exact string/bigint decimal↔raw math (no floats on the money path). `Intl.NumberFormat` is a locale *display* formatter, not a raw/decimal converter, so it doesn't fit here — display formatting stays the app's job.

packages/deepbook-predict/src/index.ts
tonylee08: Done — dropped the standalone `decode*` exports; they're reached via `client.predict.decode.*`. Kept the receipt types.

packages/deepbook-predict/src/index.ts
tonylee08: Done — dropped the generic `toRaw`/`fromRaw` from the public surface (now internal); kept the domain-specific converters (`usdcToRaw`, `priceToRaw`, …).

packages/deepbook-predict/src/client.ts
tonylee08: Done — all facade fields and methods are `#`-private now.

packages/deepbook-predict/src/client.ts
tonylee08: Done — the client is typed `ClientWithCoreApi` (`PredictCompatibleClient`), with a `predict()` extension factory.

packages/deepbook-predict/src/reads/inspect.ts
tonylee08: Done — `ReadClient = ClientWithCoreApi`; reads call `client.core.simulateTransaction`.

packages/deepbook-predict/src/client.ts
tonylee08: Done — removed the `objectReadClient()` capability check; the positions read passes the `ClientWithCoreApi` straight through to `client.core.getObject`/`listDynamicFields`.

packages/deepbook-predict/src/client.ts
tonylee08: Done — all builders are transaction thunks; the facade composes with `tx.add(builder(cfg, args))`.

packages/deepbook-predict/src/client.ts
tonylee08: Done — `const coin = tx.add(withdrawFunds(this.cfg, { ... }))`.

packages/deepbook-predict/src/client.ts
tonylee08: Kept transferring a `Coin`: a `Balance` can't be owned by an address, so returning the withdrawn funds to the owner as a `Coin` is the correct form (a balance path would just wrap into a Coin to hand it over anyway).

packages/deepbook-predict/src/client.ts
tonylee08: Both are intentional, mapping to distinct on-chain entrypoints: `mint` → `mint_exact_quantity` (buy an exact payout size), `mintAmount` → `mint_exact_amount` (spend an exact premium, floor the quantity received).

packages/deepbook-predict/src/tx/trade.ts
tonylee08: Done — builders are now `(cfg, args) => (tx) => result`; e.g. `tx.add(loadLivePricer(cfg, args))`.

packages/deepbook-predict/src/reads/parse.ts
tonylee08: Done — `parse.ts` now delegates to `@mysten/sui/bcs` (`bcs.u64` / `bcs.vector(bcs.Address)` / `bcs.option`).

packages/deepbook-predict/src/reads/inspect.ts
```diff
@@ -0,0 +1,76 @@
...
+	return (result.Transaction?.events ?? []) as { eventType?: string; bcs?: Uint8Array }[];
+}
+
+export async function inspectReturns(
+	client: ReadClient,
+	tx: Transaction,
+	sender: string = normalizeSuiAddress('0x0'),
+): Promise<Uint8Array[][]> {
+	tx.setSender(sender);
+	const result = await client.core.simulateTransaction<{ commandResults: true }>({
```
hayes-mysten: you don't need to pass the generic args explicitly

hayes-mysten (approved):

packages/deepbook-predict/src/reads/inspect.ts
tonylee08: Done (624974ec) — dropped the explicit generics from both `simulateTransaction` call sites (the only two); TS infers the include-shape from the `include` field.

PR #1153, opened 2026-07-20
feat(deepbook-v3): sync SDK with deepbook_margin v6 (codegen builders)

tonylee08:
## Description

The SDK was last aligned to `deepbook_margin` v5 ([#1069](https://github.com/MystenLabs/ts-sdks/pull/1069), 2026-05-20). This syncs it to **v6**, which is live on mainnet at `0x8af25e44`. The Move surface over that window is purely additive, so this is a catch-up with no migration burden.

### New margin entry points

| Move entry point | SDK |
| --- | --- |
| `pool_proxy::place_market_order_and_repay_loan` | `poolProxy.placeMarketOrderAndRepayLoan` |
| `pool_proxy::place_reduce_only_limit_order_and_repay_loan` | `poolProxy.placeReduceOnlyLimitOrderAndRepayLoan` |
| `pool_proxy::place_reduce_only_market_order_and_repay_loan` | `poolProxy.placeReduceOnlyMarketOrderAndRepayLoan` |
| `margin_manager::execute_conditional_orders_v3` | `marginTPSL.executeConditionalOrdersV3` |
| `margin_registry::min_open_risk_ratio` | `getMinOpenRiskRatio` |
| `margin_registry::set_min_open_risk_ratio` | `marginAdmin.setMinOpenRiskRatio` |

The `*_and_repay_loan` variants repay the loan from the fill proceeds *before* the risk check, so the gate is the net post-repay `risk_ratio` rather than the borrow floor. Same for `execute_conditional_orders_v3` — that is what lets a stop-loss fire in the `liquidation..min_borrow` band, where a swap alone only lowers the oracle-valued ratio and gets rejected by v2's gate.

`executeConditionalOrders` (v2) is deliberately left in place and unchanged, mirroring the Move side keeping v2 live. Keepers migrate on their own schedule.

### These builders are constructed from codegen (named args, not positional)

Per review feedback ([thread](https://github.com/MystenLabs/ts-sdks/pull/1153#discussion_r3624642875)): the existing margin builders use positional `moveCall` argument arrays, which risk transposing same-typed arguments (base/quote oracles, base/quote margin pools). The reason they were positional is that **`deepbook_margin` was never in the codegen config** — only `@deepbook/core` was — so there were no generated bindings to delegate to.

This PR adds `@deepbook/margin` to `sui-codegen.config.ts` and regenerates, so `src/contracts/deepbook_margin` now exists. The six new builders delegate to the generated **named-argument** functions:

```ts
poolProxyMoveCalls.placeMarketOrderAndRepayLoan({
  package: this.#config.MARGIN_PACKAGE_ID,
  arguments: { registry, marginManager, pool, baseMarginPool, quoteMarginPool,
               baseOracle, quoteOracle, clientOrderId, selfMatchingOption,
               quantity, isBid, payWithDeep },   // named — cannot transpose
  typeArguments: [baseCoin.type, quoteCoin.type],
});
```

Human-unit conversion (`convertQuantity`/`convertPrice`/`convertRate`) stays in the facade; codegen only handles call construction. The generated PTBs are byte-identical to the positional versions (verified on mainnet — see test plan).

This is scoped to the *new* builders. Migrating the ~40 existing positional margin builders to the same pattern is a proposed follow-up so the whole package moves to the generated pattern ([sdk-building](https://sdk.mystenlabs.com/sui/sdk-building)) in one pass.

### Package IDs

Realigned to deepbookv3 `Published.toml`:

| | before | after |
| --- | --- | --- |
| mainnet `MARGIN_PACKAGE_ID` | `0x124bb3d8` (v5) | `0x8af25e44` (v6) |
| testnet `DEEPBOOK_PACKAGE_ID` | `0x22be4cad` (v17) | `0xd874d241` (v20) |
| testnet `MARGIN_PACKAGE_ID` | `0xd6a42f4d` | `0xe52c1dec` (v14) |
| testnet `MARGIN_V1` | `0xd6a42f4d` | `0xb8620c24` |

The mainnet margin bump is required — the six new entry points do not exist at the v5 address.

**`MARGIN_V1` was a latent bug.** It builds `MarginApp` type tags (`deepbookAdmin.ts`), so it must be the package *original* ID. Mainnet's value was already correct; testnet's pointed at `0xd6a42f4d`, an abandoned lineage. The real testnet original ID is `0xb8620c24`, confirmed by `Published.toml` and by reading the on-chain `MarginRegistry` object's type. Testnet margin had been stale since February, so this was not reachable via the current SDK anyway.

### Not included: three undeployed core functions

deepbookv3 [#973](https://github.com/MystenLabs/deepbookv3/pull/973) added `place_post_only_limit_order`, `best_bid_price`, `best_ask_price` to core, but that commit postdates both testnet core v20 and mainnet core v8 — they return `unable to find function` on both networks today. They are deliberately **left out** of this PR and will be added when a core package carrying them ships (at which point it's a `DEEPBOOK_PACKAGE_ID` bump plus the builders). The `@deepbook/core` generated bindings are intentionally **not** regenerated here either, to keep this PR scoped to margin.

## Test plan

- [x] `tsc --noEmit`, `pnpm build` (ESM + CJS + d.ts), `pnpm vitest run` (19/19), `pnpm lint` (oxlint + prettier) — all clean.
- [x] **Byte-identical PTBs after the codegen switch**: every new builder resolves against a real deployed mainnet margin v6 `MarginManager` (no `unable to find function`, no signature/arity mismatch), and `getMinOpenRiskRatio('SUI_USDC')` returns `1.17495` through the codegen path — which also confirms the generated `0x2::object::ID` argument encodes as a pure value, not an object reference.
- [x] **Package IDs verified against chain**: `min_open_risk_ratio` exists on mainnet margin v6 and returns `unable to find function` on v5 (bump required); every changed constant matches `Published.toml`; the registry objects resolve to the expected original-ids; `MARGIN_V1` corrected value matches the on-chain `MarginRegistry` type.

Not covered: signed execution of the new margin writes needs a funded margin manager we own; those are verified to simulation/resolution only.

---

### AI Assistance Notice


packages/deepbook-v3/src/transactions/poolProxy.ts
```diff
@@ -221,6 +221,158 @@ export class PoolProxyContract {
...
+			} = params;
+			const pool = this.#config.getPool(poolKey);
+			const manager = this.#config.getMarginManager(marginManagerKey);
+			const baseCoin = this.#config.getCoin(pool.baseCoin);
+			const quoteCoin = this.#config.getCoin(pool.quoteCoin);
+			const baseMarginPool = this.#config.getMarginPool(pool.baseCoin);
+			const quoteMarginPool = this.#config.getMarginPool(pool.quoteCoin);
+			const inputPrice = convertPrice(price, FLOAT_SCALAR, quoteCoin.scalar, baseCoin.scalar);
+			const inputQuantity = convertQuantity(quantity, baseCoin.scalar);
+			return tx.moveCall({
```
hayes-mysten: some day I'd love to get deepbook in line with the sdk patterns from https://sdk.mystenlabs.com/sui/sdk-building

I do wonder for methods like these why we don't use the generated methods we have in codegen, all these positional args seem very kinda risky

tonylee08: Good call — done in cfd3d37. The reason these were positional is that `deepbook_margin` wasn't in the codegen config (only `@deepbook/core` was), so there were no generated bindings to delegate to. I added `@deepbook/margin` to `sui-codegen.config.ts` and regenerated, and the new margin builders now delegate to the generated named-argument functions (e.g. `poolProxy.placeMarketOrderAndRepayLoan` → `contracts/deepbook_margin/pool_proxy`), so the same-typed args (base/quote oracles, base/quote margin pools) can't be transposed. Human-unit conversion (`convertQuantity`/`convertPrice`/`convertRate`) stays in the facade; codegen only handles call construction.

Verified the PTBs are byte-identical: all the new builders still resolve against deployed mainnet margin v6, and `getMinOpenRiskRatio` still returns the same value.

I scoped this to the *new* builders in this PR rather than migrating the ~40 existing margin builders — happy to do that as a follow-up so the whole package moves to the generated pattern (matching the sdk-building guide) in one go.

Separately, I dropped the three core entry points (`place_post_only_limit_order`, `best_bid_price`, `best_ask_price`) from this PR — they're from deepbookv3 #973 and aren't deployed on any network yet, so they'll come back when a core package carrying them ships.

hayes-mysten (approved):

PR #1158, opened 2026-07-22
feat(suins): migrate to Pyth Pro

John-Atha:
## Description

Pyth is retiring the public Hermes endpoint in favour of a keyed Pyth Pro endpoint and upgrading the on-chain Pyth/Wormhole packages (Core→Pro cutover, see https://docs.pyth.network/price-feeds/core/upgrade/preparing/sui). This migrates the SuiNS SDK's Pyth integration to Pro.

Scoped to **testnet** for now; mainnet follows once its contracts are upgraded (see Pending).

**Price fetching**

- Fetch price updates from the keyed Pyth Pro Hermes endpoint on the v2 API (`/v2/updates/price/latest`), replacing the deprecated public v1 (`/api/latest_vaas`) host. A single host serves all networks.
- Add an optional `pythAccessToken` to `SuinsClient` and the `suins()` extension, sent as an `Authorization: Bearer` header. It is required for non-base-currency registrations/renewals (the base currency does not use the oracle).

**On-chain wiring**

- Point the Pyth and Wormhole state ids at the Pro deployments (both networks); converge the testnet price feeds to the global feed ids.
- Bump the testnet `payments` / `bbb` package ids to their upgraded versions. Add `payments.packageIdV1` (the original id) so the `PaymentsConfig` type identity stays keyed by the original package while calls target the latest one.
- Regenerate the `payments` binding and switch the non-base register/renew path to the Pro `_pro` entrypoints (`calculate_price_pro` / `handle_payment_pro`). The old Core entrypoints now abort on-chain (`ECoreFeedDeprecated`).

**Pending (mainnet, follow-up)**

- Mainnet contracts are not yet upgraded (tracked in SUIP-1105). The SDK calls the `_pro` entrypoints unconditionally, so the mainnet flow works only once mainnet is upgraded and its ids are bumped (mirroring testnet). Kept as a draft until then.

## Test plan

- **Real testnet registration executed end-to-end** paying in SUI: `status: success`, `SuinsRegistration` NFT minted, and `pyth` + `payments` events emitted (a live Pyth Pro price update settled on-chain, consumed by `handle_payment_pro`). Tx `FFc4iJqJ67kyCNTgEvH23fkoUwr1GqpJVsSwwYov94k8`.
- `live.test.ts` testnet dry-run green against the Pro endpoint + Pro contracts (skips when `VITE_PYTH_ACCESS_TOKEN` is unset). Reproducing requires the testnet `PaymentsConfig` feed ids aligned to the updated Pyth Pro feed ids (related discussion [here](https://docs.google.com/document/d/1Yh3ApaCFkAhoS_GmfXqETRwX1RxDqFuF-NGMM2M7O1s/edit?disco=AAACD6ZllQs)); that on-chain admin step is tracked on the contracts side (SUIP-1105) and is temporarily reverted on testnet now to avoid blocking users that are still on the previous package versions.
- Auth: the keyed endpoint returns `401` without a token and `200` with a valid one.

---

### AI Assistance Notice


packages/suins/src/suins-client.ts
```diff
@@ -285,11 +291,10 @@ export class SuinsClient {
 	}
 
 	async getPriceInfoObject(tx: Transaction, feed: string, feeCoin?: TransactionObjectArgument) {
-		const endpoint =
-			this.network === 'testnet'
-				? 'https://hermes-beta.pyth.network'
-				: 'https://hermes.pyth.network';
-		const connection = new SuiPriceServiceConnection(endpoint);
+		const endpoint = 'https://pyth.dourolabs.app/hermes';
```
natalypeyioti, line 300: The endpoint is hardcoded inline while every other deployment value lives in constants.ts — can it move there?

packages/suins/src/suins-client.ts
```diff
@@ -285,11 +291,10 @@ export class SuinsClient {
 	}
 
 	async getPriceInfoObject(tx: Transaction, feed: string, feeCoin?: TransactionObjectArgument) {
```
natalypeyioti, line 293: pythAccessToken is required here but never checked — without it this fails with a bare 401 from axios that doesn't mention the token. Can we throw early with a clear message?

packages/suins/src/suins-client.ts
```diff
@@ -62,10 +66,12 @@ export class SuinsClient {
 	client: ClientWithCoreApi;
 	network: SuiClientTypes.Network;
 	config: PackageInfo;
+	pythAccessToken?: string;
```
natalypeyioti: Nit: this is a credential on a public mutable field — nothing outside the class reads it, so readonly would be safer.

packages/suins/src/suins-client.ts
John-Atha: right, added it

packages/suins/src/suins-client.ts
John-Atha: I would prefer us to keep the diff here as simple as possible, meaning do not touch existing conventions of the repo, and keep it focused on just the Pyth pro compatibility.
Since the hermes endpoints were already living inline in this file, let's keep this structure as it is.
We can always suggest general refactorings of the codebase in separate PRs to the sdks team, that are purely scoped to that, if we consider them useful.

packages/suins/src/pyth/PriceServiceConnection.ts
```diff
@@ -38,11 +40,13 @@ export class PriceServiceConnection {
...
+		const response = await this.httpClient.get('/v2/updates/price/latest', {
 			params: {
-				ids: priceIds,
+				'ids[]': priceIds,
+				encoding: 'base64',
+				parsed: false,
 			},
 		});
-		return response.data;
+		return response.data.binary.data;
```
natalypeyioti: If Hermes ever changes this shape we get `Cannot read properties of undefined` rather than a real error. Worth guarding with an explicit check.

packages/suins/src/constants.ts
```diff
@@ -73,27 +74,28 @@ export const mainPackage: Config = {
...
 		},
 		/// Testnet coins will be different here for testing purposes, we can publish our own
 		coins: {
 			SUI: {
 				type: '0x0000000000000000000000000000000000000000000000000000000000000002::sui::SUI',
-				feed: '0x50c67b3fd225db8912a424dd4baed60ffdde625ed2feaaf283724f9608fea266',
+				feed: '0x23d7315113f5b1d3ba7a83604c44b94d79f4fd69af77f804fc7f920a6dc65744',
 			},
 			/// this is a test token published as 0xb48aac3f53bab328e1eb4c5b3c34f55e760f2fb3f2305ee1a474878d80f650f0::TESTNS::TESTNS
 			/// NS token is using the HFT feed since NS feed on testnet is not available
```
natalypeyioti: stale comment?

natalypeyioti (approved):

packages/suins/src/pyth/PriceServiceConnection.ts
John-Atha: Good call, added an explicit guard, it now throws Unexpected Hermes response: missing binary.data instead of the bare TypeError if the shape drifts. Worth noting for context: most of the HTTP clients in this repo (enoki, walrus upload-relay, seal) just cast the response with as T and access fields unguarded, so this is stricter than the norm here, but I agree it's the right thing for a required payload like this one. Happy to keep it.

packages/suins/src/constants.ts
John-Atha: indeed, removed. Testnet suins price feed now is the same as mainnet suins

hayes-mysten (approved):

PR #1161, opened 2026-07-22
test(sui): refetch cross-client consistency reads that straddle live-state changes

Bridgerz:
## Problem

`Localnet` e2e runs flake on `all clients return same data: getCurrentSystemState` (seen on [#1159 CI](https://github.com/MystenLabs/ts-sdks/pull/1159), unrelated PR): the failing diff is always epoch-scoped state — `distributionCounter`, `stakeSubsidy.balance`, `storageFund.nonRefundableBalance`/`totalObjectStorageRebates` — i.e. the three transport reads straddled a localnet epoch boundary. The existing `{ retry: 3 }` re-runs the whole test but each attempt has the same race, and the timestamp normalization doesn't cover epoch-scoped values.

## Fix

`expectAllClientsReturnSameData` gains an `attempts` option: on comparison mismatch it refetches the entire read set (all three transports, still concurrent) after a 500ms pause, up to N times. A genuine transport inconsistency reproduces on every refetch and still fails; only boundary-straddles converge. Default `attempts: 1` keeps every other call site's strict single-shot behavior. `getCurrentSystemState` opts in with `attempts: 3`.

Formatted with the repo prettier config.

---

### AI Assistance Notice

### DIFF
--- a/packages/sui/test/e2e/clients/core/system-state.test.ts
+++ b/packages/sui/test/e2e/clients/core/system-state.test.ts
@@ -26,6 +26,9 @@ describe('Core API - System State', () => {
 						},
 					};
 				},
+				// System state advances every localnet epoch; refetch when the
+				// three transports straddle a boundary mid-read.
+				{ attempts: 3 },
 			);
 		});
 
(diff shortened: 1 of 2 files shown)


hayes-mysten (approved):
thanks!

PR #1196, opened 2026-08-11
docs: restructure client docs around audience; fix Core API top-level parity

hayes-mysten:
## Description

Started as "document the query methods and gRPC subscriptions", turned into a restructure of the
client docs plus the SDK fix that documenting them exposed.

### Docs restructure

The client pages repeated themselves — the shared method surface was documented three times over, in
`core.mdx` and both transport pages — and none of them reflected who they were for. Reorganized
around audience (`meta.json` order: `index`, `querying`, `grpc`, `graphql`, `core`):

| Page | Purpose |
| --- | --- |
| `index.mdx` | Choosing a client; the three calling surfaces |
| `querying.mdx` *(new)* | The method surface every client shares, documented once |
| `grpc.mdx` | Only what is gRPC-specific: transports, native services, raw list RPCs, subscriptions |
| `graphql.mdx` | GraphQL as a first-class choice: batching, historical reads, typed queries, polling |
| `core.mdx` | The `ClientWithCoreApi` contract, for SDK authors |

`json-rpc.mdx` is deleted — deprecated clients belong in migration material, not in pages showing
people how to use them. A 302 from `/sui/clients/json-rpc` to the migration guide is in
`next.config.mjs`. GraphQL is now presented as a fully supported primary choice rather than an
escape hatch for what gRPC can't do.

New material — the query methods from #1137 and the gRPC subscription APIs — is folded in at a
weight proportional to how often it is actually used, rather than dominating the pages.

### `fix(sui)`: Core API top-level parity

`getCurrentSystemState`, `getProtocolConfig`, `getChainIdentifier`, and `getDynamicObjectField` were
reachable only through `client.core`. The forwards on both concrete clients are written by hand, so
a method added to the contract without a matching forward silently ends up core-only: #961 added
`getProtocolConfig` to the contract, all three implementations, an e2e test, and the wallet-sdk mock
without touching either client. All four are now forwarded.

The same diff exposed a second bug: `GraphQLCoreClient.getCurrentSystemState` and `getProtocolConfig`
declared no parameters and so silently dropped the `signal` the contract and the gRPC implementation
both accept. Both now forward it.

A unit test walks a concrete core client's prototype chain and asserts every method it finds is also
forwarded. Enumerating `CoreClient` directly does not work — its abstract members emit nothing at
runtime, so the check passes vacuously; the first version of this test did exactly that and was
caught by deleting a forward and watching it still pass. There is also a guard asserting the
enumeration finds a plausible method count, so it cannot silently go vacuous again.

### Corrections found while verifying against source

Three review passes — source accuracy, code examples, and editorial — turned up defects, most
predating this PR:

- The collection pagination example did not type-check (`TS7022`). The list methods are generic over
  `include`, so carrying the cursor in a separately annotated variable makes the page type depend on
  the argument holding the cursor, which depends on the page. The hazard is now documented inline.
- The MVR example used `@mysten/sui` — an npm package name — to "resolve" `0x2`, a system package
  never published to the registry. As written the call would throw.
- `listOwnedObjects` was documented as filtering on an exact type; both transports accept a package,
  module, type name, or full instantiation.
- `0x2::coin::transfer` does not exist, so that filter example returned an empty page forever.
- `doGasSelection` and `listDynamicFields`' `include: { value: true }` were described as
  transport-specific on both pages. They are declared identically on both clients — shared options
  simply absent from the Core API. `checksEnabled` had lost its explanation entirely while the
  migration guide still tells people to use it; both now live with `simulateTransaction`.
- `Display.output` is `Record<string, unknown>`, not `Record<string, string>` (widened in d0a401e7).
- `getChainIdentifier` returns the base58 genesis checkpoint digest; the old example showed a short
  hex id no transport returns.
- `getObject` throws on failure — only `getObjects` returns `Error` entries in place.
- An example threw `new Error(result.FailedTransaction.status.error)`, passing an `ExecutionError`
  object where a string belongs.

## Test plan

- `pnpm lint` clean (the one remaining oxlint warning is a pre-existing `@csspart` jsdoc tag in
  `dapp-kit`, untouched by this branch).
- `pnpm --filter @mysten/sui test` — 469 tests pass, including the new parity guard, which was
  mutation-tested by removing a forward and confirming it fails with the missing method name.
- `pnpm --filter @mysten/docs validate-docs` passes; it checks anchor fragments as well as page
  links, so the repointed `#objectbcs` references in `sui/bcs.mdx` and `codegen/index.mdx` are
  confirmed resolvable.
- `pnpm --filter @mysten/docs build:docs` regenerates 102 files, `json-rpc.md` gone and
  `querying.md` present; the full Next.js build exits 0.
- Corrected code examples were verified by type-checking them against the package rather than by
  eye; every GraphQL document in the docs validates against `schema.graphql` with `graphql-js`.
- Also regenerates the GraphQL schema and `gql.tada` types from upstream, adding
  `ForwardingAddressRegistryCreateTransaction` to `EndOfEpochTransactionKind`. gRPC protos were
  regenerated against `sui-apis` `main` and came out unchanged.

---

### AI Assistance Notice


hayes-mysten:
@clud-bot review

jessiemongeon1:
## Style Guide Audit

All 11 file(s) pass the style guide audit.

hayes-mysten:
@clud-bot rereview and resolve addressed comments

hayes-mysten:
@clud-bot look again

hayes-mysten:
@clud-bot review

Both threads from the last review are addressed in 1680732e:

- `raceSignal` now rejects up front when `signal.aborted` is already true. Added a regression test and mutation-tested it — removing the guard fails with `promise resolved "'value'" instead of rejecting`.
- The caller's signal is now passed into MVR resolution in `listOwnedObjects`, `listCoins`, `getBalance`, `getCoinMetadata` and `getMoveFunction` on the GraphQL client. The deprecated JSON-RPC client had the same gap in `getCoinMetadata` and `getMoveFunction`, so those are fixed too.

b48fa34a clears the last two style-guide findings (passive voice).

hayes-mysten:
@clud-bot review

Both threads addressed in e2fe3b14:

- `CoreClient.getDynamicObjectField` now forwards the signal into its MVR type resolution, matching the `getDynamicField` path.
- JSON-RPC `getMoveFunction` now forwards the signal to `getNormalizedMoveFunction`.

Rather than fix only those two, I swept all four client files with a balanced-paren scan for request and MVR calls that drop the signal. That found three more in the JSON-RPC client: `listDynamicFields`, `verifyZkLoginSignature`, and the reference gas price lookup `simulateTransaction` performs while building. Its `getChainIdentifier` also now races the signal against its cached read, matching gRPC and GraphQL.

The sweep now reports no remaining gaps, other than the three cached `getChainIdentifier` reads, which intentionally keep the shared request uncancelled and race the caller's signal instead.

hayes-mysten:
@clud-bot review

Re-requesting: the last review was pinned to `b48fa34a` and landed about three minutes before `e2fe3b14` was pushed, so it re-reported the two gaps that commit fixes.

At `e2fe3b14`:
- `CoreClient.getDynamicObjectField` passes `signal: options.signal` into `mvr.resolveType`.
- JSON-RPC `getNormalizedMoveFunction` receives `signal: options.signal`.

Plus three more found by sweeping the JSON-RPC client for the same shape (`listDynamicFields`, `verifyZkLoginSignature`, and the gas price lookup in `simulateTransaction`), and `getChainIdentifier` there now races the signal against its cached read like the other two clients.

hayes-mysten:
@clud resume?

hayes-mysten:
@clud-bot review

Fixed in 6fe22fb0. All fifteen `core.mvr.*` calls in `jsonRpc/client.ts` now pass the signal each method already had in scope (`input.signal` in `getOwnedObjects`, the destructured `signal` elsewhere).

You were right that the previous sweep was incomplete, and the reason is worth recording: I scanned a hand-picked list of `core.ts` files rather than the tree, so an entire file went unchecked. This round I swept every `mvr.resolve*` call in the package.

That now reports exactly one remaining unsignalled call: the named-packages transaction plugin, which has no signal to forward since `BuildTransactionOptions` does not carry one. Threading a signal through `Transaction#build` looks like a larger API change, so I left it and said so in the changeset rather than claiming completeness again.

hayes-mysten:
@clud-bot review

Re-requesting for `6fe22fb0`. The previous review was pinned to `e2fe3b14`; its finding (unsignalled MVR lookups in `jsonRpc/client.ts`) is fixed in the commit after it.

Both times a request posted within a few minutes of a review landing has gone unanswered, so this is a re-ping rather than a new change.

PR #1198, opened 2026-08-14
feat(deepbook-account): extract the shared account primitive from deepbook-predict

tonylee08:
## Description

The on-chain **account primitive** (`packages/account` — `AccountWrapper`, `Auth`, custody balances) is shared infrastructure: both `deepbook_core_account` and `predict` depend on it in the Move repo. It was living inside `@mysten/deepbook-predict`, which made a DeepBook-wide primitive Predict-owned.

This extracts it into a new **`@mysten/deepbook-account`** package:

- **`AccountContract`** — `createAccount`, `createAccountAndDeposit`, `depositFunds`, `withdrawFunds`, `generateAuth`, `loadAccount`, `balance`, plus `deriveAccountWrapperId` (computes an owner's canonical wrapper id off-chain, no chain read).
- **Generated Move bindings + BCS structs**, exported for callers composing their own PTBs or decoding account state/events.
- **`AccountConfig`** is deliberately minimal (`ACCOUNT_PACKAGE_ID`, `ACCOUNT_REGISTRY_ID`) and structurally satisfied by `DeepBookConfig`, so each consumer drives the builders against its **own** deployment. That matters today: the only deployed instance of the account package is Predict's.

`@mysten/deepbook-predict` now depends on it. **No public API change** — `deriveAccountWrapperId`, `generateAuth`, and `ACCUMULATOR_ROOT_ID` are still exported from predict via a thin config adapter, and every transaction it builds is byte-for-byte identical.

### Why a separate package rather than `@mysten/deepbook-v3`

Hosting it in deepbook-v3 was the first idea, but both the repo and the docs point elsewhere:

- Every existing cross-package `@mysten/*` dependency targets a **small shared primitive** (`@mysten/wallet-standard`, `@mysten/window-wallet-core`, `@mysten/utils`, `@mysten/walrus-wasm`). No product SDK depends on another product SDK — nothing depends on `@mysten/deepbook-v3` except `@mysten/docs`.
- Per [sdk-building.mdx](https://github.com/MystenLabs/ts-sdks/blob/main/packages/docs/content/sui/sdk-building.mdx), `@mysten/*` packages should be peer dependencies; making deepbook-v3 a peer of predict would force **every** Predict consumer to install deepbook-v3.
- deepbook-v3 doesn't model this package at all today (its `contracts/deepbook/account.ts` is the unrelated `@deepbook/core::account` module), and it has no account deployment of its own.

Following the precedent, `@mysten/deepbook-account` is a regular dependency of predict, with `@mysten/sui` staying a peer dep. deepbook-v3 can adopt `AccountContract` when `deepbook_core_account` deploys — `DeepBookConfig` already satisfies `AccountConfig`.

## Test plan

- **New package**: 8 unit tests covering command order and argument slots for every builder (`registry.new → share`; `new → auth → deposit_funds → share` with `share` last; auth hot-potato threading; accumulator-root and wrapper slots; `load_account → balance<T>` chaining), plus derivation determinism and package/registry sensitivity.
- **deepbook-predict**: **143/143** offline, including the pre-existing byte-for-byte `deposit` regression pin, which proves the extracted builders emit identical transactions. A new `tx-common.test.ts` asserts the adapter binds Predict's own package + registry ids.
- **Live testnet**: **11/11** `pnpm test:e2e` against the deployed package — the smoke suite's arity guards drive the moved `account_registry::new` / `account::share` builders against the real chain.
- **Workspace**: full `turbo build` green (56/56); `tsc --noEmit`, oxlint, prettier clean in both packages. `@mysten/deepbook-v3` is untouched by this diff.

---

### AI Assistance Notice


xydas97 (approved):
LGTM

packages/deepbook-account/src/account.ts
```diff
@@ -0,0 +1,206 @@
...
+import * as accountRegistry from './contracts/account/account_registry.js';
+import type { AccountConfig as GeneratedAccountConfig } from './contracts/account/config-arguments.js';
+
+/**
+ * The framework's well-known accumulator-root singleton, which settles the address
+ * balances an account's custody funds move through. The generated bindings supply it
+ * automatically; it stays exported for PTBs that compose accounts with packages this
+ * SDK has no bindings for.
+ */
+export const ACCUMULATOR_ROOT_ID = '0xacc';
```
hayes-mysten: this should not be needed anymore

tonylee08: addressed

hayes-mysten (approved):

PR #1203, opened 2026-08-17
feat(sui): unify lookup errors across Core API clients

hayes-mysten:
## Description

Builds on the transport-neutral error direction developed by @Danny-Devs in #988. Daniel is also credited as a co-author in the commit history.

Standardize object and transaction lookup misses across the gRPC, GraphQL, and JSON-RPC Core API implementations:

- Export `SuiClientError`, `ObjectError`, and a new `TransactionError` from `@mysten/sui/client`.
- Add transport-neutral `reason` and resource identity fields so consumers can use `instanceof` plus `reason` without knowing the underlying RPC.
- Return `ObjectError` for per-object gRPC failures, matching the existing GraphQL and JSON-RPC behavior.
- Throw `TransactionError` for confirmed missing transactions on all three transports.
- Preserve raw transport detail in the standard `cause` field where the transport supplies it.
- Use the existing typed `GrpcStatusCode` enum for numeric and string gRPC status mapping.

This intentionally keeps the scope smaller than #988. It does not introduce a transport-detail class hierarchy, change GraphQL error handling globally, or alter resolver aggregation.

### Compatibility

The existing required `ObjectError.code`, message, mutability, and `(code, message)` constructor are preserved. `reason`, `objectId`, `cause`, and the optional third constructor argument are additive. When the new options object is supplied, `reason` is required; the SDK always maps it explicitly rather than guessing portable semantics from a transport-specific code. Existing two-argument construction remains valid and conservatively produces `reason: "unknown"`.

Existing JSON-RPC object-error codes and messages, the GraphQL object-error message format, and the broad `(Object | Error)[]` response type remain unchanged.

Transaction mapping is deliberately narrow:

- gRPC: only `RpcError.code === GrpcStatusCode[GrpcStatusCode.NOT_FOUND]`
- GraphQL: only a null transaction result after GraphQL errors have been handled
- JSON-RPC: only the two known `-32602` missing-transaction message shapes emitted by the legacy and current local implementations

Other transport and validation failures remain unchanged. The standardized transaction subclasses intentionally replace transport-specific missing-transaction error identities; the original errors remain available as `cause`.

## Test plan

- `pnpm turbo build --filter=@mysten/sui`
- `pnpm --filter @mysten/sui test` — 487 tests pass
- `pnpm --filter @mysten/sui vitest run --config test/e2e/vitest.config.mts test/e2e/clients/core/objects.test.ts test/e2e/clients/core/transactions.test.ts` — 308 pass, 4 skipped
- `pnpm --filter @mysten/sui oxlint:check`
- `pnpm prettier:check`
- Live Testnet smoke against the built branch:
  - gRPC percent-encoded `NOT_FOUND` becomes `TransactionError` with the original `RpcError` as `cause`
  - GraphQL null transaction becomes the same `TransactionError`
  - missing objects become `ObjectError` with `reason: "notFound"` on both transports

---

### AI Assistance Notice

### DIFF
--- a/.changeset/tidy-lookup-errors.md
+++ b/.changeset/tidy-lookup-errors.md
@@ -3,3 +3,5 @@
 ---
 
 Add transport-neutral `ObjectError` and `TransactionError` lookup details across the Core API clients.
+The existing transport-specific `ObjectError.code` field and `(code, message)` constructor remain
+supported but are deprecated in favor of the explicit `reason` field and options constructor.
--- a/.changeset/tidy-lookup-errors.md
+++ b/.changeset/tidy-lookup-errors.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': minor
+---
+
+Add transport-neutral `ObjectError` and `TransactionError` lookup details across the Core API clients.
--- a/.changeset/tidy-lookup-errors.md
+++ b/.changeset/tidy-lookup-errors.md
@@ -4,4 +4,4 @@
 
 Add transport-neutral `ObjectError` and `TransactionError` lookup details across the Core API clients.
 The existing transport-specific `ObjectError.code` field and `(code, message)` constructor remain
-supported but are deprecated in favor of the explicit `reason` field and options constructor.
+supported; the transport-neutral `reason`, resource identity, and `cause` fields are additive.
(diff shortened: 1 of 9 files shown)


hayes-mysten:
@clud-bot review

Danny-Devs:
Thanks for the credit — glad #988 turned out useful in the end.

PR #1223, opened 2026-08-25
feat(deepbook-v3): host the shared account primitive on an /account subpath

tonylee08:
## Description

**Phase 1 of consolidating the DeepBook SDKs behind subpath exports**, per the discussion about organising them the way the main TS SDK does (`@mysten/deepbook-v3`, `/account`, later `/predict` and `/sessions`) and allowing inter-deps.

The on-chain account primitive is shared infrastructure: `deepbook_core_account` and `deepbook_predict` both depend on the same Move package. It briefly lived in its own npm package, which is what made a version-skew failure possible between it and Predict. Folding it in alongside spot and margin removes the cross-package edge entirely.

- **`@mysten/deepbook-v3/account`** exports `AccountContract`, the generated `account` bindings (`accountMoveCalls`, `accountRegistryMoveCalls`, `accountEvents`), and the `Account` / `AccountWrapper` BCS structs.
- **The package root's export surface is unchanged** — no existing import moves, so this is additive for the 376-release consumer base.
- **`@mysten/deepbook-predict`** now sources the primitive from the subpath. Its own API is unchanged.
- **`@mysten/deepbook-account`** is removed from the repo. Its published `0.1.0` is self-contained (peer dep on `@mysten/sui` only) and keeps working for anyone already on it — it simply stops receiving updates.

### Things worth a reviewer's eye

- **`Account` name collision, handled deliberately.** The root already exports `Account` = `@deepbook/core::account::Account` (the per-pool trading account). The primitive's `Account` is a different type with an unrelated layout. It is reachable **only** from `/account` and must never be re-exported from the root, or `import { Account } from '@mysten/deepbook-v3'` silently changes meaning for every existing consumer. There's a comment at the export site to that effect.
- **Two `u64` conventions now coexist under `src/contracts/`.** The `account` codegen entry carries `bcsOverrides` parsing `u64`/`u128`/`u256` straight to `bigint`; the DeepBook/margin/pyth entries keep `@mysten/sui/bcs`'s decimal strings. `bcsOverrides` is per-entry, so this is scoped rather than global — but it is a real inconsistency inside one package, and it is documented both in the codegen config and in the README callout a consumer actually reads. Neither representation is scaled, so a value crossing between the two surfaces needs a cast, not a conversion. The config also states that any future entry whose Move closure includes `account` must carry the same overrides, or codegen emits a second copy of those types with string integers beside the bigint copy.
- **Predict's install tree grows.** `@mysten/deepbook-account` had no runtime dependencies; `@mysten/deepbook-v3` has four, so Predict consumers now pull `axios`, `axios-retry` and `@noble/hashes` transitively. Predict's *module graph* never loads any of them — but package managers still install them. Inherent to the consolidation, noted so it isn't a surprise.
- **Root config files are now type-checked.** `tsconfig.json` scoped `tsc` to `src`, so `sui-codegen.config.ts`, `tsdown.config.ts` and `vitest.config.mts` were never checked — the hole that let a codegen config referencing an undeclared identifier pass build, lint and the full test suite in an earlier revision of Phase 3. Including them makes that a build error.
- **`sideEffects: false`** added, which lets a bundler prune what a subpath loads. No module in `src` has an import side effect, so the only change to emitted output is that pure re-export modules are tree-shaken out of the bundle instead of being imported for effect. Note this is not yet universal in the repo — several packages don't set it.
- **`typedoc.json`** gains the new entry point — otherwise the subpath silently disappears from the generated API reference. Because TypeDoc now has two entry points, existing deepbook-v3 API-reference anchors gain an `index.` prefix (`MarginManagerContract.depositBase` → `index.MarginManagerContract.depositBase`), so existing deep links into the docs will need updating.

### Release ordering, now enforced rather than documented

After this lands, `@mysten/deepbook-predict@0.3.1` depends on `@mysten/deepbook-v3@^2.1.0`, which is not on npm. The two release workflows are independent dispatches with no ordering between them, and the orchestrator iterates packages in path order — where `deepbook-predict` sorts first — so **the wrong order is the default one**. Every fresh install in that window fails to resolve.

`release-deepbook-predict.yml` now resolves the deepbook-v3 version it will require and fails with an actionable message if that version is not published yet. Verified it discriminates: `2.0.1` passes, `2.1.0` blocks. The file is deleted in Phase 3 along with the package, so the guard covers exactly the window where the hazard exists.

That job also closed a latent hole in the same file: `release` carried an explicit `if:`, which **replaces** the implicit `success()` that `needs:` would otherwise apply, so it would have published even after the smoke gate failed. Both gates are now named in the condition.

### Deployed ids ship with the package

A caller previously supplied `accountPackageId` and `accountRegistry` by hand, and the only place those ids existed in the SDK was inside Predict's config — so an `/account` consumer either imported a 51-module product subpath for two strings or transcribed them, and a redeploy meant finding every copy.

`scripts/sync-deployment.ts` generates `src/deployments/testnet.ts` from the deploy tooling's own `deployment.testnet.json`; each subpath reads a slice. `/account` exports **`getAccountConfig(network)`**, plus **`getDeployment`** and **`getUnits`** — the manifest records the deployment name, source commit and scale constants, and the SDK was dropping all of them.

**The records are frozen.** They are shared module singletons, so returning them unfrozen meant a caller who customised a deployment in place changed what every other caller saw, across subpaths. Spreading to override still works; mutation now throws.

**They are annotated with named interfaces rather than `as const`.** `as const` published the literal ids as *types*, so `getDeployment('testnet').network === 'mainnet'` was a compile error rather than `false`, spread-and-override failed, and every redeploy silently changed the published `.d.mts` with no changeset signal.

Accessors take the client's own wide network type, so `getUnits(client.network)` composes — the earlier `'testnet' | 'mainnet'` admitted exactly one value that always threw while rejecting `'devnet'` at compile time.

### The generator refuses to write a bad record

Everything it interpolates is validated and escaped. This was hardened after a review battery found the first version covered only ids: deleting `chainId` emitted `chainId: 'undefined'` past `tsc` and the whole suite, nothing was escaped (so a crafted value could emit a duplicate object key that silently wins), `network` became a filename unchecked, and a free-text label containing a newline could break out of the header comment into file scope. **17 malformed-manifest cases now throw naming the offending key, leaving the record untouched.**

`sync-deployment` no longer chains prettier: pnpm appends `-- <args>` after the whole chain, so `-- --manifest <path>` reached prettier — which reformatted the manifest in place — and never reached the generator, which silently used the default path. Formatting is `sync-deployment:format`.

The record is dependency-free and sliced, so tree-shaking keeps graphs tight: `/account` is 10 modules and its slice is a few hundred bytes carrying no predict or sessions ids.

## Test plan

- **`@mysten/deepbook-v3`: 337/337**, `tsc --noEmit` clean, build clean, oxlint + prettier clean.
- **`@mysten/deepbook-predict`: 142/142** offline and **11/11 live-testnet e2e** against the deployment, driven through the new subpath with fresh builds.
- **Root export identity proven, not eyeballed.** Built `dist/` from `main` in an isolated tree and from this branch, then compared every root export by name and runtime shape: **52 exports, identical**, and `dist/index.d.mts` byte-identical. Every pre-existing root-graph module is byte-identical.
- **Builder byte-identity proven.** Every `AccountContract` builder's PTB was built on both revisions with fixed sender/gas/budget and compared as full `tx.getData()` — **byte-identical**, including both `createAccountAndDeposit` coin shapes and a zero-amount withdrawal. `deriveAccountWrapperId` matches across 6 owners × 3 registries × 2 package ids.
- **Subpath isolation verified empirically**, not assumed: walking the emitted module graph from `dist/account.mjs` reaches **8 modules and zero spot/margin code**. Shared modules (`bcs/integers`, `contracts/utils`) are emitted once, not duplicated per entry.
- **Export map validated with `attw --pack`** (nothing in CI covers this): 🟢 for `node16 (from ESM)` and `bundler` on both `.` and `./account`. The `node10` mark is inherent to subpath exports in an ESM-only package.
- **Bump levels checked by running `changeset version`**: deepbook-v3 `2.0.1 → 2.1.0` (minor, additive subpath), predict `0.3.0 → 0.3.1` (patch). Each package carries its own changeset so neither changelog inherits the other's prose.
- Generated bindings moved with `git mv`; every relative import resolves unchanged in the new location.

## Follow-ups

- `npm deprecate '@mysten/deepbook-account' 'Moved into @mysten/deepbook-v3 — import from "@mysten/deepbook-v3/account".'` once this is released. Needs npm org access.
- Phase 2 adds `/sessions`; Phase 3 moves Predict to `/predict`.

---

### AI Assistance Notice


manolisliolios (approved):
lgtm

hayes-mysten (approved):

PR #1230, opened 2026-08-27
feat(sui): support validity transaction expirations

hayes-mysten:
## Description

Adds TypeScript SDK support for the `Validity` transaction expiration introduced by
[MystenLabs/sui#27598](https://github.com/MystenLabs/sui/pull/27598) — everything in `ValidDuring`
plus an optional `allowed_proposers` set restricting which validators may propose the transaction
in consensus. Sui fullnodes populate allowed proposers on transactions resolved through
`simulateTransaction`, so without this variant the SDK fails to parse those responses.

Covers the `AllowedProposers`/`Validity` BCS schemas, the v2 transaction schema, the generated gRPC
types, and conversion in both directions. GraphQL and JSON-RPC route through the shared BCS and
protobuf paths, so they need no query changes.

Upstream references: [MystenLabs/sui#27598](https://github.com/MystenLabs/sui/pull/27598),
[MystenLabs/sui-apis#31](https://github.com/MystenLabs/sui-apis/pull/31),
[MystenLabs/sui-rust-sdk#293](https://github.com/MystenLabs/sui-rust-sdk/pull/293).

This supersedes #1236, which implemented the same feature independently; its changes are included
here.

### Fixes to pre-existing behaviour

Two bugs surfaced while reviewing the above. Both predate this PR but are made worse by it, so
they are fixed here rather than left to bite the new variant.

**v1 JSON silently dropped windowed expirations.** `serializeV1TransactionData` collapsed every
non-`Epoch` expiration to `{ None: true }`, and `transactionDataFromV1` collapsed it again on the
way back. A `Transaction.serialize()` → `Transaction.from()` round trip — still common across
wallet and application boundaries — therefore discarded the expiration before signing, and the
rebuilt transaction signed materially broader bytes than the caller built. For `Validity` that also
drops `allowedProposers`, so the transaction becomes proposable by any validator rather than the
named set: an authorization constraint, not just an expiry.

v1's expiration union now carries `ValidDuring` and `Validity`. Note this does **not** help
already-published SDK versions — `TransactionDataBuilder.restore` hands v1 payloads to
`transactionDataFromV1` without validating them against `SerializedTransactionDataV1` first, so
their reader has the same catch-all and will still collapse these to `{ None: true }`. That is
unavoidable; what is fixed is the behaviour going forward. An unrecognised expiration now throws on
read, and an unhandled variant fails to compile on write, so the next variant added cannot
reintroduce the same silent downgrade.

**`ValidDuring` timestamps were dropped in gRPC conversion.** `minTimestamp`/`maxTimestamp` were
hardcoded to `null` in both directions. They now convert losslessly, using the same arithmetic as
upstream `ms_to_timestamp`/`timestamp_to_ms`.

### Other changes

- Allowed proposer indices are validated when encoding, but only checked non-empty when decoding.
  Upstream rejects an empty set while deserializing (the field is `NonEmpty`) but only requires
  strictly-increasing indices in `validity_check`, on submission — and `is_allowed_proposer`
  documents reading an unsorted set as a deterministic miss. Validating sortedness on decode would
  have made the GraphQL and JSON-RPC read paths reject transactions the network accepted, while
  gRPC accepted them. All three transports now agree on what is readable.
- Synchronizes `InvalidTxContext` and `ForwardingAddressRegistryCreate` with upstream BCS and
  protobuf definitions.
- Documents the upstream `InsufficientFundsForWithdraw` rename while preserving the existing SDK
  name — changing the parsed BCS `$kind` would be a breaking API change. The wire representation is
  unchanged.
- `@mysten-incubation/sponsor`: `expirationMaxEpoch` handles the new variant, which is required for
  the narrowing to compile.

## Test plan

- `pnpm turbo build --filter=@mysten/sui`
- `pnpm --filter @mysten/sui test` — typecheck + 522 unit tests
- `pnpm --filter @mysten-incubation/sponsor test` — 63 unit tests
- `pnpm lint`

Added coverage for BCS and gRPC round trips with and without proposer restrictions, millisecond
timestamps, invalid proposer sets when encoding, unsorted and empty sets when decoding, v1 and v2
JSON round trips for both `ValidDuring` and `Validity`, unchanged legacy v1 expirations,
unrecognised v1 expirations, `InvalidTxContext` effects, and the new end-of-epoch variant.

The two round-trip regression tests were confirmed non-vacuous by reverting the source fix and
observing them fail. The write-path exhaustiveness guard was confirmed by adding a fifth expiration
variant and observing `v1.ts` fail to typecheck.

The generated gRPC types were verified reproducible: regenerating from `sui-apis` at `201981c` with
protobuf-ts 2.9.6 produces no diff.

---

### AI Assistance Notice

### DIFF
--- a/.changeset/calm-proposers-validate.md
+++ b/.changeset/calm-proposers-validate.md
@@ -1,5 +1,6 @@
 ---
 '@mysten/sui': minor
+'@mysten-incubation/sponsor': patch
 ---
 
 Add BCS, transaction schema, and gRPC support for `Validity` transaction expirations and allowed proposers. Also synchronize recently added transaction and execution error variants.
--- a/.changeset/calm-proposers-validate.md
+++ b/.changeset/calm-proposers-validate.md
@@ -1,0 +1,5 @@
+---
+'@mysten/sui': minor
+---
+
+Add BCS, transaction schema, and gRPC support for `Validity` transaction expirations and allowed proposers. Also synchronize recently added transaction and execution error variants.
--- a/.changeset/calm-proposers-validate.md
+++ b/.changeset/calm-proposers-validate.md
@@ -4,3 +4,10 @@
 ---
 
 Add BCS, transaction schema, and gRPC support for `Validity` transaction expirations and allowed proposers. Also synchronize recently added transaction and execution error variants.
+
+The deprecated v1 JSON transaction format now represents `ValidDuring` and `Validity` expirations
+instead of collapsing them to `{ None: true }`. Previously a `Transaction.serialize()` ->
+`Transaction.from()` round trip silently discarded the expiration — and, for `Validity`, the set of
+validators allowed to propose the transaction — so the rebuilt transaction signed materially
+broader bytes. An expiration the v1 reader does not recognize is now an error rather than a silent
+downgrade.
--- a/packages/sponsor/src/validators.ts
+++ b/packages/sponsor/src/validators.ts
@@ -421,7 +421,10 @@ export function simulationSucceeds(): Validator {
 function expirationMaxEpoch(expiration: TransactionData['expiration']): bigint | null {
 	if (!expiration || expiration.$kind === 'None') return null;
 	if (expiration.$kind === 'Epoch') return BigInt(expiration.Epoch);
-	const max = expiration.ValidDuring.maxEpoch;
+	const max =
+		expiration.$kind === 'Validity'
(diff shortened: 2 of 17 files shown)


hayes-mysten:
@clud-bot review

hayes-mysten:
@clud-bot check latest changes

hayes-mysten:
@clud-bot pushed a tiny fix, review just this small fix

PR #1233, opened 2026-08-31
fix(slush-wallet): default personal message signing to mainnet

### LINKED ISSUE #27853: Slush zkLogin signPersonalMessage produces stale/invalid maxEpoch, breaks downstream SessionKey retrieval
## Summary
`signPersonalMessage` on a zkLogin Slush account produces an envelope with an
invalid `maxEpoch` value. This breaks any downstream flow that depends on a
correctly epoch-bound personal message signature — in our case, Seal
SessionKey retrieval fails 100% of the time for zkLogin accounts.

On-chain transaction signing on the same wallet, in the same session, is
unaffected — the fault is isolated to signPersonalMessage specifically.

## Evidence
A captured signPersonalMessage envelope from a zkLogin Slush account showed:

    maxEpoch: 8

against a current network epoch of ~1229 at time of capture.

This is not staleness or decay from an earlier session — epoch 8 corresponds
to mainnet's first fortnight (May 2023). The account in question did not
exist then. The same wallet's own on-chain transactions in the same time
window correctly carried maxEpoch: 1230, confirming transaction signing
reads the epoch correctly while signPersonalMessage does not.

## Reproduction
1. Create or use an existing zkLogin account in Slush (web or extension).
2. Call `signPersonalMessage` on any message.
3. Inspect the resulting envelope's `maxEpoch` field.
4. Compare against the actual current network epoch at time of signing.

Expected: maxEpoch reflects a value consistent with the current epoch
(matching what the same wallet's transaction signing produces).

Actual: maxEpoch returns a stale/fixed value (observed: 8), regardless of
current epoch or session freshness.

## Impact
Any dApp relying on signPersonalMessage output for epoch-bound verification
(e.g. Seal SessionKey retrieval) will fail for 100% of zkLogin Slush users,
on the first attempt, not intermittently. This is currently blocking a
production payment flow that uses Slush as its primary zkLogin onramp.

## Environment
- Wallet: Slush (web)
- Login method: zkLogin (social login)
- Confirmed via: direct envelope inspection + cross-reference against the
  same wallet's correctly-signed on-chain transactions in the same window

williamrobertson13:
## Description

- Default chainless Slush personal-message signing requests to Sui mainnet instead of the first advertised account chain, which is devnet.
- Preserve explicitly requested chains.
- Add a patch changeset.

This fixes the invalid zkLogin envelope reported in MystenLabs/sui#27853. The wallet-standard personal-message chain is optional, but the Slush adapter previously substituted account.chains[0]. Since SUI_CHAINS starts with sui:devnet, Slush created a devnet-bound zkLogin signature (observed maxEpoch 8) for a message that downstream consumers verified on mainnet.

Fixes MystenLabs/sui#27853

## Test plan

- [x] pnpm turbo build --filter=@mysten/slush-wallet
- [x] pnpm --filter @mysten/slush-wallet lint

---

### AI Assistance Notice

### DIFF
--- a/packages/slush-wallet/src/wallet/index.ts
+++ b/packages/slush-wallet/src/wallet/index.ts
@@ -22,7 +22,12 @@ import type {
 	Wallet,
 	WalletIcon,
 } from '@mysten/wallet-standard';
-import { getWallets, ReadonlyWalletAccount, SUI_CHAINS } from '@mysten/wallet-standard';
+import {
+	getWallets,
+	ReadonlyWalletAccount,
+	SUI_CHAINS,
+	SUI_MAINNET_CHAIN,
+} from '@mysten/wallet-standard';
 import { mitt, type Emitter } from '@mysten/utils';
 import type { InferOutput } from 'valibot';
 import { boolean, object, parse, string } from 'valibot';
@@ -258,7 +263,7 @@ export class SlushWallet implements Wallet {
 			type: 'sign-personal-message',
 			message: toBase64(message),
 			address: account.address,
-			chain: chain ?? account.chains[0],
+			chain: chain ?? SUI_MAINNET_CHAIN,
 			session: getSessionFromStorage(),
 		});
 
--- a/.changeset/tidy-slushes-sign.md
+++ b/.changeset/tidy-slushes-sign.md
@@ -1,0 +1,5 @@
+---
+'@mysten/slush-wallet': patch
+---
+
+Default personal message signing requests without a chain to Sui mainnet.


packages/slush-wallet/src/wallet/index.ts
```diff
@@ -258,7 +263,7 @@ export class SlushWallet implements Wallet {
 			type: 'sign-personal-message',
 			message: toBase64(message),
 			address: account.address,
-			chain: chain ?? account.chains[0],
+			chain: chain ?? SUI_MAINNET_CHAIN,
```
williamrobertson13, line 266: this is just a better default, we might want to change this behavior and make chain optional in both web/ext so the absence of the chain parameter means that the wallet auto-picks based on the selected network

hayes-mysten (approved):

PR #1247, opened 2026-09-04
feat(sui): support allowance-funded coins and balances

hayes-mysten:
## Description

Spend from an allowance with `tx.balance({ allowance, balance, type })` or `tx.coin({ allowance, balance, type })`. The SDK resolves the funder, constructs the withdrawal, and redeems it into the requested Move value:

```ts
const coin = tx.coin({
  allowance: allowanceId,
  balance: 10_000_000n,
  type: USDC,
});
tx.transferObjects([coin], recipient);
```

App-bound allowances use the same helpers with a `SpendPermit<A>` from the app's authorization call. The transaction must still be signed by the spender:

```ts
const permit = tx.moveCall({
  target: `${appPackage}::subscriptions::authorize_payment`,
  arguments: [tx.object(subscriptionId)],
});
const balance = tx.balance({
  balance: 10_000_000n,
  type: USDC,
  allowance: { objectId: allowanceId, app: { type: APP_TYPE, permit } },
});
```

`balance` accepts bigint, safe integer numbers, and decimal strings. An optional `funder` on the allowance reference skips metadata lookup. Otherwise, lookup checks the allowance's coin type and app binding. App permits remain transaction arguments through copying, serialization, and intent expansion. A permit authorizes one spend without binding the amount or recipient; apps enforcing those must redeem inside their own Move function using a low-level withdrawal.

Allowance spends never fall back to sender funds and do not pay gas. The separate `AllowanceBalance` and `CoinWithBalance` intents share a resolver that reserves allowance withdrawals before coin selection. Gas selection also accounts for reservations, including self-funded allowances and funders sponsoring gas. Wallet analysis attributes non-app and app-bound allowance spending to the funder and recipient. If withdrawal redemption cannot be tracked (for example, inside a custom Move call), balance-flow analysis fails and auto-approval is blocked.

The combined built-in resolver handles allowance withdrawals before ordinary coin selection. Custom resolver overrides are responsible for their own ordering. Allowance metadata batching is delegated to Core, preserving its existing behavior. MVR coin and app aliases resolve before metadata checks. Spending upgrades existing read-only allowance inputs even when the funder is supplied. String amounts require nonempty decimal digits.

For mixed funding, serialization can preserve both intents or only `CoinWithBalance`. Preserving only `AllowanceBalance` remains rejected because ordinary coin selection needs its reservations first.

Wire support covers BCS, transaction inputs, gRPC protos, GraphQL schema, and shared transaction conversion. A separate allowance guide covers creation, limits, spending, app permits, and links to the Sui framework docs; general coin/balance docs link to it.

Rebased onto `main` at `9e8ba6bd`. Retains the PR's `sui-tools` image `892d777ccdf414f13b9421641831fc57462a8c6e` (arm64 suffix where applicable): main's Sui 1.70.0 pin was tested and fails allowance creation with `TypeNotFound`. The app test fixture pins its framework dependency to the same revision. JSON-RPC genesis placeholders and the specific `gasless_allowed_token_types` parity exception are updated for this image.

**Requires allowance-enabled protocol v137. Do not merge until devnet supports it.**

The deprecated native JSON-RPC OpenRPC types still lack `senderAllowance`; unified Core reads support it through the shared BCS parser.

## Test plan

- Sui type checks and all 628 unit tests pass, including app mismatches, serialization, permit/result remapping, mixed funding, and gas reservations.
- Wallet SDK: 154 passed, 1 existing skip; app attribution now uses the public helper.
- Allowance e2e: 43 tests passed on the pinned image, covering both app helpers with/without known funders, custom app authorization, funder debits, sponsorship, and all three transports.
- Sponsor: 71 tests passed, including type checks and examples.
- Read-only allowance inputs: 9 targeted e2e tests passed across all three transports, covering known-funder spending from initially read-only inputs.
- Before the app-helper addition, the full Sui e2e suite passed: 52 files, 937 passed, 1 expected failure, 14 skipped.
- All 14 wallet/SDK dependency builds, lint, docs generation, and docs validation pass.

---

### AI Assistance Notice

Withdrawal sources use `tx.withdrawal({ amount, type, from: 'allowance', allowance, funder })`. The exported `WithdrawalOptions` union requires allowance and funder together for this source. `from: 'sender'` is the default; `from: 'sponsor'` is also supported. Callers do not construct `$kind` values.


packages/sui/src/transactions/Transaction.ts
```diff
@@ -682,8 +682,19 @@ export class Transaction {
...
 	 */
-	withdrawal({ amount, type }: { amount: number | bigint | string; type?: string | null }): {
+	withdrawal({
+		amount,
+		type,
+		withdrawFrom,
+	}: {
+		amount: number | bigint | string;
+		type?: string | null;
+		withdrawFrom?: WithdrawFrom;
```
hayes-mysten, line 696: I'm not sure that this is the right format we want.  It may be better to add a withdrawFromAllowance or something instead.  $kind is not a user friendly way to consume this API

hayes-mysten:
@clud-bot review

hayes-mysten:
@clud-bot rereview

hayes-mysten:
@clud-bot review latest changes

PR #1273, opened 2026-09-16
fix(sui): break named-package utility import cycle

hayes-mysten:
## Description

Break the circular dependency between `sui-types` and `move-registry` that can cause Next.js 16 Turbopack production builds to fail with `Cannot access 'isValidNamedPackage' before initialization` when `coinWithBalance` normalizes the SUI type at module load time.

Move the named-package validator and its constants into `named-packages.ts`, which imports only `suins`. Both utilities use this module, and existing exports and validation behavior are preserved. Includes a patch changeset.

## Test plan

- All 63 unit test files passed (652 tests).
- Source and test TypeScript checks passed.
- Formatting and `git diff --check` passed.
- Checks used installed binaries directly because pnpm stopped at an out-of-sync dependency prompt.
- The Next.js Turbopack reproduction was not rerun locally.

---

### AI Assistance Notice


hayes-mysten:
@clud-bot please review this PR.

